PR #196, opened 2021-10-18
Add aliases for Azure Classic's Overview and Setup pages

praneetloke:


praneetloke:
We will do this at the CFN-level. Closing this.

Issue #275, opened 2021-10-29
a bunch of packages have the wrong name in the libraries/sdk packages section


susanev:
## overview
a bunch of packages have the wrong name in the libraries/sdk packages section
prolly due to some copy/paste mistakes

e.g., https://www.pulumi.com/registry/packages/gcp-global-cloudrun/#sdk-packages

Issue #272, opened 2021-10-29
move sdk packages section from overview to install+config page, for all packages


susanev:
## overview
it was intended that the install info live on the install+config page, but access to that information is still in the overview pages
move sdk packages/libraries section from overview to install+config page, for all packages

Issue #222, opened 2021-10-19
[Epic] Single-repository Pulumi Packages


infin8x:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Epic details

Today, a Pulumi Package's contents are spread across a few places (see list below). To make it easier for anyone in the Pulumi Community to author a package and prepare for #206, we should update our publishing processes to source all of that content from the package's repository. We should then move the content for all of our existing packages into their repositories.

- The **code** and the **schema** are in the package's repository
- The **API documentation** (generated from the schema) is in https://github.com/pulumi/docs
- The **hand-written documentation** (Overview, Installation & Configuration) is in https://github.com/pulumi/registry
- The **how-to guides** are in https://github.com/pulumi/examples

## Work items

- [ ] Interested Parties 
- [ ] Quality plan 
- [ ] #223
- [ ] #224
- [ ] #221
- [ ] Update the Pulumi Packages publishing guide to match the changes made in this Epic
- [ ] Update any relevant Pulumi developer documentation to match the changes made in this Epic

## Launch readiness 🎁

[Link](link-to-release-tracker)

susanev:
we are going a different direction with this, so closing.

PR #280, opened 2021-10-29
relocate install links and other small fixes

susanev:
## overview
fixes https://github.com/pulumi/registry/issues/275 and fixes https://github.com/pulumi/registry/issues/272

## details

- [x] standardize sdk packages/libraries heading to: `Installation`
- [x] standardize first line in the installation section to be: `The <name> provider is available as a package in all Pulumi languages:`
- [x] move installation from the overview to the install+config page
- [x] fix some heading level issues
- [x] fix some providers that had the wrong name in their docs
- [x] remove quote styles for things that are not quotes (if anyone really wants these to be note styles let me know)
- [x] remove images that were on some of the provider overview pages, the images are now in the headers
- [x] removed links to the repos, as that's now the source code link in the header

other things i think we should also do...

- [x] move the sentences related to authentication and configuration to the configuration sections (if they exist)
- [x] fix the link text for project readmes to be specific about the project name
- [x] AWS QuickStart Redshift is not using a heading for the example in overview
- [x] missed quote styles in AWS QuickStart Redshift (this is in a shortcode): https://github.com/pulumi/pulumi-hugo/pull/807
- [x] missed quote styles in azuread
- [x] missed removing link to kong repo

### DIFF
--- a/themes/default/content/registry/packages/aiven/installation-configuration.md
+++ b/themes/default/content/registry/packages/aiven/installation-configuration.md
@@ -34,4 +34,4 @@ before it can be used to create resources. Once the credentials obtained, there 
     ```
 
 Remember to pass `--secret` when setting `aiven:apiToken` so that it is properly encrypted. The complete list of
-configuration parameters is in the [Avien Provider README](https://github.com/pulumi/pulumi-aiven/blob/master/README.md).
+configuration parameters is in the [Aiven Provider README](https://github.com/pulumi/pulumi-aiven/blob/master/README.md).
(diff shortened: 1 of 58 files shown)


infin8x (approved):
Thanks for taking a close look through these! I think this is fine as a step for now to get us closer to the designs as you intended. But, food for thought in the future: I want us to think about how a Pulumi Package can stand alone (e.g. without being published to Registry) as a single repository (https://github.com/pulumi/registry/issues/222). To that end, it would be good for the package's README.md (in the root of the package's repository) to contain a lot of the useful information that we currently spread across the Overview and Installation & Configuration pages.

djgrove (approved):
Thanks for doing this Susan!

PR #286, opened 2021-10-30
Change config prefix from `aws:` to `aws-native:`

d4nyll:
Using `aws:` in the `Pulumi.xyz.yaml` file didn't work for me. It will return with an error like the following:

```
Diagnostics:
  aws-native:codestarnotifications:NotificationRule (dev_api-code2ecr_slack):
    error: missing required property 'region'
```

I had to change the prefix to `aws-native:` (e.g. `aws-native:region: ap-northeast-2`)

```yaml
config:
  # Find more AWS-related configuration options at https://www.pulumi.com/docs/intro/cloud-providers/aws/
  aws:region: ap-northeast-2
  aws:profile: dev
  aws:defaultTags:
    # This will add a tag of `pulumi: true` to all resources managed by Pulumi
    # When you see this tag in the console, you should note _not_ to modify it in the console
    tags:
      pulumi: true
  aws-native:region: ap-northeast-2
  aws-native:profile: dev
```


lukehoban (approved):
Thanks!

lukehoban:
Looks like the same changes were also raised in https://github.com/pulumi/registry/pull/270 yesterday, so will close this one out and get that one merged.

PR #290, opened 2021-11-01
Format Kubernetes config as table

infin8x:
Thanks @susanev for moving this config over to the install/configure page. I suggest we also format it as a table to match the other native providers.


susanev:
Yah tots agree, was just relocating the content as it existed. i don't have
access to github right now, but if you can get someone else to approve feel
free to merge.

cnunciato (approved):
Looks good!

PR #298, opened 2021-11-03
Update the README

cnunciato:


susanev:
@cnunciato is this something we still want to get in?

cnunciato:
Yep, I want to update the READMEs of all these repos, just had to set it aside for a bit. Plan to come back to it

CONTRIBUTING.md
```diff
@@ -0,0 +1,88 @@
...
+
+    ```html
+    {{ partial "head.html" . }}
+    ```
+
+- **Front-matter variables.** You can define a front-matter variable in the YAML section at the top of a file. For instance, the you could add the following front matter `foo: "bar"`, and then reference the variable in markdown with the syntax `{{< param foo >}}`.
+
+  - **no_on_this_page** Specify this variable to prevent displaying an "On This Page" TOC on the right nav for the page.
+  - **block_external_search_index** Specify this variable to prevent crawlers from indexing the page.
+## Style guide
```
susanev, line 54: for this section might be cool to get anything not in the hugo style guide there, and then link to it?

cnunciato: For sure, yeah. We can treat that new doc as the canonical one now, and then link all others to it. (This WIP predates that doc, and this file (CONTRIBUTING.md) is just a straight-up copy from [pulumi/docs/CONTRIBUTING.md](https://github.com/pulumi/docs/blob/master/CONTRIBUTING.md) that I was planning on trimming down to just what makes sense to have in this repo.)

cnunciato:
Reworking this locally, will resubmit.

PR #326, opened 2021-11-10
Regen metadata for all packages to include repo_url and schema_file_path

praneetloke:
This PR contains regenerated metadata files to bring in changes from https://github.com/pulumi/docs/pull/6788. Specifically, the `metadata` command in `resourcedocsgen` tool will now also include a package's repository and schema file path information in the metadata file.

A special note about the path to the schema file. In many cases, packages use the location `/provider/cmd/pulumi-resource-<package-name>/schema.json`. That is considered the default location. If the location of a package's schema file differs from that, the `gen_package_metadata.sh` script does accept an optional argument that is the path of the schema file, which is then subsequently passed to the tool as well.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/pulumi/registry
 go 1.16
 
 require (
-	github.com/pulumi/pulumi-hugo/themes/default v0.0.0-20211112201932-5e025d42b72d // indirect
+	github.com/pulumi/pulumi-hugo/themes/default v0.0.0-20211113032058-53c36f72ce11 // indirect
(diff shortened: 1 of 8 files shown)


praneetloke:
Also I am sorry for all the diffs resulting from ordering the fields in each file 😁  I updated the `resourcedocsgen` tool to use a single library for working with YAML. That ordered all the property names.

cnunciato (approved):

cnunciato (approved):
🚀

praneetloke:
Alright merging this now.

Issue #284, opened 2021-10-29
new packages being added to registry before expected


susanev:
## overview
see [internal slack convvo](https://pulumi.slack.com/archives/C01QV7KBZN0/p1635541885047800)
tailscale showed up on the browse page and had API docs before the rest of the pages were added and the package was ready to be released

## questions

- should we change something so these don't automerge until all the files, and the logo are in place?
- do we need some kind of draft feature?

susanev:
cc/ @praneetloke

EvanBoyle:
Should be fixed by #237

praneetloke:
While chatting about the status of #237 with @cnunciato this issue came up. We think that we could potentially fix this by simply adding a check in our `list.html` layout to only include a package if its hand-authored files are present. That could be an easy fix if that works.

susanev:
@praneetloke do you know if this is fixed now?

cnunciato:
This is not fixed, but it could be an easy one, provided we're okay with failing the website build when it happens. (Specifically, making the links on the index page `relref`s, so building them would fail when there isn't a package page to link so.) I'll prep something real quick for consideration.

praneetloke:
Thank you for taking this on, @cnunciato. This would be a nice quality of life improvement for the automations.

PR #363, opened 2021-11-15
Delete artifactory.yaml

praneetloke:
Remove the metadata file until the overview and installation pages are added to avoid listing on the registry


praneetloke:
I'll go ahead and merge this since it is just removing the metadata file for something that's not ready to be listed yet. The problem is captured with #284.

Issue #237, opened 2021-10-19
Move package API docs from `docs` to this repo to facilitate a single-place of ownership


praneetloke:
We want to move all API docs and how-to guides from the `docs` repo to `registry` to their respective folders. We are looking at ways to allow us to still run `make serve` (for development within the repo) and not have Hugo kill our dev machines. For that @cnunciato will look into excluding the `api-docs` specifically from Hugo processing those in dev mode. We would add an alternate `make` target for devs who wish to see the API docs locally that would involve doing a `build`. So there would still be a way for devs to “see” the API docs locally which in today's world one can't unless they copy the `api-docs` folder over for a package from `docs` to here just to see something work end-to-end. This would be an improvement over what we have today.

Then we would open a PR in `docs` repo to delete all API/how-to guides related folders. That is, `content/registry/**/*` as well as `static/registry`. We would also move some of the `tools/` over to the `registry` repo as well. This means we’ll need to update the existing docs generation workflow in that repo to do a “redirect” of the `repository_dispatch` event to the `registry` repo where we will have the same exact workflow. This will buy us time to be able to update all of our providers to post a `repository_dispatch`, when we release new versions, to the `registry` repo instead. This is so we don’t break the  mechanism that is in-place today.

This helps the self-service contribution story (#206) immensely as everything related to packages would be housed in the same repository. It would also make it easy for us to author automation that does the right things.

UPDATE: We eventually ended up doing something entirely different from the above. See https://github.com/pulumi/registry/issues/237#issuecomment-963384022.

praneetloke:
Last week, @cnunciato and I tried to pull in updates, from our PR branch related to this, into a `docs` repo PR branch and hit a problem with the `registry` repo module ending up being too large for Go's module systems. The problem we hit was the infamous `module source tree too big`. In re-thinking why we generate the API docs at all, we have an alternative proposal that actually solves a few more things about our current process of API docs generation.

We would not generate API docs AOT and check them into a repository anymore since they don't serve any other purpose than to include them in a build. Instead, we think we can defer the generation of the API docs to just-in-time by simply relying on the metadata files as the source artifact as to which packages the API docs would need to be built for. To be precise at the time of site build we could invoke `resourcedocsgen` to walk the metadata files dir and simply run the generator for every one of those packages. Almost all of the information needed to generate documentation is in those files. The only two pieces of information that we do not persist today are the package repo and the path to the Pulumi schema file in the repo. Those can be easily added.

[...]

PR #364, opened 2021-11-15
Generate API docs in the PR build and update the README

praneetloke:
Related to #237.

I have updated the README file in this repo to show how one could generate docs for any package (or all packages, if they wish) locally. I also updated the PR workflow such that it generates the API docs for all packages so that we can have full previews of API docs in PRs.

### DIFF
--- a/scripts/ci/build.sh
+++ b/scripts/ci/build.sh
@@ -12,7 +12,6 @@ go install github.com/pulumi/docs/tools/resourcedocsgen@master
 PKGS=(
     "aiven"
     "aws"
-    "azure-native"
 )
 
 echo "Generating API docs for ${PKGS[*]}..."


kimberleyamackenzie (approved):

praneetloke:
@cnunciato have you seen this before? I see that we set [`GOGC`](https://github.com/pulumi/docs/blob/master/scripts/build-site.sh#L37) in `docs` to a small percentage. Is that what you would recommend here as well?

```
 ./scripts/ci/build.sh: line 22:  4041 Killed                  hugo --minify --templateMetrics --buildDrafts --buildFuture -e "preview"
Makefile:27: recipe for target 'ci-pull-request' failed
      4042 Done                    | grep -v -e 'WARN .* REF_NOT_FOUND'
```

Alternatively, I can also change this PR to not generate all API docs but maybe just a few? Just letting you know that's an option worth considering too since generating all API docs means that the Hugo build part will take a few more minutes than it does today. It's a small tradeoff I think to having the API docs for all packages in the PR preview which I do think avoids confusion for people wondering why they can't see some API docs.

cnunciato:
@praneetloke Ah, yes. We shouldn't build all resource docs here on every PR (at least not ideally, I woudn't think), as that'd slow down the PR preview significantly. Building one or two seems fine, though.

scripts/ci/build.sh
```diff
@@ -7,6 +7,15 @@ source ./scripts/ci/common.sh
 # URL to the Pulumi conversion service.
 export PULUMI_CONVERT_URL="${PULUMI_CONVERT_URL:-$(pulumi stack output --stack pulumi/tf2pulumi-service/production url)}"
 
+echo "Generating API docs for all packages for PR preview..."
+echo ""
+go install github.com/pulumi/docs/tools/resourcedocsgen@master
+resourcedocsgen docs registry \
```
cnunciato, line 13: Yeah if we could limit this to just a couple of packages, like aws and aiven or something, that'd be ideal, from an iteration-speed perspective.

cnunciato (approved):
My only request would be to limit the number of packages in PR builds to one or two. Otherwise, just one suggestion for the README.

scripts/ci/build.sh
praneetloke: So the full build took 22 mins. with all the providers. I am slightly inclined to leave it that way. I thought it was going to be more than 30 mins for the PR build based on what we see in the `docs` repo. Maybe it's just me but I don't feel 22 mins is _that_ bad?

README.md
```diff
@@ -78,14 +78,48 @@ If you want to link to a page that exists on https://pulumi.com but not in this
...
-cp ../docs/static/registry/packages/navs/aws.json ./themes/default/static/registry/packages/navs/
+# Copy the AWS how-to guides from a local/sibling clone of pulumi/docs.
+cp ../docs/content/registry/packages/aws/how-to-guides ./themes/default/content/registry/packages/aws/
 ```
 
+#### Temporarily generating API docs for packages
+
+The API docs for packages can be generated on-demand using the `resourcedocsgen` tool. The PR build for this repo, for instance,
+generates the API docs for all packages from the latest commit hash of the PR being built. Likewise, as long as your branch
+is pushed up to GH, you can generate API docs using any commit hash from your PR branch locally as well. This is because
```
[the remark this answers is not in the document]
praneetloke: > but I suspect most users who've cloned the registry to work on a branch would be happy just to generate docs with that same branch (assuming this is even possible -- which now that I write this, I realize it may not be?)

hmm I don't quite follow the question here. Are you asking if it is possible to generate API docs using a specific branch of the registry? If yes, then yes. It's possible with the `--branch` flag. If you omit the `--branch` flag, then `master` branch will be the default.

.github/workflows/pull-request.yml
```diff
@@ -5,8 +5,9 @@ on:
       - master
 jobs:
   preview:
-    # Only run this job for events that originate on this repository.
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    # Only run this job for events that originate on this repository and if the automation/merge label is not applied to the PR.
+    # PRs with the automation/tfgen-provider-docs label contain only metadata file changes and don't need a PR build.
+    if: github.event.pull_request.head.repo.full_name == github.repository && !contains(github.event.pull_request.labels, "automation/tfgen-provider-docs")
```
praneetloke, line 10: I added this part to skip PR builds for those PRs that get opened through automation which contain only metadata file changes. We auto-merge those and don't really need the PR preview links for those since we don't typically pay attention to those.

scripts/ci/build.sh
praneetloke: Let me know what you think. I am hoping we can try it out with a full docs build to favor having all docs in the preview. Of course, if it becomes a major pain point for PRs we can certainly change the command to generate docs for just a handful of them.

README.md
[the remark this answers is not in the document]
praneetloke: I updated this based on what I think I understood. Not sure if this is now a bit more clear to read. Can you please check the latest and let me know what you think?

scripts/ci/build.sh
cnunciato: To me, it all comes down to who'll be consuming these PRs. Compared to 2-3 minutes for a PR preview today, having to wait 20 minutes per commit is a pretty big deal, especially when you're iterating on a PR and want to be able to view and share preview links with others. Speed of iteration is important here, IMO, so I'd fairly strongly prefer we continued to optimize for that, and made it _possible_ for people to run full builds if they want to. After all, it's highly unlikely anyone's going to be looking closely at long-tail API-docs content in any PR here; more likely, they're going to spot a few things in a provider or two, and move on. So I'd really like to keep the speed of iterating if at all possible, since Docs PRs will always be complete. What do you think?

praneetloke: > It also seems unlikely folks will be looking closely at long-tail API-docs content in any one PR

That's a good point. How about `aws`, `azure-native`, `gcp` and `kubernetes`? They are also the featured packages right now. I can see how long those take. Should be less than a minute.

README.md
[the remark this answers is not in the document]
cnunciato: Ok yeah, sorry for my confusing words! What I meant was the most common use case is probably something like this:

1. Clone master and make a feature branch
1. Start developing
1. Notice you don't have the API docs you want
1. Build them

So the README should probably lead with making it clear how to do just that. (Which I believe would look like the example I quoted -- since what `--branch` would they be providing if they haven't pushed a branch yet?) The more advanced use case woudl then be "I want to pull API docs from somewhere _other_ than master", in which case we can explain how to do that by showing an example with `--branch` or `commitSha`. Does that make sense?

scripts/ci/build.sh
cnunciato: Awesome, sounds good. An extra minute or so seems like a totally reasonable thing in exchange for the value of having a few representative packages, for sure.

praneetloke: > An extra minute or so seems like a totally reasonable thing in exchange for the value of having a representative package or two.

So it turns out that just generating, building, _and_ copying the docs for those packages takes 10 mins. With the Hugo build being the bottleneck. When I slim down the list to just two big providers `aws`, and `azure-native`, the total time comes down to 8 mins. With just `aiven` and `aws` it is 5 mins. I am inclined to leave it at `aws` and `aiven` for now and will leave expanding that list for another day if we feel that we want to do that.

By the way, in all of those trials, _generating_ the docs using the `resourcedocsgen` tool was very quick in contrast with the other things involved in building the site. It probably comes as no surprise that the more pages we add to Hugo, the longer it takes to build.

cnunciato: That sounds good to me. Thanks @praneetloke!

PR #375, opened 2021-11-18
Fix if-condition in the PR workflow file

praneetloke:
I noticed that an if-condition that I updated with https://github.com/pulumi/registry/pull/364 doesn't actually work the way I thought it did when I originally tested it with that PR. I noticed that a PR build attempted to run despite having the `automation/tfgen-provider-docs` label. It seems that I can't use a unary operator to test conditionals and must use the binary operator equivalent. GHA's [`Expressions`](https://docs.github.com/en/actions/learn-github-actions/expressions) doc doesn't have a clear example for doing logical checks but I confirmed the change in this PR really does work the way I expect it to through a [testing PR](#374) I had opened just before this one.


cnunciato (approved):
👍 Sorry I missed this one.

PR #419, opened 2021-11-24
remove tutorials from docs left nav

### LINKED ISSUE #312: remove tutorials from the left nav in docs
## overview
on 12/04 we should evaluate whether there is significant usage of the tutorials item in the left nav in docs (which sends you to registry) and if there is not we should remove it.

susanev:
this is removing the tutroials docs left nav which was left around because of concerns during the registry launch
however, its a pretty confusing experience because it sends them to the registry landing page where there are no tutorials so we are removing it

fixes #312

### DIFF
--- a/themes/default/content/registry/_index.md
+++ b/themes/default/content/registry/_index.md
@@ -11,10 +11,6 @@ menu:
         identifier: cloud-providers
         name: Cloud Providers
         weight: 6
-    userguides:
-        identifier: tutorials
-        name: Tutorials
-        weight: 1
 aliases:
     - /registry/packages
     - /docs/tutorials


susanev:
@cnunciato thoughts on keeping the ` /docs/reference/tutorials/` alias?

cnunciato (approved):
> thoughts on keeping the `/docs/reference/tutorials/` alias

If we removed this, any bookmarks or search-engine links pointing to it would 404, so since there's no pressing need to do that, leaving it here a while longer makes sense.

At some point we should do an audit of our aliases and other redirects to clear out the ones we don't need anymore (just so we aren't spending time building pages and maintaining code we don't have to). I'll make a note and get something filed for it. For now, though, this seems fine. 👍

PR #448, opened 2021-12-09
Add google-native configuration details

danielrbradley:
Related to updating notes in the provider readme: https://github.com/pulumi/pulumi-google-native/pull/253


mikhailshilkov (approved):
Somehow I assumed this was auto-generated but apparently it's not. Thank you for updating!

danielrbradley:
@mikhailshilkov think automating this would be a really good idea!

mikhailshilkov:
> think automating this would be a really good idea!

Is there anything that prevent this? If not, mind opening an issue to track automation?

PR #519, opened 2022-01-05
Regenerate the metadata file for pulumi-eks

cnunciato:
This PR updates the Registry to use the latest version of [pulumi/eks](https://github.com/pulumi/pulumi-eks). It's the result of running 

```
./scripts/gen_package_metadata.sh eks
``` 

from the root of pulumi/docs. (I ran the tool for all other `component` packages as well, but `eks` seems to be the only one that turns up a diff.)

### DIFF
--- a/themes/default/data/registry/packages/eks.yaml
+++ b/themes/default/data/registry/packages/eks.yaml
@@ -10,5 +10,5 @@ publisher: Pulumi
 repo_url: https://github.com/pulumi/pulumi-eks
 schema_file_path: /provider/cmd/pulumi-resource-eks/schema.json
 title: Amazon EKS
-updated_on: 1636570432
-version: v0.35.0
+updated_on: 1638563109
+version: v0.36.0


justinvp (approved):
LGTM

Do our other packages automatically open a PR in this repo when they're updated? If so, what do we need to add to the `pulumi-eks` repo to do that?

cnunciato:
They do for providers, yeah; that's a good question. @praneetloke or @stack72 would probably have the best answer, I'm not exactly sure what's preventing us from doing the same thing for components.

praneetloke:
@cnunciato `pulumi-eks` is setup for [triggering a `repository_dispatch` against the `docs` repo](https://github.com/pulumi/pulumi-eks/blob/master/.github/workflows/release.yml#L244). But it seems there was a [problem](https://github.com/pulumi/docs/runs/4413104753?check_suite_focus=true) with the `ensure` `make` target in the `docs` repo that prevented a bunch of providers from getting updated metadata files. I don't know if we get build failure notifications for any of the workflows I  the `docs` repo? Perhaps that could be used to monitor future problems like this.

stack72:
This has all been fixed up in the docs repo - thanks @praneetloke !!!

cnunciato:
Awesome, thank you both!

PR #556, opened 2022-01-13
Update to version 0.1.8 of scaleway provider

jaxxstorm:


themes/default/data/registry/packages/scaleway.yaml
```diff
@@ -1,14 +1,14 @@
-category: Infrastructure
```
praneetloke, line 1: Did you want to keep your package categorized as `infrastructure`? If so, you could add a tag `category/infrastructure` to the `Keywords` slice in your [`resources.go`](https://github.com/jaxxstorm/pulumi-scaleway/blob/master/provider/resources.go#L79) file. See the description for `Keywords` in our [docs](https://www.pulumi.com/docs/guides/pulumi-packages/schema/#pulumi-package-schema).

themes/default/data/registry/packages/scaleway.yaml
```diff
@@ -1,14 +1,14 @@
...
-logo_url: "https://raw.githubusercontent.com/jaxxstorm/pulumi-scaleway/master/assets/scaleway-svgrepo-com.svg"
+logo_url: ""
 name: scaleway
 native: false
 package_status: public_preview
-publisher: jaxxstorm
+publisher: Lee Briggs
 repo_url: https://github.com/jaxxstorm/pulumi-scaleway
 schema_file_path: /provider/cmd/pulumi-resource-scaleway/schema.json
 title: scaleway
```
cnunciato, line 12: Nit, this should be title-capped as well.

```suggestion
title: Scaleway
```

cnunciato:
@praneetloke Do these changes need to be applied before this can merge? I'm assuming these changes all need to be made in the upstream repo -- is that right?

susanev:
the previous pr added the schema to this repo
https://github.com/pulumi/registry/pull/532

praneetloke:
@cnunciato I left comments in this PR to resolve in the package's schema which should then flow into the metadata. Once those are resolved this can be merged.

susanev:
oh, @cnunciato can we open up a pr for Lee over in that repo then?

praneetloke:
The [`ProviderInfo`](https://github.com/jaxxstorm/pulumi-scaleway/blob/master/provider/resources.go#L77 ) struct for this package in its repo should be updated with `DisplayName`, `Publisher`, and the `category/infrastructure` tag in the `Keywords` slice, as well as the `LogoURL` property. Then regenerating the metadata will include those always in future revisions.

stack72:
Folks, this isn't addressable in v0.1.8 of the provider as this would require a v0.1.9 - I suggest we merge this and then we can make the upstream changes

praneetloke:
@jaxxstorm can you update the metadata file manually for this PR and make sure your package is setting the necessary fields for future releases?

jaxxstorm:
I am having a very hard time figuring out what's needed to merge this PR.

I've updated my `resources.go` to include what I believe is being asked for in `master`. I've rerun the following command:

```
registrygen metadata --repoSlug jaxxstorm/pulumi-scaleway --version v0.1.8 --schemaFile=/provider/cmd/pulumi-resource-scaleway/schema.json --publisher "Lee Briggs" --category cloud
```

This is not meeting the feedback from this PR. I've also pushed several changes to this PR that address the feedback. What do I need to do?

themes/default/data/registry/packages/scaleway.yaml
```diff
@@ -1,14 +1,14 @@
-category: Infrastructure
+category: Cloud
 component: false
 description: A Pulumi package for creating and managing scaleway cloud resources.
 featured: false
-logo_url: "https://raw.githubusercontent.com/jaxxstorm/pulumi-scaleway/master/assets/scaleway-svgrepo-com.svg"
-name: scaleway
+logo_url: ""
+name: Scaleway
```
praneetloke, line 6: This should remain as the package name so please revert this. The `title` property is the display name as called out by Christian in his [comment](https://github.com/pulumi/registry/pull/556/files#r785054269).

praneetloke:
@jaxxstorm updating `resources.go` is good but I guess your package's schema was probably not regenerated as a result of just updating that. In case you didn't know updates to `resources.go` will end up in your serialized schema only if you regenerate the schema. It's something I missed mentioning for future versions of your package. Regenerating the metadata with an updated `resources.go` will do nothing until your schema is updated and you make a new release for it so that you can point the metadata generator at that updated version.

For this PR, in addition to accepting Christian's suggestion for the `title` property, if you revert the changes in your metadata file as I called out in https://github.com/pulumi/registry/pull/556#discussion_r784325895 and https://github.com/pulumi/registry/pull/556#discussion_r784326787 that would be enough for this PR.

susanev:
@jaxxstorm do you have what you need? if not, would folks be up for jumping on a zoom together to help fill in the gaps and make sure this can get in quickly?

jaxxstorm:
I believe this is good to go now?

PR #559, opened 2022-01-14
Use relrefs to build links to package pages

### LINKED ISSUE #284: new packages being added to registry before expected

cnunciato:
Passes the current page context into the `package` partial so it can be used to build proper `relref`s.

Fixes #284.

### DIFF
--- a/themes/default/content/registry/packages/scaleway/_index.md
+++ b/themes/default/content/registry/packages/scaleway/_index.md
@@ -33,7 +33,7 @@ import pulumi_scaleway as scaleway
 
 public_ip = scaleway.InstanceIp("example")
 
-server = scaleway.InstanceServer("example", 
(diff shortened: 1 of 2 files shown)


cnunciato:
Looks like there are issues with the content that went in with #532. I'll fix those up as well and submit here.

susanev:
@cnunciato yah, mentioned in this pr review too https://github.com/pulumi/registry/pull/556

praneetloke (approved):

PR #570, opened 2022-01-19
Upgrade Hugo, Node, Go versions

cnunciato:
Follow-on work from https://github.com/pulumi/pulumi-hugo/pull/983. (See that PR for context. Same idea here.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules
 themes/*/resources
 public
 .DS_Store
+.hugo_build.lock
--- a/config/_default/config.yml
+++ b/config/_default/config.yml
@@ -38,6 +38,13 @@ module:
         - source: archetypes
           target: archetypes
 
+security:
+  funcs:
+    getenv:
+      - ASSET_BUNDLE_ID
+      - REPO_THEME_PATH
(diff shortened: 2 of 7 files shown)


config/_default/config.yml
```diff
@@ -38,6 +38,13 @@ module:
         - source: archetypes
           target: archetypes
 
+security:
+  funcs:
+    getenv:
+      - ASSET_BUNDLE_ID
+      - REPO_THEME_PATH
+      - PULUMI_CONVERT_URL
```
cnunciato, line 46: This already exists in the pulumi-hugo repo -- it was added with https://github.com/pulumi/pulumi-hugo/pull/971.

.gitignore
```diff
@@ -5,3 +5,4 @@ node_modules
 themes/*/resources
 public
 .DS_Store
+.hugo_build.lock
```
cnunciato, line 8: Just noticed this in this repo; will need to add to pulumi-hugo as well.

cnunciato: https://github.com/pulumi/pulumi-hugo/pull/984

susanev (approved):

PR #582, opened 2022-01-25
[ci] Change the target for the providers registry PR

stack72:
currently, the providers send a PR to the docs repo and need the provider
docs and registry repo checked out. With the change to use the
registrygen CLI tool, we can now point a provider update
directly at the registry repo and open the PR there


cnunciato (approved):
LGTM. Thanks for doing this!

PR #681, opened 2022-03-01
Add information about configuring access key for DigitalOcean's Spaces

praneetloke:
I got tripped-up by this when I tried to use just the DO token to provision a DO Bucket. I then came across https://github.com/pulumi/pulumi-digitalocean/issues/279. It would be nice to get this added to the docs as suggested in https://github.com/pulumi/pulumi-digitalocean/issues/279#issuecomment-995624192.


stack72 (approved):
Thanks @praneetloke <3

Issue #577, opened 2022-01-20
better communicate what packages are community authored packages


susanev:
## overview
right now, our packages all say who they are authored by, either Pulumi or an individual. If it is authored by an individual we consider it a community package. we want to better represent this information so its immediately clear to folks what packages are authored by our community. we could explore adjusting the current treatment or adding an additional design element to indicate this.

susanev:
@anita-trimbur id like you to take this on during the next iteration. ill fill you in on a few more details in our 1-1 today.

anita-trimbur:
## Constraints
- A new "community" designation cannot be treated as the existing tags. Existing tags are limited to type and use case.
- Relatedly, "community" cannot at this time be a filterable parameter.
- Prior discussion concludes that packages authored by Pulumi employees will be considered community contributions being that they are not officially supported.

## Terraform Registry Audit
- Color-coded badges to differentiate "Official" and "Verified" providers
- Absence of badge indicates community supplied providers
- Allows filtering by contribution type

anita-trimbur:
## [Figma File](https://www.figma.com/file/d8mG8QlCG7yNBVvEdrtM4K/Registry-Community-Tags?node-id=0%3A1)

## Design Changes
- Add icons and labeling for "Community" and "Pulumi Official" contributions to all Registry cards with tooltips
- Repeat icons and labeling in the header of each package page

![Group 4](https://user-images.githubusercontent.com/97129308/152837266-57cde246-6b87-49aa-a579-7f5206e1748e.png)
![Group 7](https://user-images.githubusercontent.com/97129308/152837277-03bdf168-d175-4229-b0b0-b3cf9eec8ce0.png)
![Group 5 (1)](https://user-images.githubusercontent.com/97129308/152837583-3c97b3d0-ac69-4d7f-bf64-d3428bc1d19a.png)

## Handoff Notes
- Community tooltip: "Package contributed by a community member"
- Pulumi Official tooltip: "Package maintained by Pulumi"

## Additional Notes
- The language "Pulumi Official" echoes language used in Terraform's Registry. The alternative "Pulumi Maintained" is also plausible verbiage.
- The use of a Community logo was discussed and decided against.

susanev:
@anita-trimbur these are great, thank you for doing this work!!

one thing to note is that there was some work done but not implemented to add in install information on the package pages in the spot where you have placed the official badge. i do not think you should make any changes right now. but we should revisit that placement when we do add in the install experience.

[...]

PR #693, opened 2022-03-02
Maintainer tags on packages

kimberleyamackenzie:
This PR adds tags on all packages (on both all package views and single package pages) to indicate whether a package is authored/maintained by the community or by Pulumi.  

Screen Shot 2022-03-02 at 9 45 59 AM
Screen Shot 2022-03-02 at 9 46 05 AM

Screen Shot 2022-03-02 at 9 46 49 AM
Screen Shot 2022-03-02 at 9 47 01 AM

fixes https://github.com/pulumi/registry/issues/577


themes/default/layouts/partials/registry/package.html
```diff
@@ -66,13 +72,38 @@ <h3 class="text-gray-900 no-anchor">
...
+                                {{ if (eq (.package.publisher) "Pulumi") }}
+                                    <img src="/icons/check-verified.svg">
+                                    </div>
+                                    <span slot="content">
+                                        Package maintained by Pulumi
+                                    </span>
+                                {{ end }}
+                    
+                                {{ if not (eq (.package.publisher) "Pulumi") }}
+                                <img src="/icons/people.svg">
```
cnunciato, line 98: I'd bump this in to make the nesting clear, as you have above.

themes/default/layouts/partials/registry/package.html
```diff
@@ -66,13 +72,38 @@ <h3 class="text-gray-900 no-anchor">
...
+
+                    <div class="maintainer flex items-center flex-grow justify-end mt-2 absolute bottom-8 h-14">
+                        <div class="text-right">
+                            <pulumi-tooltip>
+                                <div class="flex items-center pt-1">
+                                <span class="pr-2">{{ $packageLabel }}</span>
+                    
+                                {{ if (eq (.package.publisher) "Pulumi") }}
+                                    <img src="/icons/check-verified.svg">
+                                    </div>
```
cnunciato, line 91: Should this lone closing `div` be here like this? If so, maybe there's a better way to structure the markup such that this asymmetry wouldn't be necessary?

themes/default/layouts/partials/registry/package.html
```diff
@@ -66,13 +72,38 @@ <h3 class="text-gray-900 no-anchor">
...
 
                     <!-- Tags are anchored to the bottom of the card. -->
                     <ul class="package-tags mt-2 absolute bottom-8">
                         <li class="package-tag-type" data-filter-group="type" data-filter-value="{{ $packageType }}">{{ $packageTypeName }}</li>
                         <li class="package-tag-category" data-filter-group="category" data-filter-value="{{ $packageCategory }}">{{ $packageCategoryName }}</li>
                     </ul>
+
+                    <div class="maintainer flex items-center flex-grow justify-end mt-2 absolute bottom-8 h-14">
+                        <div class="text-right">
+                            <pulumi-tooltip>
```
cnunciato, line 85: Not a huge deal, but the tooltip seems to be floatin' off in space a bit:

image

themes/default/layouts/partials/registry/package/header.html
```diff
@@ -24,6 +24,38 @@
 
     <div class="lg:flex mb-2">
         <h1 class="text-4xl no-anchor">{{ with .GetPage $packageHome }}{{ .Title }}{{ end }}</h1>
+
+        {{ with (index $.Site.Data.registry.packages ($packageName)) }}
+            <div class="flex-grow justify-end">
+                {{ if (eq (.publisher) "Pulumi") }}
+                    <div class="flex lg:justify-end mt-2.5">
+                        <pulumi-tooltip>
+                            <div class="h-10 flex py-2 px-6 bg-blue-100 rounded-xl items-center">
```
cnunciato, line 33: Also a somewhat minorish thing, but with the blue background, rounded corners, and cursor on hover, this looks a lot like a clickable button.

cnunciato (approved):
Nice -- just a few minor things for you to consider.

themes/default/layouts/partials/registry/package/header.html
susanev: @anita-trimbur any thoughts on this?

anita-trimbur: That's a fair point, especially on account of the hover. That said, we do need some extra emphasis and separation from other text on the page to maintain the proper information hierarchy. We already use an underline on this page for selected tabs. Further reducing the opacity of that bg color could help mitigate the button association. @susanev, what do you think?

susanev: @anita-trimbur sure, id be good with reducing the opacity a bit. we actually have a similar problem in docs right now so id also be good shipping as is and revisiting the use of our primary color.

anita-trimbur:
Looking good! Just a few styling changes:

1.  Christian brings up the point that the bg colors the tags use on the package page could be mistaken for buttons. For now, let's try reducing the opacity of those bg colors from 10% down to 5% and see if that helps break association. 
2. Can we use Gilroy for the font on the tag itself?
3. I'm betting there's not a great way to do this, so it's not a huge deal if this is a no-go. But on the package page, can the tags be right-aligned with the right edge of the main container instead of the edge of that rightmost column?

Issue #558, opened 2022-01-14
Improve Search Results


jkodroff:
The search feature on the registry doesn't work the way I would expect. I searched for "azure", but I would've expected the first result to be the Azure Native and Classic providers, but I got a bunch of lower-level Azure docs. Maybe this could work similar to how Searchlight reports categorized results, e.g. searching for "azure":

* Providers:
    * Azure Native
    * Azure Classic
* (other categories of content TBD - I don't know the registry well enough to suggest something)

susanev:
@kimberleyamackenzie i think we may be able to tie this into the "improve docs user experience" initiative for this quarter. i believe @cnunciato has given you admin access to swifttype. let me know if you have interest in contributing to making this better.

kimberleyamackenzie:
@susanev yeah, for sure! definitely makes sense as part of that initiative + a great chance to get more familiar with swifttype generally for me

cnunciato:
Thanks for the help on this! Hit me up if you have any questions as you get going.

kimberleyamackenzie:
I have weighted the tags as described above.  A couple speedbumps we're running into:

- Using Azure as an example of something we have native and classic versions of:  with the weighted tags along, searching "azure" still didn't bring the package landing pages to the top.  This is because the package titles and overview language are actually "Azure Native" and "Azure Classic."  I solved this for our featured packages by customizing the search results and pinning the landing pages for classic + native to the top of the results.
- But, this is related to a larger challenge when titles on the page don't necessarily match the search terms folks might be using.  A good example of this is https://www.pulumi.com/registry/packages/azuread/.  If I search "AzureAD", with the new weighting, the landing page is the first result.  However, if I search "Azure active directory," the overview doesn't make it into the first page of results, because we don't use that full name in the tags or title.  We can work around this for one-off instances by setting synonyms in swiftype - but we should consider doing a pass to make sure we have all of these set where we want them to be, or otherwise update the meta descriptions to use full names.

I'm going to call this issue complete, and I have created https://github.com/pulumi/registry/issues/762 to dive deeper into the full name search issue.

PR #745, opened 2022-03-16
Add swifttype  body embedded data attribute tags to improve search

kimberleyamackenzie:
This PR adds some meta tags to registry package pages so that we can improve search experience, to address https://github.com/pulumi/registry/issues/558.

Swiftype docs: https://swiftype.com/documentation/site-search/crawler-configuration/meta-tags?size=n_10_n#embedded

### In this PR
With this PR, I added tags to the package title and to the overview description on the package landing page.  This by itself should have 0 impact - no visible changes in the UI, just the updated markup, and no impact on search.  

### Next steps
Based on all my doc-reading (please correct me if I'm wrong, swifttype is new to me!), I believe that the next steps should be adding weights to the two new tags.  The highest weight we currently have is 3.5 (https://app.swiftype.com/engines/pulumi-docs/document_types/page/custom_queries/65161/edit).  Given that, I propose:

- `package-overview-description`: 4.5.  Optimize the chances that searches for "Azure" or "aiven provider" give the landing page for the package as the top result.
- `package-title`: 4.  This title appears on all of the top-level package pages - installation, how-tos, etc. Optimize the chances these top-level pages are ranked higher than subpages ("how to do a hello-world").

### DIFF
--- a/themes/default/layouts/partials/registry/package/header.html
+++ b/themes/default/layouts/partials/registry/package/header.html
@@ -23,7 +23,7 @@
     {{ end }}
 
     <div class="lg:flex mb-2 max-w-5xl">
-        <h1 class="text-4xl no-anchor">{{ with .GetPage $packageHome }}{{ .Title }}{{ end }}</h1>
+        <h1 data-swiftype-name="package-title" data-swiftype-type="text" class="text-4xl no-anchor">{{ with .GetPage $packageHome }}{{ .Title }}{{ end }}</h1>
 
         {{ with (index $.Site.Data.registry.packages ($packageName)) }}
             <div class="flex-grow justify-end">


susanev (approved):
sounds great, thank you!

cnunciato:
@kimberleyamackenzie I ask because I haven't done this myself: How relevant is it that this content be wrapped in a `div` tag within the body of the page? I ask because in theory, since this is page-level metadata (it describes this whole page), it'd be slightly better/cleaner to capture it as page-level frontmatter and then use that frontmatter to populate both the page itself and the any Swiftype-specific `meta` tags. What do you think?

kimberleyamackenzie:
@cnunciato We do use some shortcodes on these pages already, but I for sure understand keeping them to a minimum.  I was looking to not break up the info (right now this content is inserted to the template all together, and at least for me, makes it easier to understand what comes from where and how it manifests on the page), but I can pivot to your suggested approach.  Could you point me to where we define the `schema.json` to update this?

cnunciato:
Caught up with @kimberleyamackenzie over Zoom (thanks!) and we're going to try using the frontmatter-param approach to see if we can get the outcome we want.

kimberleyamackenzie:
Based on @cnunciato's suggestion, I actually took out most of this PR and moved the logic over to hugo - I'll open that PR shortly, but in the meantime, this PR still touches one file, specifically adding the tag to the package title.  Tagging for re-review to look at that singular change.

susanev:
thanks yall for collaborating together on this!! 💜

cnunciato (approved):
🤜 🤛

PR #752, opened 2022-03-18
Bump headings a level in How-To Guides

cnunciato:
The README files we ingest from pulumi/examples are written such that the top-level  heading (the H1) is the title of the example itself, which works fine on GitHub, but not so well in the Registry, where these guides are situated within the context of a package (whose name currently "owns" the H1). Since we want at most one H1 per page, this Hugo Markdown-rendering hook shifts all headings on how-to guide pages down by one level, turning H1s into H2s, H2s into H3s, and so on.

## Before

image

## After

image

Fixes https://github.com/pulumi/marketing/issues/129.

Note that this change does not address any of the other work mentioned in https://github.com/pulumi/marketing/issues/129 (changing existing page headings, etc.). That work will be addressed separately in https://github.com/pulumi/marketing/issues/135.

### DIFF
--- a/themes/default/layouts/registry/_markup/render-heading.html
+++ b/themes/default/layouts/registry/_markup/render-heading.html
@@ -1,0 +1,18 @@
+{{/*
+    The README files we ingest from pulumi/examples are written such that the top-level
+    heading (the H1) is the title of the example itself, which works fine on GitHub, but not
+    so well in the Registry, where these guides are situated within the context of a package
+    (whose name is the H1). Since we want at most one H1 per page, this Hugo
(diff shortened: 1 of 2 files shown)


susanev:
@cnunciato would it make sense to get a preview with the how-to-guides from the examples repo with at least one of the packages? looks like only the hand-authored packages show up in the preview.

cnunciato:
@susanev Sure, I'll see if I can do that. There's no difference between how handwritten content and generated content are treated, though -- here's a local build of what it looks like with this branch and one of the generated packages:

image

susanev:
ah k, no worries then!

Brant-Soler (approved):
This looks good, thank you for requesting my review!

PR #794, opened 2022-03-30
Fix a typo in the DigitalOcean config docs

cnunciato:


stack72 (approved):

susanev:
@cnunciato i know its not a link, but should we use something other than xxx as the fake content in this?

PR #854, opened 2022-04-18
Add Pulumi HCP provider metadata

christophermaier:
Adds https://github.com/grapl-security/pulumi-hcp as a provider.

cc: @stack72


stack72:
Thank you so much for this @christophermaier

PR #856, opened 2022-04-18
Update layout for right navigation

kimberleyamackenzie:
This PR is the registry counterpart of https://github.com/pulumi/pulumi-hugo/pull/1386.

The main change here is to remove the right sidebar below the main content for API docs at any breakpoint below xxl (1536px) on package pages. Currently, that behavior happens below 1024px - by making this happen on larger screens, we're creating enough space to give the language chooser a little more room and accommodate additional language options. See https://github.com/pulumi/theme/issues/106 for more background.

### Pages affected by these changes
- Package API docs: (ex: https://www.pulumi.com/registry/packages/aws/api-docs/)
- Package installation, overview, and how-to-guides should be mostly unchanged, but now have a max width of 768px for the main content to meet accessibility guidelines: (ex: https://www.pulumi.com/registry/packages/aws/installation-configuration/)

### Changes you should see
- At and above 1536px, and below 1024px, layouts should not be changed.
- Below 1536px, you should see the right navigation disappear, as it currently does below 1024.
- The all package page should not be affected

### Screenshots + testing
A few screenshots below of the new below-1536 behavior , and I have manually tested against prod, but please do look at the preview to ensure there is no unexpected regressions and I haven't missed anything.

Screen Shot 2022-04-18 at 3 24 57 PM

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/pulumi/registry
 go 1.16
 
 require (
-	github.com/pulumi/pulumi-hugo/themes/default v0.0.0-20220416000535-57ea539e6d1e // indirect
+	github.com/pulumi/pulumi-hugo/themes/default v0.0.0-20220421012911-557e8debf301 // indirect
 	github.com/pulumi/registry/themes/default v0.0.0-20211011171710-45eb4e243ab7 // indirect
 	github.com/pulumi/theme v0.0.0-20220418160521-95bc5e47bca5 // indirect
 )
--- a/themes/default/data/registry/packages/aws.yaml
+++ b/themes/default/data/registry/packages/aws.yaml
@@ -11,5 +11,5 @@ publisher: Pulumi
 repo_url: https://github.com/pulumi/pulumi-aws
 schema_file_path: provider/cmd/pulumi-resource-aws/schema.json
 title: AWS Classic
-updated_on: 1650047579
-version: v5.1.3
+updated_on: 1650542448
+version: v5.1.4
--- a/themes/default/data/registry/packages/azure.yaml
+++ b/themes/default/data/registry/packages/azure.yaml
@@ -13,5 +13,5 @@ publisher: Pulumi
 repo_url: https://github.com/pulumi/pulumi-azure
 schema_file_path: provider/cmd/pulumi-resource-azure/schema.json
 title: Azure Classic
-updated_on: 1649692124
-version: v5.1.0
+updated_on: 1650384777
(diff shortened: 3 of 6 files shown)


themes/default/content/registry/packages/hcp/installation-configuration.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: HCP Setup
-meta_desc: Information on how to install the HCP provider.
+meta_desc: Information on how to install the HashiCorp Cloud Platform Provider for Pulumi.
```
kimberleyamackenzie, line 3: The changes to this file are fixes for linting errors that got merged from an unrelated PR: https://github.com/pulumi/registry/pull/854/files.  I just addressed the two errors to unblock my build (meta tag too short and missing blank lint).

sean1588 (approved):
This all seems to behave for me in the preview in the way you describe at the specified breakpoints. :tada:

susanev:
@anita-trimbur i am worried about the line lengths on the overview and install+config pages when we drop the right panel. accessibility best practices say 45-75 characters per line, and everyone will have trouble tracking those lines and keeping track of their position. these lines are 153 character+. some options to explore are adding a max-width on paragraphs, or increasing the font sizes of paragraphs when we drop the right panel so that fewer characters appear on a single line.

i would be okay not blocking on this if its a lot of extra effort but would want to do it soon after launch.

cc/ @kimberleyamackenzie

anita-trimbur:
I actually might have had a misunderstanding here. I was under the impression that this was scoped only to the API docs, where we have a left sidebar similar to Docs. I agree with @susanev about the line length and would support a max-width on the paragraphs if it's simple for @kimberleyamackenzie to implement.

anita-trimbur:
The pages in Registry where there is no left sidebar (Overview, Installation & Config, How-to-Guides) already exceed recommended line lengths on the website currently. If we were to introduce a new max-width, the main content area should never exceed 800px.

anita-trimbur:
Just for my own clarity: the most recent preview shows the max-width changes but not the sidebar change in API docs, but this PR does also remove the sidebar in API docs as shown in the previous preview?

kimberleyamackenzie:
@anita-trimbur I think the most recent preview does show the sidebar changes I would expect to see in API docs, let's maybe sync so I can understand what's missing?

anita-trimbur (approved):
This looks good to go! Kimberley and I ironed out my self-induced confusion and everything is as expected.

PR #878, opened 2022-04-24
Add workflow to check for community package updates

### LINKED ISSUE #853: Enhance ability to accept 3rd party packages
As part of the work to allow 3rd party packages to be provided to the registry we need to reduce the barrier to entry on getting the packages into the registry. Currently, we expect a user to generate a metadata file on each new version that needs published into the registry. We should remove this barrier. 

A user should be "on-board" their provider into the registry adding a small snippet of metadata to the registry repository via a PR. This is a 1 time job that ensures that we are aware of their package. At this time, we (Pulumi team) will need to ensure that the package is fit for purpose - we will discuss this in another issue

The user will need to give us the following metadata about the package:

* Repository URL (e.g. github.com/pulumi/pulumi-aws)
* Schema Path relative to the root of the repository

When the users PR to onboard has been accepted, we need to be able to create the appropriate registry metadata file for the package. We can determine the following from the package schema

* Name
* Category
* Publisher
* Description

Using the repository URL, we should be able to determine the version of the provider AND the date published. This will allow us to construct the registry schema file

On a cron (maybe 90 minutes?), we can then check all of the community providers for latest versions using a GHA

We can extend the registrygen cli tool to allow us to do this. We can make it do something like the following:

* registrygen check-new-version <repo slug> <schemafile>

Internally that cli command can load the existing registry metadata for the package. It can use the repo slug and the schema file location to use the GitHub API to be able to determine if the current version [in our registry] and in GitHub are different. if they are different then we can regenerate the package metadata and open a PR to our registry.

jkisk:
This workflow will run on a schedule to detect the release of new community package updates.  If a release differs from the registry version a PR will be generated with updated metadata.  This change also adds 3 community packages to the automation : checkly, eventstorecloud, and hcp.

Authors of new community packages will now only need to add their repo slug and schema file location to `package-list.json` to be included in the registry, by PR and with approval of a Pulumi maintainer.

Resolves #853

### DIFF
--- a/community-packages/generate_package_list.py
+++ b/community-packages/generate_package_list.py
@@ -2,8 +2,6 @@
 
 import json
 
-import json
-
 # Opening JSON file
 f = open('package-list.json')
 
--- a/.github/workflows/generate-package-metadata.yml
+++ b/.github/workflows/generate-package-metadata.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Sleep to prevent hitting secondary rate limits
         run: sleep 1
       - name: Install Registrygen CLI
-        uses: jaxxstorm/action-install-gh-release@v1.2.0
+        uses: jaxxstorm/action-install-gh-release@v1.5.0
         with:
           repo: pulumi/registrygen
       - name: Check for a new version


community-packages/package-list.json
```diff
@@ -1,5 +1,17 @@
...
+      "schemaFile": "provider/cmd/pulumi-resource-checkly/schema.json"
+    },
+    {
+      "repoSlug": "EventStore/pulumi-eventstorecloud",
+      "schemaFile": "provider/cmd/pulumi-resource-eventstorecloud/schema.json"
+    },
+    {
+      "repoSlug": "grapl-security/pulumi-hcp",
+      "schemaFile": "provider/cmd/pulumi-resource-hcp/schema.json"
+    },
```
guineveresaenger, line 14: Should we add buildkite here as well? https://github.com/pulumi/registry/pull/874

jkisk: We should! But I think after we merge that PR.

community-packages/generate_package_list.py
```diff
@@ -1,3 +1,5 @@
+#! /usr/bin/env python3
+
 import json
```
stack72, line 3: I think I have dupe imports here - can you clean them up?

stack72 (approved):
Can you update the version of jaxxstorm GHA to help with the retry of installing a package and then we can ship

community-packages/generate_package_list.py
jkisk: sure thing!

PR #881, opened 2022-04-25
Use registrygen tool and link to repository

guineveresaenger:
For ad-hoc registry docs gen, we want to use the new and shiny
https://github.com/pulumi/registrygen. This PR updates that reference in
the README.

Part of https://github.com/pulumi/platform-providers-team/issues/24.

I notice we also use `resourcedocsgen` in the Makefile for this repo,
which is something to follow up on, unless I'm mistaken in our desire to
phase out `resourcedocsgen`.


stack72 (approved):
So happy to see that! thank you :)

PR #886, opened 2022-04-25
Add buildkite and use updated registrygen command

jkisk:
Adds a new community package to the list for automated updates and modifies the call to `registrygen pkgversion` to take a repo slug instead of an owner flag. Do not merge until https://github.com/pulumi/registrygen/pull/8 is merged and released.

### DIFF
--- a/.github/workflows/generate-package-metadata.yml
+++ b/.github/workflows/generate-package-metadata.yml
@@ -26,10 +26,6 @@ jobs:
       max-parallel: 1
       matrix: ${{fromJson(needs.generate-packages-list.outputs.matrix)}}
     steps:
-      - name: Provider RepoSlug
-        run: echo ${{ matrix.repoSlug }}
-      - name: SchemaFile
-        run: echo ${{ matrix.schemaFile }}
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-package-metadata.yml
```diff
@@ -30,13 +30,6 @@ jobs:
         run: echo ${{ matrix.repoSlug }}
       - name: SchemaFile
```
stack72, line 31: We don't need these echo statements - this was purely for initial testing that my matrix worked :)

.github/workflows/generate-package-metadata.yml
```diff
@@ -30,13 +30,6 @@ jobs:
...
         run: echo ${{ matrix.schemaFile }}
-        # Isolate publisher and provider short names from repo slug.
-      - name: Get publisher
```
stack72, line 34: Love that we don't need to regex!!

stack72 (approved):

guineveresaenger (approved):

PR #901, opened 2022-05-03
fix slack logo

susanev:
adjusting this so it uses the slack logo that's over in hugo
https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/static/logos/pkg/slack.svg

this doesn't have the surrounding whitespace so it grows to fit the space

im not sure if im editing this in the right place
feel free to direct me elsewhere

### DIFF
--- a/themes/default/data/registry/packages/slack.yaml
+++ b/themes/default/data/registry/packages/slack.yaml
@@ -2,7 +2,7 @@ category: Utility
 component: false
 description: A Pulumi package for managing Slack workspaces.
 featured: false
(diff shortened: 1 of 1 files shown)


stack72 (approved):

stack72:
We really need to come up with a better strategy here - we provider that LogoUrl so we should not need to ask you to do this :/

susanev:
@stack72 the logo url that was provided has extra white space around it, and it is missing the word mark. i also want this to better, but in the meantime, im trying to keep things in a good state until that happens.

PR #902, opened 2022-05-03
update databricks logo

susanev:
the logo url that was used uses white text, which is invisible in the registry.
this removes that URL and uses the logo that was added to hugo
https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/static/logos/pkg/databricks.svg

i have no idea why `logo_url: ""` is not working for this specific package
so using the direct url for it

### DIFF
--- a/themes/default/data/registry/packages/databricks.yaml
+++ b/themes/default/data/registry/packages/databricks.yaml
@@ -2,7 +2,7 @@ category: Infrastructure
 component: false
 description: A Pulumi package for creating and managing databricks cloud resources.
 featured: false
-logo_url: ""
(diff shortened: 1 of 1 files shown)


themes/default/data/registry/packages/databricks.yaml
```diff
@@ -2,7 +2,7 @@ category: Infrastructure
 component: false
 description: A Pulumi package for creating and managing databricks cloud resources.
 featured: false
-logo_url: https://databricks.com/wp-content/uploads/2021/10/db-nav-logo.svg
+logo_url: ""
```
susanev, line 5:
```suggestion
logo_url: "https://raw.githubusercontent.com/pulumi/pulumi-hugo/a56a9ec047a1257e3ca7d49588a86286fe0214c9/themes/default/static/logos/pkg/databricks.svg"
```

PR #916, opened 2022-05-07
adjusting awsx name+title

susanev:


stack72 (commented):
This will get overwritten - we shouldn't change these files

susanev:
Yah I closed it; let's talk sometime about the names of these things that appear in the registry.

Issue #834, opened 2022-04-08
edit the page links on overview and install+config pages 404


susanev:
edit the page links on overview and install+config pages 404
they send you to hugo instead of registry

may be a regression from 
https://github.com/pulumi/registry/pull/800

susanev:
@cnunciato i totally forgot about this issue, do you know if there is a quick fix that we could do for this now? im honestly more interested in these working than the how-to guides if i had to pick one. could we maybe revert #800 until we have more time?

cnunciato:
> im honestly more interested in these working than the how-to guides if i had to pick one.

Totally agree. I don't think I'll have time to do actually this this weekend, but if you do, I'd +1 a change that made the Edit links work on the Overview and Install & Config pages and disabled it on the How-to Guides for now.

PR #917, opened 2022-05-07
removing edit this page links

susanev:
proposing removing the edit this page links until we have time to fix the 404 issue
tracked in https://github.com/pulumi/registry/issues/834


cnunciato (approved):
👍 as a fix the broken links, but would you mind filing an issue to re-enable them on the Overview and Install pages? Thanks!

Issue #717, opened 2022-03-09
hovering over the pulumi/community badge icon does not show tool tip


susanev:
## overview

if you hover over the pulumi/community badge icon it does not show tool tip, it only displays when you hover over the text. we should treat these as a unit and display a tool tip no matter what part you are hovering over. id imagine that more folks are hovering over the icon than the text to get more information, and currently that does not display anything.

susanev:
@anita-trimbur let me know if theres context around this behaviour that i may be missing

PR #945, opened 2022-05-19
Update tooltip behavior on list view

kimberleyamackenzie:
This PR updates the tooltip behavior for the text/icon indicating whether a package is Pulumi-maintained or community-maintained.  Previously, the tooltip would only show when hovering the text on the card.  This PR makes it so the text shows on hover of either text or the icon image.  The individual package page already had this behavior, so this fix was only needed in the list view.

Ticket: https://github.com/pulumi/registry/issues/717

### Verification
Some screenshots below, but you can't see my cursor, so the best bet is to visit `/registry/` at the preview link to try it out for yourself!

Screen Shot 2022-05-19 at 11 04 34 AM
Screen Shot 2022-05-19 at 11 08 41 AM

### DIFF
--- a/themes/default/layouts/partials/registry/package.html
+++ b/themes/default/layouts/partials/registry/package.html
@@ -85,24 +85,24 @@
                             {{ if (eq (.package.publisher) "Pulumi") }}
                                 <div class="flex items-center pt-1">
                                     <pulumi-tooltip>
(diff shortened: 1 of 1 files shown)


susanev (approved):
thanks!

sean1588 (approved):
:ship:

anita-trimbur (approved):

PR #954, opened 2022-05-23
[internal] Fix up fail-fast in provider metadata

stack72:
As this job creates multiple PRs, we don't want to block other providers on a single provider failure


danielrbradley (approved):
Approved - so if a single provider fails we still want to generate other providers PRs.

PR #1025, opened 2022-06-03
Add AWS IAM Component

zchase:
This PR adds the overview pages and the data file for the AWS IAM package. It wasn't clear to me how the `api-docs` are generated in the pipeline but my assumption was it was generated at the `docs` level during the build. Please let me know if there are additional steps I need to take to get the docs published!

### DIFF
--- a/themes/default/data/registry/packages/aws-iam.yaml
+++ b/themes/default/data/registry/packages/aws-iam.yaml
@@ -4,7 +4,7 @@ description: ""
 featured: false
 logo_url: ""
 name: aws-iam
-native: true
+native: false
 package_status: public_preview
 publisher: Pulumi
 repo_url: https://github.com/pulumi/pulumi-aws-iam


themes/default/data/registry/packages/aws-iam.yaml
```diff
@@ -0,0 +1,14 @@
+category: Cloud
+component: true
+description: ""
+featured: false
+logo_url: ""
+name: aws-iam
+native: true
```
stack72, line 7: This shouldn't be a native - you are missing keywords - see aws-apigateway for the ones you want

stack72 (commented):
1 small change needed @zchase

susanev:
@zchase heres a logo for you: https://github.com/pulumi/pulumi-hugo/pull/1582

themes/default/data/registry/packages/aws-iam.yaml
zchase: @stack72 update the `native` value but not sure what else you're referring to. I compared it to the aws-apigateway data file and they look similar now.

stack72 (approved):

PR #1027, opened 2022-06-05
fix link for event store cloud access tokens

susanev:
the current link requires signing in
this links to the docs where the steps are described


stack72 (approved):
@susanev this is coming from their repo - we can't only change it here in this repo, we also need to fix it there or it will come back with the next publish https://github.com/EventStore/pulumi-eventstorecloud/tree/main/docs

Issue #910, opened 2022-05-04
Typo in the Confluent Cloud example


krisajenkins:
[themes/default/content/registry/packages/confluent/_index.md](https://www.pulumi.com/registry/packages/confluent/)

Minor typo I just spotted. The import (`ccloud`) doesn't match the usage (`confluent.ConfluentEnvironment`). So this code

```typescript
import * as confluent from "@pulumi/confluent";

const env = new ccloud.ConfluentEnvironment("ts-environment");
```

Should read:

```typescript
import * as confluent from "@pulumi/confluent";

const env = new confluent.ConfluentEnvironment("ts-environment");
```

PR #1029, opened 2022-06-05
Adding the new Confluent Official Provider

stack72:
Fixes: #910 

This is a different provider to the confluent provider


stack72:
I need to self merge this - the metadata was merged automatically and without this, it will cause the site to break

PR #1053, opened 2022-06-09
add aws static website package

sean1588:
This PR adds the AWS static site package. 

Re-opening this since it had a bu that caused the pipeline to fail yesterday because the schema file path was incorrect. I have update the file path now.

### DIFF
--- a/themes/default/data/registry/packages/aws-iam.yaml
+++ b/themes/default/data/registry/packages/aws-iam.yaml
@@ -1,6 +1,6 @@
 category: Cloud
 component: true
-description: ""
+description: "A Pulumi component to deploy a static website to AWS"
 featured: false
 logo_url: ""
 name: aws-iam


themes/default/data/registry/packages/aws-static-website.yaml
```diff
@@ -0,0 +1,14 @@
+category: Cloud
+component: true
+description: ""
```
cnunciato, line 3: Should we have a description here? Looks like other packages have one.

cnunciato (approved):
This PR is good, but I think we need some follow-up additions to the aws-static-website repo to fix up the rendering of this section, which is missing some fields:

image


I think this is happening because `schema.json` is missing the `repository` and `license` fields. (See here for an example from pulumi/eks: https://github.com/pulumi/pulumi-eks/blob/f60263061a85d2817ecd0fd4c3dbc6505e4a8df9/provider/cmd/pulumi-resource-eks/schema.json#L10-L11)

themes/default/data/registry/packages/aws-static-website.yaml
sean1588: just added one. thanks. I'll get those license fields updated in the provider repo.

Issue #1052, opened 2022-06-09
align names on browse and package pages


susanev:
sometimes the names we list on the browse page are different than the package page
we should align these

some examples:
- elastic is ec on the browse page and ElasticCloud (EC) on the package page
- artifactory is artifactory on the browse page and Artifactory on the package page
- astra is astra on the browse page and Astra on the package page
- awsx is awsx on the browse page and AWSx on the package page
- miniflux is Miniflux on the browse page and AWS Miniflux on the package page

questions:
- why do we have different metadata for these things, can they be the same?

related issues:
- https://github.com/pulumi/registry/issues/839

lukehoban:
I do agree things are fairly "weird" here today.  

A few thoughts:
1. Should we require a display name for all packages?  If not, should we just *only* use the raw package name for all packages?
2. If no to both of the above, should we always show the raw package name in addition to display name when it is available?
3. Is there a 3rd piece of metadata?  How is it that Elastic Cloud is named named "ec" on the index page but "ElasticCloud (EC)" on the description page?

susanev:
for the most part this is now done. we still don't have a single source of truth for this, but the two pieces of data is now fixed.

for anyone else who comes across this...
- browse page uses display name for the package if it exists, otherwise it uses the package name from the meta data; but there are exceptions here: https://github.com/pulumi/pulumi/blob/master/pkg/codegen/docs/gen.go#L57-L152 which controls the name that shows on the browse page
- for the package page, the name comes from the title in the meta data which is in this repo

PR #1062, opened 2022-06-13
align miniflux names on the browse and package pages

susanev:
this brings the miniflux names on the package and browse page in alignment with each other

part of https://github.com/pulumi/registry/issues/1052

### DIFF
--- a/themes/default/data/registry/packages/aws-miniflux.yaml
+++ b/themes/default/data/registry/packages/aws-miniflux.yaml
@@ -9,6 +9,6 @@ package_status: public_preview
 publisher: Christian Nunciato
 repo_url: "https://github.com/pulumi/pulumi-aws-miniflux"
 schema_file_path: /schema.json
-title: Miniflux
+title: AWS Miniflux
 updated_on: 1634334277
 version: v0.1.0


cnunciato (changes requested):
This current package name is correct; it’s not an AWS service, it just happens to be written to run on AWS. (The other pages should be changed to reflect this.)

It does raise an interesting question though of what sort of convention we should use for these kinds of components. Probably something we should discuss.

cnunciato:
> @cnunciato what do you mean the other pages? the package page heading is currently AWS Miniflux, this is wrong? would Miniflux on AWS work for you?
> 
> As a note, almost all the other components are named in the format the current Miniflux package page is named.

Yep, this is why I bring it up. I'm not sure [Cloud] [Package] is the best display-naming convention for packages that aren't tied to a provider-specific offering. This convention definitely works for packages that _are_ tied to a provider-specific product, like AWS IAM, AWS API Gateway, EKS, etc., as "IAM", "API Gateway", and so on are all AWS products. But it doesn't quite work for things like Miniflux or "static website", applications or concepts exist separate from any one provider. 

Particularly when listed alongside packages like AWS EKS and the like, "AWS Miniflux" reads a bit like it could be an AWS managed service; similarly, "Azure Justrun" -- I had to go digging to find out whether "Justrun" was an Azure service or some other, more generic concept implemented for deployment on Azure. For generic/non-provider-specific things like these, we should probably consider an alternative display-naming convention. Your suggestion of "Miniflux on AWS" (or "Static website on AWS") seem clear to me, yeah, so that could be one way to go. Another might be to let display names just remain provider agnostic as they are now, which would allow us to list multiple "Miniflux" packages on the same page with a clear distinction of some sort to distinguish them from one another cloud-wise.

cc @lukehoban @stack72 who may have thoughts & opinions on this as well. (I don't have strong opinions myself other than wanting to keep things as clear and unambiguous as we can.)

susanev:
@cnunciato can you open up a new issue instead related to component naming? seems a bit weird to include that In this pr which was only trying to fix the misaligning between the package name and the name shown on the browse page. ill close the pr since you don't support it.

cnunciato:
Sure thing. Meantime, here's a PR that I think accomplishes what you were after with this one (i.e., align the display names for this package): https://github.com/pulumi/registry/pull/1072

PR #1072, opened 2022-06-14
Fix Miniflux display naming

cnunciato:
Makes all the display names match the package title (and fixes a couple of typos & things). (See https://github.com/pulumi/registry/pull/1062 for discussion.)


susanev (approved):
@cnunciato can you pls open an issue related to your thoughts on component names? im glad these are getting aligned, but now this component is misaligned from the rest of the components. i think we should work through that and figure out a plan moving forward.

cnunciato:
>can you pls open an issue related to your thoughts on component names

Yep, do.

PR #1276, opened 2022-07-21
Add docs for Metabase Component

zchase:
This PR adds the docs for a [Metabase component](https://github.com/pulumi/pulumi-metabase).


sean1588:
@zchase this is awesome!! just one thing, should we get a graphic for the metabase logo added here?

zchase:
@sean1588 I added it in this PR: https://github.com/pulumi/pulumi-hugo/pull/1765

I believe that is the way to get the image on but the preview here was probably generated before the PR made it in. Let me see if I can cut another preview.

sean1588 (approved):

PR #1301, opened 2022-07-28
(docs) Remove unused imports in Kubernetes example

covik:
Last two imports are unused and could mislead juniors


stack72 (approved):
thanks so much for this @Covik - appreciate the fixups!!

PR #1339, opened 2022-08-11
Adding a note to the AWSx docs to suggest it's for beta version

stack72:


cnunciato (approved):
Thanks! 

Do we need to update anything in the [Crosswalk docs](https://www.pulumi.com/docs/guides/crosswalk/aws) similarly? I seem to remember people hitting the same issues there, but could be wrong.

stack72:
> Thanks!
> 
> Do we need to update anything in the [Crosswalk docs](https://www.pulumi.com/docs/guides/crosswalk/aws) similarly? I seem to remember people hitting the same issues there, but could be wrong.

Yeah PR for that coming

PR #1405, opened 2022-08-26
feat: Add reference to pulumi-time provider in Pulumiverse

tmeckel:
This PR will add the pulumi-time provider from Pulumiverse to the registry


stack72:
Thank you @tmeckel :)

stack72 (approved):

PR #1432, opened 2022-08-31
feat: add the gandi provider

rawkode:


stack72:
> The Gandi provider doesn't show up in the preview

This is only adding the Gandi provider to the automation flow that will generate and add it - that will run in a coule of hours

PR #1514, opened 2022-09-17
added location for configuration option

mr-j90:
I added the needed edits for the location configuration option for the `location.` I also added where to add them from the environment variables standpoint; please let me know if I am missing anything.

#1604
------------
I was looking under "Installation Configuration" to figure out how to solve the following error:
error: Code="LocationRequired" Message="The location property is required for this definition."

For other providers, like AWS, this setup is listed under Configuration Options under the "Installation & Configuration" tab.

We should include this option for the configuration setup for this provider as well.

The docs live here:

https://github.com/pulumi/registry/tree/master/themes/default/content/registry/packages/azure-native


susanev (approved):
thank you!!

thomas11:
Thank you for the contribution, @JustJordanT! I don't think it's 100% correct, though. As the existing docs state, this location setting _will only be used if one was not supplied explicitly_. I have a sample Pulumi program here that does not use this setting and it works, because it passes the `Location` argument when constructing new Azure resources.

In my view, the existing docs are correct. Did you encounter a situation where setting this location config was the only way to make a program work?

themes/default/content/registry/packages/azure-native/installation-configuration.md
```diff
@@ -111,6 +111,7 @@ Remember to pass `--secret` when setting `clientSecret` so that it is properly e
     $ pulumi config set azure-native:clientSecret <clientSecret> --secret
     $ pulumi config set azure-native:tenantId <tenantID>
     $ pulumi config set azure-native:subscriptionId <subscriptionId>
+    $ pulumi config set azure-native:location <locationName>
```
thomas11, line 114:
```suggestion
    # optional default location, otherwise set in code
    $ pulumi config set azure-native:location <locationName>
```

themes/default/content/registry/packages/azure-native/installation-configuration.md
```diff
@@ -157,7 +161,7 @@ Use `pulumi config set azure-native:<option>` or pass options to the [constructo
...
 | `environment`               | Required          | The cloud environment to use. It can also be sourced from the ARM_ENVIRONMENT environment variable. Supported values are: `public` (default), `usgovernment`, `china`.                                                                                                                                 |
-| `location`                  | Optional          | The location to use. ResourceGroups will consult this property for a default location, if one was not supplied explicitly.                                                                                                                                                                             |
+| `location`                  | Optional          | The location to use. ResourceGroups will consult this property for a default location, if one was not supplied explicitly. This must be set in either the resource or the environment vars                                                                                                                                                                        |
```
thomas11, line 164:
```suggestion
| `location`                  | Optional          | The location to use. ResourceGroups will consult this property for a default location, if one was not supplied explicitly when defining the resource. It can also be sourced from the ARM_LOCATION_NAME environment variable.                                      
|
```

JustJordanT:
> Hi @JustJordanT , I added some suggestions. Overall it looks great, thanks again!

Thanks for the direction, I was mean to circle back on this!

thomas11 (approved):

PR #1558, opened 2022-10-04
Add self signed cert package

zchase:
This PR adds the files needs to list the [Self Signed Cert Package](https://github.com/pulumi/pulumi-tls-self-signed-cert) in the registry.

### DIFF
--- a/themes/default/data/registry/packages/self-signed-cert.yaml
+++ b/themes/default/data/registry/packages/self-signed-cert.yaml
@@ -3,7 +3,7 @@ component: true
 description: A Pulumi component that creates a self-signed certificate.
 featured: false
 logo_url: ""
-name: self-signed-cert
+name: tls-self-signed-cert
 native: false
 package_status: public_preview
 publisher: Pulumi


themes/default/data/registry/packages/self-signed-cert.yaml
```diff
@@ -0,0 +1,14 @@
+category: Cloud
+component: true
+description: A Pulumi component that creates a self-signed certificate.
+featured: false
+logo_url: ""
+name: self-signed-cert
```
sean1588, line 6:
```suggestion
name: tls-self-signed-cert
```
this is causing the "view source code" link to 404 since it is looking for a different repo name

sean1588 (approved):
awesome!! :raised_hands: 

just one comment about fixing the name... other than that lgtm!! :ship:

themes/default/data/registry/packages/self-signed-cert.yaml
zchase: Ah good catch, thanks!!

PR #1607, opened 2022-10-20
Update generate-package-metadata.yml

jkodroff:


jkodroff:
Job failing before this change: https://github.com/pulumi/registry/actions/runs/3290465587/jobs/5423351372
Job succeeding after this change: https://github.com/pulumi/registry/actions/runs/3290681431

PR #1643, opened 2022-10-28
Fixing the import library names in the Confluent Cloud examples.

krisajenkins:
There's a confusion here between importing the library as `confluent` and then using it as `ccloud`.

Note: I've fixed the TypeScript and Python examples. I _suspect_ the C# one is wrong too, but I don't know enough C# to change it confidently.


ringods (approved):

ringods:
Tnx @krisajenkins

PR #1669, opened 2022-11-04
Update installation-configuration.md

jooooel:
When following the instructions about how to set the apiToken alongside the Pulumi stack, I received this error: `error: config value for 'cloudflare:apiToken' looks like a secret; rerun with --secret to encrypt it, or --plaintext if you meant to store in plaintext`

So the docs should probably mention that as well I suppose 😊


susanev (approved):
thx!

PR #1710, opened 2022-11-21
Update AWSX version manually

t0yv0:


danielrbradley (approved):

danielrbradley:
I'm seeing the version in the preview correctly, but the API tab doesn't work ... guessing this is handled somewhere else in the pipeline?

t0yv0:
http://registry-origin-pr-1710-23a3f43b.s3-website.us-west-2.amazonaws.com/registry/packages/awsx/ here right? I cannot click API docs tab either. 

I'm actually not sure. Perhaps someone from @pulumi/content-engineering can help evaluate if this is ok to merge or something further needs fixing. 🙇

ringods:
@t0yv0 as far as I know the build pipelines, the API doc generation is indeed done elsewhere. I have never seen API docs generated in these preview sites for Pulumiverse providers. I would merge this given it is the same type of metadata change as happens for every other provider version bump.

PR #1720, opened 2022-11-23
Fix configuration keys in Alicloud installation instructions

guineveresaenger:
A user reported not being able to configure access keys using the provider's configuration instructions.
This change provides the correct configuration keys.
Fixes https://github.com/pulumi/pulumi-alicloud/issues/326.


guineveresaenger:
Actually, this is not necessary. Closing this PR.

PR #1724, opened 2022-11-26
Update PyPI link with correct link for Pulumi AWS EKS Components.

AdrianM10:
The current link has an additional forward slash (/) resulting in the user being greeted with an error 404 page on the PyPI website.


susanev (approved):
thank you!

PR #1733, opened 2022-11-28
add related resource links

sean1588:
do not merge - these PRs are still a work in progress

theme PR: https://github.com/pulumi/theme/pull/383
see theme PR for discussion.


zchase (approved):
LGTM --- Do we need the `go.mod` and `go.sum` changes? Though not sure if it really matters since I think at the end of day the get rewritten by the time it hits production

Issue #1704, opened 2022-11-18
Restyle package page navigation


anita-trimbur:
## overview

as part of our registry epic, we're making style changes to the navigation tabs on our package pages for a clearer hierarchy. 

## designs

**note** designs show material icons, but we'll be using font awesome icons in the registry. listing below the icon equivalents for findability.

* font size 16px
* bold selected tabs
* gray 300 line underneath menu
* unselected tabs 
  * gray 600 text
* selected tabs 
  * blue 100 background
  * primary blue text
  * top bottom padding 6px
  * left right padding 16px
  
#### font awesome icon names

overview: `file-lines`
installation and configuration: `gear`
api docs: `book`
how-to guides: `circle-question`

the designs have a filled icon when selected and an outlined icon when unselected. i'm pretty sure that we don't have access to most outlined icons in font awesome. if this is the case, **we should use all filled icons.**

<img width="808" alt="registry nav" src="https://user-images.githubusercontent.com/97129308/202811002-d856e56c-58dd-4739-a282-d9f12ec52656.png">

PR #1746, opened 2022-11-30
Update styles for registry navigation

kimberleyamackenzie:
Fixes: https://github.com/pulumi/registry/issues/1704

Update the styles on Registry package page navigation.

Screen Shot 2022-11-30 at 10 23 02 AM
Screen Shot 2022-11-30 at 10 23 09 AM
Screen Shot 2022-11-30 at 10 23 16 AM

### Responsive behavior
Screen Shot 2022-11-30 at 10 49 38 AM
Screen Shot 2022-11-30 at 10 50 22 AM
Screen Shot 2022-11-30 at 10 50 45 AM
Screen Shot 2022-11-30 at 10 50 51 AM

### DIFF
--- a/themes/default/layouts/partials/count-pill.html
+++ b/themes/default/layouts/partials/count-pill.html
@@ -1,4 +1,0 @@
-{{ $textColor := cond .selected "text-blue-600" "text-gray-600" }}
-{{ $borderColor := cond .selected "border-blue-600" "border-gray-600" }}
-{{ $backgroundColor := cond .selected "bg-violet-600" "" }}
-<span class="px-2 ml-2 rounded-full text-sm border bg-transparent font-normal {{ $textColor }} {{ $borderColor }}">{{ .count }}</span>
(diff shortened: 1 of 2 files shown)


anita-trimbur:
maybe i'm just hyper fixated on this, but does the change from medium to bold on selection make the page look jumpy? would it look more fluid if we keep the font weight medium even for the selected tabs?

kimberleyamackenzie:
@anita-trimbur personally i think the page looks jumpy even on prod, regardless of the bold change - if you give it a look there, what do you think?

anita-trimbur:
huh i guess it does do that already, doesn't it? yeah you're right, let's leave it as is!

anita-trimbur (approved):

susanev:
@anita-trimbur ive been meaning to ask you or @jonelleb to add not using font-weight to indicate state change to the design decisions doc because we have ran into this a few times before and it always causes issues. id drop the weight change from this, even if there are other issues on the page.

kimberleyamackenzie:
@anita-trimbur Here's that updated behavior:

- [ ] Default text color to gray 800
- [ ] Hover text color is blue 600
- [ ] No switch to bold for selected tabs


https://user-images.githubusercontent.com/25756367/205183887-1e2034b3-3380-4b59-a3e3-03e17e1950f1.mp4

PR #1755, opened 2022-12-01
fix some titles

susanev:
so im like 20% this will work and not be overwritten
but lets try it and then try something else if it doesn't work


kimberleyamackenzie (approved):

susanev:
gonna try this first instead
https://github.com/pulumi/pulumi/pull/11506

PR #1757, opened 2022-12-01
Update Registry breadcrumb behavior

kimberleyamackenzie:
Update the Registry breadcrumb to appear on all package pages (not just API docs), and fix behavior regression with duplicated package name in breadcrumb.

Get ready for some long explanation!

### What's happening now in prod
Currently, in prod, there's some unexpected behavior for the breadcrumb visible on a package's API docs.  The original implementation of this breadcrumb was to _only_ show the package name (`My great package`) which would link to the api docs index page, and then any subsequent parts of the docs tree (`Super cool resource`), so the end user would see `My great package / Super cool resource`.

However, in mid-September, a change shipped to pulumi-hugo that added a breadcrumb for a different set of pages, but the template block shared the same name as the one we use in the Registry breadcrumb.  This resulted in Registry unexpectedly inheriting the behavior from the Hugo breadcrumb, instead of its own.  Users now see something like `Pulumi Registry / My great package / My great package / Super cool resource`, where the first `My great package` links to the package's overview page, and the second `My great package` links to the API docs index for that package.  Both have front matter titles of `My great package`, so they repeat.

### What this PR does
- [ ] Renames the template block for this breadcrumb, so it's no longer overridden by the unrelated breadcrumb in pulumi-hugo
- [ ] Updates the behavior of this breadcrumb, because we actually _want_ most of the regressed behavior - showing the full path.
- [ ] What we don't want is the duplicated names, so this hard-codes some logic to show "API Docs" in the breadcrumb if the page url is the api docs index page
- [ ] I looked into whether that could be addressed at the docs gen level instead, by updating the page title, but there is not a straightforward way to do this

If you poke around the preview, you should find breadcrumbs that replace the `All packages` link at the top of the page.  There should only be the single breadcrumb in the header, no sub pages should have duplicated breadcrumbs.

Screen Shot 2022-12-01 at 2 28 44 PM
Screen Shot 2022-12-01 at 2 28 49 PM
Screen Shot 2022-12-01 at 2 28 56 PM

### DIFF
--- a/themes/default/layouts/partials/registry/package/header.html
+++ b/themes/default/layouts/partials/registry/package/header.html
@@ -12,7 +12,7 @@
 
 <div class="container">
     <div class="lg:flex mb-6">
-        {{ partial "registry/package/how-to-guides/breadcrumb.html" . }}
+        {{ partial "registry/package/breadcrumb.html" . }}
     </div>
 
     {{ with (index $.Site.Data.registry.packages ($packageName)) }}
--- a/themes/default/layouts/partials/registry/package/breadcrumb.html
+++ b/themes/default/layouts/partials/registry/package/breadcrumb.html
@@ -1,4 +1,3 @@
-   
 <!-- Based on https://gohugo.io/content-management/sections/#example-breadcrumb-navigation -->
 
 <ol class="breadcrumb">


themes/default/layouts/partials/registry/package/how-to-guides/breadcrumb.html
```diff
@@ -1,20 +1,26 @@
...
+        {{ template "registrybreadcrumbnav" (dict "p1" .p1.Parent "p2" .p2) }}
+    {{ end }}
+    <!-- If the page is a redirect page, don't include it in the breadcrumb. -->
+    {{ if not (isset .p1.Params "redirect_to") }}
+        {{ $lastUrlSlug := index (last 1 (split (delimit (split .p1.RelPermalink "/") "," "") ",")) 0 }}
+        <li {{ if eq .p1 .p2 }}class="active"{{ end }}><a data-track="breadcrumb" href="{{ .p1.RelPermalink }}">
+            <!-- If we're showing the API Docs, use that as the breadcrumb, not the title - which
+            is a repeat of the package name. -->
+            {{ if eq $lastUrlSlug "api-docs" }}
+                API Docs
```
sean1588, line 20: good idea!

sean1588 (approved):
this looks good and works well for me!!

:shipit:

themes/default/layouts/partials/registry/package/header.html
```diff
@@ -11,9 +11,8 @@
 {{ end }}
 
 <div class="container">
-    <div class="mb-8 text-blue-600 font-display">
-        <span>&larr;</span>
-        <a href="{{ relref . "/registry" }}" class="hover:underline">All packages</a>
+    <div class="lg:flex mb-6">
+        {{ partial "registry/package/how-to-guides/breadcrumb.html" . }}
```
cnunciato, line 15: Bit of a nit, but should this partial just be moved to`registry/package/breadcrumb.html`, since it's the same across all pages now (and not specific to `how-to-guides`?

kimberleyamackenzie: It wasn't specific to how to guides previously either (it was used in API docs too), but good call that now's a great time to move it, will do!

PR #1787, opened 2022-12-05
Update Registry breadcrumb

kimberleyamackenzie:
Basically https://github.com/pulumi/registry/pull/1757/ all over again - that PR automatically closed when its upstream branch got merged.  This accomplishes the same thing, with one small addition - hardcoding the Installation & Configuration breadcrumb to `Installation & Configuration` instead of `Package Name Installation & Configuration`.  

Screen Shot 2022-12-05 at 9 34 51 AM
Screen Shot 2022-12-05 at 9 35 00 AM
Screen Shot 2022-12-05 at 9 35 13 AM
Screen Shot 2022-12-05 at 9 35 21 AM

### DIFF
--- a/themes/default/layouts/partials/registry/package/breadcrumb.html
+++ b/themes/default/layouts/partials/registry/package/breadcrumb.html
@@ -19,6 +19,8 @@
                 API Docs
             {{ else if eq $lastUrlSlug "installation-configuration" }}
                 Installation & Configuration
+            {{ else if eq $lastUrlSlug "how-to-guides" }}
+                How-to Guides
             {{ else }}
                 {{ .p1.Title }}
             {{ end }}


susanev:
@kimberleyamackenzie would it be possible to drop the package-name from the how-to guides breadcrumb?

kimberleyamackenzie:
@susanev done

anita-trimbur:
i don't think this happens in prod, but is anyone else noticing the content shift horizontally when moving between the api docs and any of the other tabs now?

shift

anita-trimbur:
ah that makes sense, cool cool. 

i guess my only other question is will the the breadcrumbs still handle the same once sean changes the h1s? so like, the new h1 for [this page ](https://www.pulumi.com/registry/packages/aws/api-docs/accessanalyzer/analyzer/) would be `aws.accessanalyzer.Analyzer`. will that change affect the breadcrumb structure from the expected `AWS / API Docs / accessanalyzer / Analyzer`?

sean1588:
> i guess my only other question is will the the breadcrumbs still handle the same once sean changes the h1s? so like, the new h1 for [this page ](https://www.pulumi.com/registry/packages/aws/api-docs/accessanalyzer/analyzer/)would be aws.accessanalyzer.Analyzer. will that change affect the breadcrumb structure from the expected AWS / API Docs / accessanalyzer / Analyzer?

i believe the breadcrumb structure will stay the same after I get the h1 changes in, so in this case, it will still say  AWS / API Docs / accessanalyzer / Analyzer.

kimberleyamackenzie:
jinx

anita-trimbur:
thanks for clarifying! everything else lgtm, approving now <3

anita-trimbur (approved):

sean1588 (approved):
lgtm!!

:ship:

Issue #1705, opened 2022-11-18
create compact alerts in new package page card


anita-trimbur:
## overview

this issue is a reworking of https://github.com/pulumi/registry/issues/1356

we would like to ensure that folks who are entering into our api docs from web searches don't miss important alerts only visible from the overview page. these designs introduce a compact message into the [new package page cards](https://github.com/pulumi/registry/issues/1531) in the right column.

## designs

<img width="384" alt="Group 558" src="https://user-images.githubusercontent.com/97129308/202812233-5fb91da7-90c9-49e5-86ca-7687a2a776bf.png">

Issue #1703, opened 2022-11-18
Change H1s to match selected tab


anita-trimbur:
## overview

we would like to change the h1s in our package packages to reflect the selected tab. this is a change intended to improve accessibility and seo. #1702 

## example

currently we have an h1 in the heading of the package page that is only the name of the package. eg. `Datadog` or `Azure Native`. whenever page content changes, the h1 should reflect that. 

now, we would like to move the h1 out of the heading and into the content area. the new h1s are as follows:

* the overview tab will remain as `Packagename` (e.g. `Datadog`, `Azure Native`)
* the installation tab should read `Packagename: Installation & Configuration`
* the guides tab should read `Packagename: How-to-Guides`

the API docs are an exception and are handled in another issue https://github.com/pulumi/registry/issues/1702

## reference for h1 position

<img width="1440" alt="1280+ no alert (2)" src="https://user-images.githubusercontent.com/97129308/202808527-9d320d18-7077-4302-8a50-85baa388a4f8.png">

Issue #1702, opened 2022-11-18
Change H1s in the API docs to match title tag


anita-trimbur:
## overview

we would like to change the primary h1s in the API docs to correspond with the title tag that displays in browser. this is useful for both accessibility and seo. 

## example

[this resource in the datadog api docs ](https://www.pulumi.com/registry/packages/datadog/api-docs/aws/integration/) has two h1s on the page. the first appears in the header reads `Datadog`, referring to the name of the package. the second h1 appears in the content area for the api docs and reads `Integration`, referring to the particular resource. 

instead, we want only a single h1 on this page that appears inside the content area that matches the title tag for this page: `datadog.aws.Integration`.

## designs

_the api docs nav pictured in this frame is **not** included in this body of work_

<img width="1728" alt="1536+ (2)" src="https://user-images.githubusercontent.com/97129308/202806403-ad389aba-c325-4824-92b1-f2edc497f24e.png">

susanev:
@anita-trimbur can you add a lil note on this one in the description, that the nav in the picture is not being implemented as part of this work?

Issue #1531, opened 2022-09-22
Move package info into card


anita-trimbur:
## Overview

To make the headers more compact on our package pages, we'd like to move package info including

* logos
* publication version and dates
* repo links

into a card in a new right page column. 

[Associated epic](https://github.com/pulumi/home/issues/2250)

## Designs

_responsive designs in figma file_

<img width="1440" alt="1280+ no alert (1)" src="https://user-images.githubusercontent.com/97129308/202802244-abf5e906-f7c4-4fdd-be45-2592f3a4105e.png">

## Changes
* two column layout
* logo, publication date/author, and source code link moved into card
* new compact alert added to header (separated here https://github.com/pulumi/registry/issues/1705)
* max-height of logos 24px UNLESS logo width is greater than 164px
* logos wider than 164px should be reduced to a max-width of 164px but **retain a container height of 24px for consistency**
* `source code` link text should communicate the repo we're linking to (e.g. `pulumi/pulumi-azure`)

### 0.79
design work wrapping up pending responsive resizing designs

### 0.80
reviewed and costed eng work with sean 
met with luke and confirmed priority order of work, made design adjustments
made adjustments following design critique

### 0.81
eng work to be picked up by sean and potentially some by kimberley during this milestone

PR #1797, opened 2022-12-07
registry package pages

sean1588:
theme: https://github.com/pulumi/theme/pull/397

issues:
https://github.com/pulumi/registry/issues/1531 - create package card
https://github.com/pulumi/registry/issues/1705 - compact card alerts
https://github.com/pulumi/registry/issues/1702 - H1s for API Docs
https://github.com/pulumi/registry/issues/1703 - H1s match selected tabs

figma designs: 
https://www.figma.com/file/bVEoSwAvU7KDpuPNri332F/Compact-Registry?node-id=117%3A3379&t=aHUxQCTiyRaFs11P-0
https://www.figma.com/file/bVEoSwAvU7KDpuPNri332F/Compact-Registry?node-id=227%3A5397&t=3sCE2WZGUSX9Dm5q-0

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.16
 require (
 	github.com/pulumi/pulumi-hugo/themes/default v0.0.0-20221211165432-e59734a63d61 // indirect
 	github.com/pulumi/registry/themes/default v0.0.0-20211011171710-45eb4e243ab7 // indirect
-	github.com/pulumi/theme v0.0.0-20221209005854-7b2549855128
+	github.com/pulumi/theme v0.0.0-20221212184602-3dc24256e023
 )
 
 // The override is needed because this repo is currently private and module at themes/default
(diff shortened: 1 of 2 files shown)


anita-trimbur:
aahhh so exciting!! so glad to see this come together

a few minor style nits for the card itself: 
* the font and size of every element on the open card should be Inter 12px
* the only font that's different is the "last published" on the collapsed card which is Inter 10px gray600`#5A5A5A`
* (this is my bad, it's wrong in figma) we should use blue600 `#4D5BD9` for that info icon to match the note containers
* the info icon and the github icon should be the same size -- i think without the padding it'll be about 15x15
* i'm pretty sure the border color is a little light on the open card, should be gray300 `#DCDCDC`. on the closed card it's gray400 `#B0B0B0` along with the down/up chevron

sean1588:
awesome. thanks @anita-trimbur. I'll get those changes made!

anita-trimbur:
wow you got those changes in fast! 

last nit: can we shrink that external link icon by the source code repo? maybe down to like 10x10. i think that link text can also be 12px.

anita-trimbur:
following up on our team discussion regarding grays. here's the breakdown

grays

tailwind is using the blue-grays that we were trying to get rid of. the _colors.scss file nearly matches the service grays in figma with the exception of gray 200, so it shouldn't be problematic to align gray200 as `#EFEFEF`. i think the difference in border color on the collapsed card was a color sampling error i made in figma -- sean you're right that it should match gray300 as it is defined in the service/.scss file.

anita-trimbur (approved):
🚀🚀🚀

themes/default/layouts/partials/registry/package/package-alert.html
```diff
@@ -0,0 +1,37 @@
+{{ $package := .Scratch.Get "package" }}
```
kimberleyamackenzie, line 1: TIL about scratch!

kimberleyamackenzie (approved):
Amazing!! 🎉 🚢 :shipit:

Issue #1847, opened 2022-12-16
fix header sizes for h1 and h2 tags


sean1588:
revert size changes to headers, 36px for h1, 32px for h2 and standardize header styles

PR #1848, opened 2022-12-16
standardize headings styles

sean1588:
issue: https://github.com/pulumi/registry/issues/1847

theme pr: https://github.com/pulumi/theme/pull/405

I removed the styling from the h1 tags that we have. The code for the h2 tag styling is in the theme PR. There were also some margins we had on some of these tags. I can add them back in if we think they are needed. But seems to look ok to me.

How-To Pages:

On the root page of the how-to guides, I did move the h3s to h2s so that there is no loner a jump from h1 to h3 on that page.
However concerning tag styling that may be on the individual how-to-guide sub pages, I will address with the issue specifically concerning those, since there is a good chance I will need to go through all of those pages and make changes as well as move them from h3 -> h2.

### DIFF
--- a/themes/default/layouts/registry/how-to.html
+++ b/themes/default/layouts/registry/how-to.html
@@ -13,7 +13,7 @@
                     There are no guides yet for this package.
                 {{ else }}
                     {{ if not (eq (len (where .Pages "Params.language" "==" nil)) 0) }}
-                        <h3>Featured Guides</h3>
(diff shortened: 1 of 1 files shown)


anita-trimbur (approved):

kimberleyamackenzie (approved):
Love theme as the single source of truth here, thank you @sean1588 !

Issue #1823, opened 2022-12-10
update how-to guide headings to use page title


susanev:
seo doesnt like different pages having the same h1
and since we have the same how-to guide in multiple languages on different pages this leads to issues

the titles are customized to include an or bar and the language (see breadcrumb link), lets update the h1s to use the same thing

<img width="727" alt="Screen Shot 2022-12-10 at 12 00 47 PM" src="https://user-images.githubusercontent.com/5489125/206873249-7e67724f-b30e-4ccc-ba67-d98d1986c824.png">

susanev:
@sean1588 after your current pr lands, up for taking this?

sean1588:
@susanev yeah I can take a look at this.

PR #1849, opened 2022-12-16
Update how-to-guide headings

sean1588:
fixes: https://github.com/pulumi/registry/issues/1823

This PR moves to always using the title as the heading for the how to guide pages so that they always match the breadcrumb, since that is the same way the breadcrumb is being populated. Prior to this we were using the h1 property in the front matter if it existed and would fall back to using the title if it wasn't present. I have switched it to always use the title so we always match the breadcrumb.

_I added in the how to guides for aws so we can test it out. I'll remove the commit with these additional files once we approve the way it looks._

image

### DIFF
--- a/themes/default/layouts/registry/how-to-guide.html
+++ b/themes/default/layouts/registry/how-to-guide.html
@@ -1,5 +1,4 @@
 {{ define "main" }}
-{{ $lastUrlSlug := index (last 1 (split (delimit (split $.Page.File "/") "," "") ",")) 0 }}
     <div class="flex flex-row container mx-auto px-4 pb-8 mt-2">
         <div class="min-w-0">
             {{ partial "registry/package/header.html" . }}


susanev (commented):
headings look good, but i think the content on the how-to guides is possibly being constrained too much, at least with the heading

Screen Shot 2022-12-15 at 5 43 25 PM
Screen Shot 2022-12-15 at 5 42 38 PM

anita-trimbur:
im seeing the same content width issue as susan at other breakpoints in the how-to-guides also.
Screen Shot 2022-12-16 at 12 38 15 PM

themes/default/layouts/registry/how-to-guide.html
```diff
@@ -1,11 +1,12 @@
 {{ define "main" }}
+{{ $lastUrlSlug := index (last 1 (split (delimit (split $.Page.File "/") "," "") ",")) 0 }}
```
kimberleyamackenzie, line 2: Is this needed? or maybe a rebase leftover?

kimberleyamackenzie (approved):
1 comment on the code, but LGTM!

themes/default/layouts/registry/how-to-guide.html
sean1588: yes thank you so much. will remove!

anita-trimbur (approved):

Issue #1881, opened 2023-01-04
registry left nav is not showing ellipses at larger screen widths.


sean1588:
![image](https://user-images.githubusercontent.com/16751381/210624350-f6a291f1-7ecc-4370-ac60-bb8a2f9cc12a.png)

PR #1882, opened 2023-01-04
fix left nav not showing ellipses

sean1588:
fixes: https://github.com/pulumi/registry/issues/1881
theme: https://github.com/pulumi/theme/pull/422

this PR address the issue with registry left nav cutting off the ellipses with long package names.

greater than 1536px
image

 between 1280px  and  1536px
image

less than 1280px
image

### DIFF
--- a/themes/default/layouts/registry/api.html
+++ b/themes/default/layouts/registry/api.html
@@ -4,7 +4,7 @@
     <div class="flex-col min-w-0">
         {{ partial "registry/package/header.html" . }}
         <div class="lg:flex">
-            <div class="lg:w-9/12 xxl:w-8/12">
+            <div class="xxl:w-8/12">
                 <!--Don't show the title in the top-most index page.-->
                 {{ if not (in .Page.File "api-docs/_index.md") }}
                     {{ if and (.Params.h1) (not .Params.notitle) }}


susanev:
@sean1588 is it possible to not shrink the left-nav width at all? i think previously it did not shrink its width, and instead, the page content became narrower. @kimberleyamackenzie should correct me if im wrong.

sean1588:
I rewinded my branch to before all the registry changes to see how things looked before and tried to match as close as possible to before the new registry changes were in. We did have a limit on how wide the left nav got and did show ellipses on really long package names. Here is how it looked prior to all the registry changes:
image

susanev:
@sean1588 the left nav in your screenshot there looks wider than your screenshot above at 1280, that's the thing im trying to highlight.

sean1588:
The screenshot i showed was above 1536 breakpoint. The left nav did shrink a bit between 1280 and 1536 prior to the registry changes. I can widen the left nav between 1280 and 1536, but it looks like there is a lot of padding on the right side since the names that are rendered get shorter. After doing some digging around this seems to be due to us truncating the characters in the stencil component that is in the theme repo that is used to render this. I can go ahead and make some adjustments there to get remove what is being done between 1280 and 1536 altogether and just have a wider nav from 1280 on up if that sounds good to you all.

kimberleyamackenzie:
The code in this PR + theme LGTM - I honestly don't remember the previous behavior in regards to responsive sizing of the available width for the nav itself - my only question is @sean1588 can you verify locally that this behaves as expected for API docs with modules that are children of other modules? I think Kubernetes is the provider where this happens most commonly, like in the screenshot below.
Screenshot 2023-01-04 at 1 34 41 PM

themes/default/layouts/registry/api.html
```diff
@@ -19,7 +19,7 @@ <h1>{{ .Params.title }}: API Docs</h1>
                     {{ .Content }}
                 </section>
             </div>
-            <div class="w-3/12 pr-8 mb-2 pt-8 lg:order-first lg:pt-0 lg:mt-0">
+            <div class="flex-shrink-0 pr-8 mb-2 pt-8 lg:order-first lg:pt-0 lg:mt-0">
```
sean1588, line 22: I moved this to using `flex-shrink` since we moved to using flex to layout the page with the new registry changes that were shipped. When we were using the 12 column grid to determine the width of this column, the width was actually being determined in two places - both here (e.g. `w-3/12`) and in the stencil component based on the character limitations set at different breakpoints. This ended up resulting in some awkward extra white space at certain screen widths when I updated the stencil component to remove the breakpoint between 1280 and 1536. Now the column will just shrink to the width needed based on the character count set in the stencil component and the main content area will naturally expand to fill the remaining space on the screen.

themes/default/layouts/registry/api.html
```diff
@@ -4,7 +4,7 @@
     <div class="flex-col min-w-0">
         {{ partial "registry/package/header.html" . }}
         <div class="lg:flex">
-            <div class="lg:w-9/12 xxl:w-12/12">
+            <div class="xxl:w-8/12">
```
sean1588, line 7: updated this to just be `w-8/12` above xxl breakpoint because this is where the right nav gets shown, so we need to account  for that space on the screen when it gets rendered above that point. Below, that point the main content area should just expand naturally.

sean1588:
thanks @kimberleyamackenzie - good point! Here is a screenshot from kubernetes locally. Also, I think I may have pushed a couple changes after you had reviewed or maybe it was just the preview link that got posted afterwards. You may want to check the latest preview, just to confirm.

image

susanev (approved):
thx Sean!

kimberleyamackenzie (approved):

PR #1884, opened 2023-01-04
copy over hand-authored how-to-guides

susanev:
fixes https://github.com/pulumi/registry/issues/834

this will allow the edit this page on hand-authored guides not to 404, after this lands, ill remove them over in hugo

edit this page on the hand-authored guides will still fail on the preview because its not on master yet, but once this merges they will work


susanev:
@kimberleyamackenzie yah the generated ones from the examples repo still come in via the docs repo. so i think that's still good. but i think this readme needs some work. ill try and fix that up later.

PR #1886, opened 2023-01-05
fix up confluent

susanev:
this is a partial fix
the rest of the fix is in https://github.com/pulumi/pulumi/pull/11775

this fixes the package page issues (except for the package name in the card)
the browse page and the package name in the card is fixed as part of the pulumi/pulumi pr

i still think we can land this ahead of 11775 because its still a step forward

### DIFF
--- a/themes/default/layouts/partials/registry/package/package-alert.html
+++ b/themes/default/layouts/partials/registry/package/package-alert.html
@@ -38,7 +38,7 @@
 <div class="text-xs mt-4">
     <i class="fas fa-info-circle text-blue-600 mr-1 my-1"></i>
     <span>
-        This package is deprecated. We recommend using the <a class="text-blue-600 underline" href="/registry/packages/confluentcloud/">new Confluent Cloud</a> provider.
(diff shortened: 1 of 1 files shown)


themes/default/layouts/partials/registry/package/package-alert.html
```diff
@@ -34,4 +34,11 @@
         Google Cloud Native is in public preview. <a class="text-blue-600 underline" href="/registry/packages/gcp/">Google Cloud Classic</a> remains fully supported.
     </span>
 </div>
+{{ else if (eq $package "confluent") }}
+<div class="text-xs mt-4">
+    <i class="fas fa-info-circle text-blue-600 mr-1 my-1"></i>
+    <span>
+        This package is deprecated. We recommend using the <a class="text-blue-600 underline" href="/registry/packages/confluentcloud/">new Confluent Cloud</a> provider.
```
susanev, line 41:
```suggestion
        This package is deprecated. We recommend using the <a class="text-blue-600 underline" href="/registry/packages/confluentcloud/">new Confluent Cloud</a> package.
```

mnlumi (approved):
Thank you!

PR #1888, opened 2023-01-05
Fix azuread docs: subscription id is not a config setting

thomas11:
See pulumi/pulumi-azuread#323 - since Active Directory is not subscription-based, setting a subscription results in
```
could not validate provider configuration: 1 error occurred:
* Invalid or unknown key
```


guineveresaenger (approved):
Thank you for catching this!

thomas11:
> Thank you for catching this!

Sadly, a user caught it, but thanks :)

PR #1889, opened 2023-01-05
docs: add yaml to azure-native version guide

dirien:
Hi,

This PR adds an YAML example to the azure native version guide.

### DIFF
--- a/themes/default/content/registry/packages/azure-native/version-guide.md
+++ b/themes/default/content/registry/packages/azure-native/version-guide.md
@@ -75,7 +75,7 @@ keyvault.NewVault(/*...*/)
 
 {{% choosable language java %}}
 
-Currently no support for versions in Java.
+Java is not supported.
 
 {{% /choosable %}}
 


themes/default/content/registry/packages/azure-native/version-guide.md
```diff
@@ -73,6 +73,25 @@ keyvault.NewVault(/*...*/)
 
 {{% /choosable %}}
 
+{{% choosable language java %}}
+
+Currently no support for versions in Java.
+
+{{% /choosable %}}
```
susanev, line 80: should we just not include it?

dirien: We have customers explicitly asking for it! I wanted to avoid them asking in Slack or Support for after reading this page!

susanev: hmm? i read that thread differently, in that they were asking for java support and assumed the lack of java meant it wasn't supported, which was correct. we can merge this, but we should probably adjust the standard because on all other pages, when a language isn't supported we don't show the tab.

themes/default/content/registry/packages/azure-native/version-guide.md
```diff
@@ -73,6 +73,25 @@ keyvault.NewVault(/*...*/)
 
 {{% /choosable %}}
 
+{{% choosable language java %}}
+
+Currently no support for versions in Java.
```
susanev, line 78:
```suggestion
Java is not supported.
```

we try and avoid the use of the word _currently_ because it implies it will happen in the future, which isn't always true

susanev (approved):

themes/default/content/registry/packages/azure-native/version-guide.md
dirien: True, good feedback!

PR #1950, opened 2023-01-24
Add ZScaler ZPA provider

ringods:
Working with @willguibr to get the ZScaler providers listed in the registry.

Signed-off-by: Ringo De Smet <ringo@de-smet.name>


dirien (approved):
LGTM!

PR #2002, opened 2023-01-30
Update GKE kubeconfig example in how-to guide

lblackstone:
The existing example won't work for newer Kubernetes clusters, so update with the required changes.

Update copied from https://github.com/pulumi/examples/blob/129bb4f3c13ef94b34fc28091f31baa3827cc9bc/gcp-ts-gke/cluster.ts


rquitales (approved):
LGTM

PR #2023, opened 2023-02-02
add slack alerts

sean1588:
part of issue: https://github.com/pulumi/home/issues/2566

This PR adds slack notifications on failures for workflows in this repo. These are enabled for the workflows that run on push to master, the provider docs build, and the scheduled job that updates the package metadata. They are not turned on any jobs that run before a pull request merges. These notifications will get posted to the docs-ops channel in slack, which contains a link to the failed job and the name of the workflow that failed (listed as the actions url in the image below). 

The image below says "Pull Request" as the name of the workflow that failed, because I was using that workflow just for testing (they are not actually turned on for pull request failures though).

image


sean1588:
this PR was already approved here https://github.com/pulumi/registry/pull/1978
so going to just merge it. YOLO!

PR #2083, opened 2023-02-16
Update package-list.json for Upstash

burak-upstash:
One question: I have updated relevant urls on the provider repo, only thing left to do is to change the repo name. Haven't done that since I am not sure how this will affect previous releases. Can you give a quick feedback on the [upstash pulumi provider repo](https://github.com/upstash/upstash-pulumi-provider) please?

### DIFF
--- a/community-packages/package-list.json
+++ b/community-packages/package-list.json
@@ -77,7 +77,7 @@
       "schemaFile": "provider/cmd/pulumi-resource-fusionauth/schema.json"
     },
     {
-      "repoSlug": "upstash/upstash-pulumi-provider",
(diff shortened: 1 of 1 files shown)


ringods:
@burak-upstash Github documented that, after renaming a repository, they redirect all web traffic.

https://docs.github.com/en/repositories/creating-and-managing-repositories/renaming-a-repository

I would assume that any retrieval for the plugin binary on the old URL will be redirected to the new URL.

Worst case: looking at the current version number, you are still at a v0 prefix, meaning this isn't a stable provider yet and you still can have breaking changes.

burak-upstash:
Awesome, thanks for the comment

PR #2120, opened 2023-02-27
Using the repoSlug as the job name for package update checks

ringods:
A job can have [a name](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idname). 

I configured the name to be the `repoSlug` from the list of packages to make the name shorter in the Github UI. The package being checked becomes readable this way.


sean1588:
oh nice! thanks @ringods! that was fast!

sean1588 (approved):

PR #2132, opened 2023-03-02
Add AzAPI provider to registry

dirien:
Hi,

this PR adds the `pulumi-azapi` provider to the registry


dirien:
@ringods Any idea, why I don't see the provider in the registry?

susanev:
@ringods moving forward, can you hold off on merging these until Sean or i have had a quick look? this caused a build failure because of a formatting issue.

i promise we want your help, but we are just getting started and need to setup a few more things before we feel good about contributors merging these prs.

thx for considering!

PR #2142, opened 2023-03-03
fix json error in package-list

sean1588:
There is an error in the package-list.json file caused by a trailing comma. This is causing the check package updates job to fail since it depends on this json file. 

https://github.com/pulumi/registry/actions/runs/4325845570/attempts/2


dirien (approved):
Oh no 🙈! Good catch @sean1588 !

sean1588:
no worries! thanks @dirien!

dirien:
@sean1588, I still wonder why this provider do not show up in the list though! 🤔

PR #2147, opened 2023-03-04
fixed python example documentation

NCMartins:
while copy pasting the python example on the proxmox example from the website we get several errors. This PR aims to fix those errors.

- usage of : instead of = 
- comment // instead of #
- import fixing


ringods:
Hello @NCMartins, please contribute your changes to the upstream repository:

https://github.com/muhlba91/pulumi-proxmoxve

During the publishing process, we take over docs & examples from that repository. If we accept your changes here, these would be overwritten when the next release of this provider gets published.

NCMartins:
@ringods Hi!

Ok, got it! Thanks for the feedback!

PR #2229, opened 2023-03-20
fusionauth: Close choosable

abhinav:
The choosable for fusionauth wasn't closed.
This causes pulumi-hugo to fail with the following error
(cleaned up for readability):

    Error:
    Error building site: "[..]/registry/packages/fusionauth/_index.md:16:1":
    failed to extract shortcode:
    shortcode "chooser" must be closed or self-closed

This fixes it by adding the missing `/choosable`.

Example of valid case on another page:
https://github.com/pulumi/registry/blob/80334639da2647f62920cfe67c9ff2b3dc87ebdc/themes/default/content/registry/packages/akamai/_index.md?plain=1#L12-L31

Refs pulumi/pulumi-hugo#2572

PR #2231, opened 2023-03-21
twingate: Fix unclosed chooser

abhinav:
Same as #2229, twingate didn't have its chooser closed.
This was probably working on older versions of Hugo silently,
but recent releases have turned this into an error.

Refs pulumi/pulumi-hugo#2572


dixler (approved):
nice catch.

PR #2275, opened 2023-03-31
Fix Java tab for Docker overview

guineveresaenger:
The Overview tab had some historic errors. This PR removes the redundant Javascript tab and adds a Java section.

Fixes pulumi/pulumi-docker#383.

- fix up chooser blocks for Java
- Java example


susanev (approved):
before and after pics would be rad for next time

PR #2306, opened 2023-04-04
Remove unlinked kubernetes introduction page

mnlumi:
This page does not have any references pointing here and contains a broken link as well.


susanev (approved):
yolo.

PR #2331, opened 2023-04-11
ssh->http

ericrudder:
typo fix


susanev:
@ericrudder i was about to open this pr, but on closer read of the issue noticed its going to also require port changes tot he c# example.

ericrudder:
hmm ... so much for the "wrong->right" tutorial when "wrong" is "right"

also changed a port90 to an 80 while i was there

jkodroff (approved):

PR #2338, opened 2023-04-12
Add example of how to include subfolders for s3 website

jtrinklein:
I've updated the S3 static website how-to documentation to include an example of how to handle subfolders.

There were a couple gotchas that gave me quite a bit of headache when I worked through this myself and I thought it would be helpful to provide my solution to anyone looking to do the same thing.

### DIFF
--- a/themes/default/content/registry/packages/aws/how-to-guides/s3-website.md
+++ b/themes/default/content/registry/packages/aws/how-to-guides/s3-website.md
@@ -37,33 +37,6 @@ const path = require("path");
 // Create an S3 bucket
 let siteBucket = new aws.s3.Bucket("s3-website-bucket");
 
-const addFolderContents = (siteDir) => {
-  // For each file in the directory, create an S3 object stored in `siteBucket`
-  for (let item of fs.readdirSync(siteDir)) {
-    let filePath = path.join(siteDir, item);
-    let object = new aws.s3.BucketObject(item, {
-      bucket: siteBucket,
-      source: new pulumi.asset.FileAsset(filePath),     // use FileAsset to point to a file
-      contentType: mime.getType(filePath) || undefined, // set the MIME type of the file
-    });
-  }
-}
-
-addFolderContents("www"); // base directory for content files
-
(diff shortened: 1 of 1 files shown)


susanev:
thx for your contribution! we will take a look as soon as we can <3

themes/default/content/registry/packages/aws/how-to-guides/s3-website.md
```diff
@@ -31,29 +31,66 @@ Open up `index.js` using your preferred text editor and add the following code.
...
+addFolderContents("www"); // base directory for content files
+
 exports.bucketName = siteBucket.bucket; // create a stack export for bucket name
 ```
 
 Notice how we are using the [@pulumi/aws] npm package to create an Amazon S3 bucket with [aws.s3.Bucket]. This uses the s3 module and the Bucket resource for the package. For each file in `www`, an [aws.s3.BucketObject] gets created, using the helper [pulumi.asset.FileAsset] to reference a local file.
 
 In order to serve the files to a browser, the content type for each S3 object must be set. For this, the [NPM `mime` package](https://www.npmjs.com/package/mime) is used. Just like a regular Node program, Pulumi programs can use any Node.js package.
 
+#### Optional: Include subfolders in your site
```
sean1588, line 61: thanks @jtrinklein for your contribution! 

Is there a specific reason why we need a separate section to differentiate being able to add sub directories? If I was to call `addFolderContents`, my expectation would be that it adds the contents of the folder specified (including the sub directories), which is what this version of it does. Can we consolidate these by replacing the `addFolderContents` function you added in the first example with this version of it and then get rid of this one? Let me know what you think. 

Thanks again for contributing!

jtrinklein: I didn't have any specific reason not to just replace the original code block other than using this format seemed to fit the flow of the document. I personally agree that `addFolderContents` should add everything, including subfolders, by default. I figured I would present it first as broken out steps and get some feedback on it. I'm more than happy to just update the original code block.

sean1588 (approved):
thanks @jtrinklein for making those changes. looks good to me!

PR #2346, opened 2023-04-13
add codeowners file

susanev:
we have had a large number of prs that have merged from this repo
that have broken the build for the entire website
so putting some protection in place
until we improve the rest of docs processes


susanev:
this isnt really protecting the core of the problem, but this will allow us to at least know when folks are adding new packages

sean1588 (approved):
good idea. thank you susan!

PR #2363, opened 2023-04-18
update create bucket command with permissions

sean1588:


scripts/ci/sync.sh
```diff
@@ -40,6 +40,10 @@ destination_bucket_uri="s3://${destination_bucket}"
 # created in another account, in which case subsequent operations on the bucket will also
 # fail, causing this script to exit nonzero. In either case, it's okay to continue.
 aws s3 mb $destination_bucket_uri --region "$(aws_region)" || true
+aws s3api put-public-access-block --bucket $destination_bucket --public-access-block-configuration BlockPublicAcls=false
```
sean1588, line 43: https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-control-block-public-access.html

scripts/ci/sync.sh
```diff
@@ -40,6 +40,10 @@ destination_bucket_uri="s3://${destination_bucket}"
...
+aws s3api put-bucket-ownership-controls --bucket $destination_bucket --ownership-controls="Rules=[{ObjectOwnership=ObjectWriter}]"
```
sean1588, line 44: this had to be added because if not the line below will fail with
```Bucket cannot have ACLs set with ObjectOwnership's BucketOwnerEnforced setting```

jazzyfresh (approved):

lukehoban (approved):

cnunciato (approved):
We should circle back and add some comments as to what these lines do, but not need to block on this now. 👍

susanev (approved):
thank you Sean!! 💪🏻

PR #2427, opened 2023-05-02
Added Port as a community package to registry

danielsinai:


dirien (approved):
LGTM!

susanev:
hi folks! this needs docs, before we can merge this into registry.
https://www.pulumi.com/docs/guides/pulumi-packages/how-to-author/

susanev:
good call @ringods i hadnt looked at that yet, that will need to change as well.

danielsinai:
I will handle it and update the PR soon, thanks 😄 
BTW, why is the requirement of the repo name to end with the `port` suffix?

ringods:
@danielsinai once your repo is registered, we have code in a few places, one of which is in Github Actions, which does pattern matching on the name of the repo to check for new releases of your provider. We are not in a situation at the moment that we can drop that requirement.

dirien:
@danielsinai: New PR for the new registry of the provider with correct repo name `pulumi-port` 


https://github.com/pulumi/registry/pull/2814

Issue #2373, opened 2023-04-20
The website shows sign in on the registry home page instead of the Pulumi cloud link


nickpulumi:
### What happened?

I am logged in to the app.pulumi.com and this shows correctly on the docs site but not on the registry site. When I click to login on the registry page it logs me in correctly.  

Related to https://github.com/pulumi/pulumi-service/pull/13196 

## Current behavior

<img width="1185" alt="Screenshot 2023-04-17 at 2 33 03 PM" src="https://user-images.githubusercontent.com/54290146/232615601-b5e3b495-7535-49d3-9d7a-25590c6979ee.png">

### Expected Behavior

<img width="1185" alt="Screenshot 2023-04-17 at 2 32 43 PM" src="https://user-images.githubusercontent.com/54290146/232615539-2d4cb588-28b8-4062-983e-9ded300d364d.png">

### Steps to reproduce

Log in to app.pulumi.com
go to [pulumi.com/registry](https://www.pulumi.com/registry/) 

### Output of `pulumi about`

n/a

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

nickpulumi:
@cnunciato

PR #2428, opened 2023-05-02
update registy prac nav w/ new sign in/pulumi cloud functionality

susanev:
i dont know how to test this
but i think this will work 🤷🏻 

this is using a partial that was added in hugo
https://github.com/pulumi/pulumi-hugo/pull/2703

fixes https://github.com/pulumi/registry/issues/2373


cnunciato (approved):
This works. Thanks. 🚀

PR #2435, opened 2023-05-03
link to ai from right panel

susanev:
for folks new to this repo, api docs only work for aws classic, and aiven in the preview link

Screenshot 2023-05-02 at 7 49 17 PM
Screenshot 2023-05-02 at 7 49 01 PM
Screenshot 2023-05-02 at 7 48 57 PM

### DIFF
--- a/themes/default/layouts/partials/registry/package/right-nav.html
+++ b/themes/default/layouts/partials/registry/package/right-nav.html
@@ -27,7 +27,7 @@
             </h4>
             <section class="text-sm">
                 <p class="m-2">Need help using this package?</p>
-                <p class="m-2">Try out <i class="fas fa-robot mr-1 text-xs relative text-gray-800" style="top: -1px;"></i><a target="_blank" href="/ai/">Pulumi AI</a>!</p>
+                <p class="m-2">Try out <i class="fas fa-robot mr-1 text-xs relative text-gray-800" style="top: -1px;"></i><a target="_blank" href="/ai/" class="underline">Pulumi AI</a>!</p>
             </section>
         </div>
     {{ end }}


susanev:
okay so the api docs preview looks super weird; @sean1588 is this cause related resources don't work in the previews? anyway to get an idea of what this will look like b before shipping?

themes/default/layouts/partials/registry/package/right-nav.html
```diff
@@ -21,8 +21,18 @@ <h4 class="no-anchor">
                 </div>
             {{ end }}
         </div>
+        <div class="pl-3 flex-1 overflow-y-auto border-b-2 pb-3 pt-2">
+            <h4 class="no-anchor mt-2 font-semibold font-body text-sm" style="color: #484848;">
+                <a href="#" class="contents-heading">Pulumi AI</a>
+            </h4>
+            <section class="text-sm">
+                <p class="m-2">Need help using this package?</p>
+                <p class="m-2">Try out <i class="fas fa-robot mr-1 text-xs relative text-gray-800" style="top: -1px;"></i><a target="_blank" href="/ai/" class="underline">Pulumi AI</a>!</p>
```
lukehoban, line 30: I would love to contextualize these links even more.

When on https://www.pulumi.com/registry/packages/aws/api-docs/ec2/instance/, I'd love to send the user to `/ai` with a query string setting the prompt to "Create an `aws.ece.Instance` resource" or similar.

I'd also love to not say "using this package", but to contextualize this to the individual resource.

For pages that aren't resource pages, we can back off to "use this package" or "use this module" with a similar approach.

Relatedly, I might phrase as "Explore `aws.ec2.Instance` with Pulumi AI" instead of "Try out Pulumi AI".

themes/default/layouts/partials/registry/package/right-nav.html
```diff
@@ -21,8 +21,18 @@ <h4 class="no-anchor">
                 </div>
             {{ end }}
         </div>
+        <div class="pl-3 flex-1 overflow-y-auto border-b-2 pb-3 pt-2">
```
lukehoban, line 24: Slightly worried that putting this after the contents section will push it off the screen on many displays.  Given it's a fixed height, might it make sense to place above Contents?

themes/default/layouts/partials/registry/package/right-nav.html
lukehoban: My very poor attempt at this from Friday :-). 

```html
<img width=20 src="/registry/packages/einsteinypus.svg" style="display: inline-block"/> Explore <a href="/ai?prompt=Create%20an%20%60{{ $.Page.Params.title_tag }}%60%20resource">{{ $.Page.Params.title_tag }}</a>
```

themes/default/layouts/partials/registry/package/right-nav.html
susanev: im pretty worried about putting this above contents; we are trying to keep that in a consistent place across all doc types. does it being added to the practitioner nav help at all? also have prs opened for those.

let me chat with @anita-trimbur she might have some better ideas about placement.

susanev:
closing in favor of https://github.com/pulumi/registry/pull/2437

PR #2457, opened 2023-05-08
add docs registry tab navigation

sean1588:
depends on: https://github.com/pulumi/pulumi-hugo/pull/2780

These changes are to support the PR open in hugo which has all the styling changes

### DIFF
--- a/themes/default/layouts/partials/registry/header.html
+++ b/themes/default/layouts/partials/registry/header.html
@@ -73,30 +73,6 @@
     </div>
     {{ .Scratch.Set "mode" "registry" }}
     {{ partial "docs/search.html" . }}
-    <!-- <div class="header-pinned-container hidden md:flex xl:w-3/12 px-4 lg:px-8 xl:px-5">
-        <div class="header-hero-title pr-3">
-            <div class="flex items-center">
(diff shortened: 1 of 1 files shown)


themes/default/layouts/partials/registry/header.html
```diff
@@ -70,37 +70,9 @@
...
-        <div class="header-pinned items-center transition-all hidden">
-            <img class="h-6 mr-12" src="/images/logo/logo-on-white.svg" alt="Pulumi logo" />
-            <div class="get-started-pinned-search-container">
-                <a class="get-started-pinned-search-button" data-track="get-started-pinned-search" 
-                href="{{ relref . "/docs/get-started" }}">Get Started</a>
-            </div>
-        </div>
     </div>
+    {{ .Scratch.Set "mode" "registry" }}
+    {{ partial "docs/search.html" . }}
```
sean1588, line 75: this is now using a shared partial between docs and registry, which is defined in the hugo PR. Note, the specific scripts to work with swift-type setting between the two have remained here, so the search results logic themselves will still be the same as before. It is just the search bar component styling that is now shared.

susanev (approved):

PR #2493, opened 2023-05-15
chore: bump rediscloud version

rawkode:


susanev:
@rawkode youll need to make updates and do a new release from the github repo, and then we will automatically pick up changes as part of our automatic process. making a version change here, without that, wont bring in the new changes. i hope this makes sense, let me know if ya have questions or if i can help in anyway.

Issue #2121, opened 2023-02-28
`set-output` in Github Action workflows deprecated by 31 May 2023


ringods:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

`set-output` in Github Action workflows will be fully deprecated by 31 May 2023.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### Affected area/feature

The workflow file `.github/workflows/generate-package-metadata.yml` is using `set-output`.

susanev:
@sean1588 is this fixed now? can we close this?

sean1588:
i think this should be good now. If anyone happens to come across more of these set-output warnings  please let me know.

https://github.com/pulumi/registry/pull/2577
https://github.com/pulumi/registry/pull/2539
https://github.com/pulumi/registry/pull/2538
https://github.com/pulumi/registry/pull/2509
https://github.com/pulumi/pulumi-hugo/pull/2936
https://github.com/pulumi/pulumi-hugo/pull/2903

PR #2509, opened 2023-05-17
remove deprecated set-output commands in workflow files

sean1588:
part of: https://github.com/pulumi/registry/issues/2121

`set-output` is being deprecated and will no longer be able to use it by may 31st. This PR removes this from the workflow file. However we are still getting the errors because there is a [custom actionbeing referenced here that set-output is embedded in](https://github.com/jkisk/list-open-pulls/blob/main/index.js#L19) it that is using this still. So I will need to probably fork it and get that addressed elsewhere to get rid of the last deprecation warning.

Here is the run of the workflow i manually kicked off with these changes
https://github.com/pulumi/registry/actions/runs/5006003820

### DIFF
--- a/.github/workflows/generate-package-metadata.yml
+++ b/.github/workflows/generate-package-metadata.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - id: set-matrix
-        run: echo "matrix=$(python generate_package_list.py)'" >> $GITHUB_OUTPUT
+        run: echo "matrix=$(python generate_package_list.py)" >> $GITHUB_OUTPUT
         working-directory: community-packages
       - name: List open pull requests
         uses: jkisk/list-open-pulls@v1.0.1


.github/workflows/generate-package-metadata.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - id: set-matrix
-        run: echo "::set-output name=matrix::$(python generate_package_list.py)'"
+        run: echo "matrix=$(python generate_package_list.py)'" >> $GITHUB_OUTPUT
         working-directory: community-packages
       - name: List open pull requests
         uses: jkisk/list-open-pulls@v1.0.1
```
sean1588, line 18: This is a custom action that is still using set-output. We will continue to get the deprecation errors until this is updated. This code is managed in jamie's personal repo. I think I will likely need to just do a fork of this to fix it or find another way to list out the pull requests maybe using a shell script or something.

https://github.com/jkisk/list-open-pulls/blob/main/index.js#L19

.github/workflows/generate-package-metadata.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - id: set-matrix
-        run: echo "::set-output name=matrix::$(python generate_package_list.py)'"
+        run: echo "matrix=$(python generate_package_list.py)'" >> $GITHUB_OUTPUT
```
sean1588, line 15: there is a trailing `'` here. This was present before. Was this a typo or something that always existed or is it some type of delimiter or something?

susanev (approved):

.github/workflows/generate-package-metadata.yml
danielrbradley: I'd reckon the `'` is a typo and should be removed. This is fed into a JSON parse so it'll be quite obvious if it fails.

danielrbradley (approved):
This looks good to me - have done a few of these conversions myself and this looks correct 👍 

Would also recommend giving it a quick check using [actionlint](https://github.com/rhysd/actionlint) locally (install [`shellcheck`](https://github.com/koalaman/shellcheck) too for best results. You can also configure this to report for all PRs using this action: https://github.com/marketplace/actions/lint-action

.github/workflows/generate-package-metadata.yml
sean1588: thank you @danielrbradley!

PR #2577, opened 2023-05-26
more set output deprecation removals

sean1588:
OK, I think this should be the last of them for registry. 

This does a few version bumps as well as removes an action we were using before, and just doing the equivalent using bash.

When it comes to doing trivial things that can be done with a few lines of code I would highly recommend that we do that rather than relying on some external action that is maintained elsewhere. This makes it difficult to get fixes out for these things when the action is being maintained by an external party and they are not keeping up with the latest github compatibillity issues. Unless there is some advantage we are gaining by using these actions that I am not aware of.

### DIFF
--- a/.github/workflows/generate-package-metadata.yml
+++ b/.github/workflows/generate-package-metadata.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Get provider short name
         id: regex-prov
         run: |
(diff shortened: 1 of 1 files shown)


susanev (approved):
thx Sean!! good with developing some standards for when we use external actions

PR #2620, opened 2023-06-04
feat: add muhlba91/pulumi-buildkite

muhlba91:
- adds `muhlba91/pulumi-buildkite`provider for http://buildkite.com
- uses terraform bridge with https://github.com/buildkite/terraform-provider-buildkite

### DIFF
--- a/community-packages/package-list.json
+++ b/community-packages/package-list.json
@@ -73,6 +73,10 @@
       "schemaFile": "provider/cmd/pulumi-resource-proxmoxve/schema.json"
     },
     {
+      "repoSlug": "muhlba91/pulumi-buildkite",
+      "schemaFile": "provider/cmd/pulumi-resource-buildkite/schema.json"
+    },
+    {
       "repoSlug": "theogravity/pulumi-fusionauth",
       "schemaFile": "provider/cmd/pulumi-resource-fusionauth/schema.json"
     },


susanev:
hi @muhlba91, thank you for your contribution! we already have a [buildkite provider](https://www.pulumi.com/registry/packages/buildkite/), would you be interested in contributing additions to that instead of creating a new one? if not, can you give me a few details about why that provider is not meeting your needs and why you cannot contribute? thank you!! have a wonderful day! 💜

muhlba91:
hi,

thank you for your note.
yes, I am aware of this provider and was using it until a few months ago when it broke.

i see the following issues with the existing provider:
- the provider scope got renamed from `@grapl` to `@pulumiverse` with no new release effectively breaking the provider as the download url cannot be found anymore (the whole org is renamed).
  -  **the provider is unusable and broken in its current stage!**
  - i opened an issue requesting a new release ~1 month ago (https://github.com/pulumiverse/pulumi-buildkite/issues/23) with no reaction as of now.
  - the same is with your PR (https://github.com/pulumiverse/pulumi-buildkite/pull/21) which is still open since dec 2022.
- last release and commit: dec 2022
- the underlying terraform provider's release is pretty old (https://github.com/pulumiverse/pulumi-buildkite/blob/main/provider/shim/go.mod#L6).
  - the used version `0.11.0` is effectively released in may 2022.
  - the current version (`0.18.0`) is released 4 days ago.
  - the new release supports Organization settings which I would like to use.

due to the non-reactiveness of the developers, no updates happening, and the provider being effectively broken, i don't see an alternative to be able to use pulumi with buildkite other than creating and maintaining a new provider.
i'm open to suggestions but in the current situation there's no way anyone can rely (or even use!) on the existing provider.

(fyi: i have experience in maintaining and updating a provider: https://github.com/muhlba91/pulumi-proxmoxve)

susanev:
@muhlba91 thank you so much for the detailed reply. i need to chat with some folks internally but will get back to you in a few days.

ringods:
Hello @muhlba91, sorry for not noticing the issue in the pulumiverse repo. I'm one of the Pulumiverse board members, which doesn't automatically mean we also maintain all the providers there. With your interest in the buildkite provider, do you want to take over the maintenance of the provider in the pulumiverse organization?

muhlba91:
hi,

thank you for your note, and explanation! :)

yes, sure. i‘d be happy taking over the maintenance in the pulumiverse org. let me know how to proceed, and how i can get the required accesses.

ringods:
@muhlba91 I'll loop you in on a Pulumiverse Github issue. We can take the communication there.

muhlba91:
thank you @ringods.

@susanev i’m closing this issue in fact. thanks a lot!

PR #2706, opened 2023-06-16
fix layouts

susanev:
im not really sure how these were missed, i think its cause the package was having troubles around when this changed; will also submit a pr in their repo


susanev:
lol i already opened one, just looks like it didn't make it into their release
https://github.com/impart-security/pulumi-impart/pull/2

sean1588:
ohh thanks so much for fixing this @susanev!! :heart:

PR #2714, opened 2023-06-16
Add validation script for packages

sean1588:
Add a script to validate docs content dir exists for each of the registry packages.

### DIFF
--- a/scripts/ci/validate-packages.sh
+++ b/scripts/ci/validate-packages.sh
@@ -14,6 +14,15 @@ ls -l "themes/default/data/registry/packages" | tail -n +2 | awk '{print $9}' | 
         echo "ERROR: no content files found for package, $pkg"
         exit 1
     fi
+
+    # check to see that package _index.md file exists and is using the correct layout.
+    layout=$(cat "themes/default/content/registry/packages/$pkg/_index.md" | yq -f eval '.layout')
+    if [[ "$layout" != "package" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/ci/validate-packages.sh
```diff
@@ -0,0 +1,19 @@
...
+# capture all existing packages in the content dir.
+pkg_dirs=$(ls -l "themes/default/content/registry/packages/" | tail -n +2 | awk '{print $9}')
+
+# validates that all packages also have related content directory present.
+ls -l "themes/default/data/registry/packages" | tail -n +2 | awk '{print $9}' |  while read line; do
+    # extract the name from the frontmatter
+    pkg=$(cat "themes/default/data/registry/packages/${line}" | yq ".name")
+    # check to see that a dir exists for the package
+    if [[ "$pkg_dirs" != *"$pkg"* ]]; then
+        echo "ERROR: no content files found for package, $pkg"
```
cnunciato, line 14: Nice. One question though -- since we're here, and checking that the directory exists, what do you think about going a step further and also checking that both content files exist also -- and ideally that they contain the right data (e.g., the correct `layout`)?

sean1588: yeah I can add some extra checks for those too

sean1588:
@cnunciato I update the script to validate the layout and validate that at least an index.md file exists. let me know what you think.

cnunciato (approved):
LGTM. Thanks!

PR #2745, opened 2023-06-23
add lbrlabs-eks package

jaxxstorm:


community-packages/package-list.json
```diff
@@ -227,6 +227,10 @@
     {
       "repoSlug": "pulumiverse/pulumi-matchbox",
       "schemaFile": "provider/cmd/pulumi-resource-matchbox/schema.json"
+    },
+    {
+      "repoSlug": "lbrlabs/pulumi-lbrlabs/eks",
```
EvanBoyle, line 232: I believe this should be `lbrlabs/pulumi-lbrlabs-eks`

cnunciato: @jaxxstorm You'll also need to change the `layout` param in your `_index.md` doc to `package`. (Apparently this was changed recently so just calling it out so you aren't surprised when the docs fail to render.) We should also get a bit more content into the examples in those docs before merging. (Examples look TBD-ish.)

cnunciato:
@jaxxstorm How should users think about this package differently from https://www.pulumi.com/registry/packages/eks/?

community-packages/package-list.json
jaxxstorm: ah yes, good catch!

jaxxstorm:
> @jaxxstorm How should users think about this package differently from [pulumi.com/registry/packages/eks](https://www.pulumi.com/registry/packages/eks/)?

@cnunciato this essentially my replacement for EKS which greatly reduces the complexity of the inputs, has provides an opinionated set of addons and mechanisms like an ingress controller and cert-manager

pierskarsenbarg (approved):

community-packages/package-list.json
```diff
@@ -227,6 +227,10 @@
     {
       "repoSlug": "pulumiverse/pulumi-matchbox",
       "schemaFile": "provider/cmd/pulumi-resource-matchbox/schema.json"
+    },
+    {
+      "repoSlug": "lbrlabs/pulumi-lbrlabs-eks",
+      "schemaFile": "provider/cmd/pulumi-resource-lbrlabs-eks/schema.json"
```
lukehoban, line 233: I don't see a schema.json in this location in the repo.  I'm not an expert on how this works, but does this work?

lukehoban (approved):

community-packages/package-list.json
cnunciato: Yeah, I believe this needs to be this:

```suggestion
      "schemaFile": "schema.yaml"
```

sean1588 (approved):

PR #2796, opened 2023-06-30
add cockroach db provider

jaxxstorm:


cnunciato:
@jaxxstorm You'll need to update the `layout` params in your docs pages to use `package` rather than `overview` and `install` as before.

I know this is a pain and I wish it weren't necessary, but a recent change to the way these pages are build basically forces package authors to do this. Out of empathy, I've [submitted this PR to make the old names still work](https://github.com/pulumi/registry/pull/2796), but it introduces obvious problems of its own, so it'd be better if you could just update those fields (and any scripts or manual workflows you have for this) so we don't keep running into it.

jaxxstorm:
I've updated the package but it won't get the new names until the next version is cut

cnunciato:
@jaxxstorm Ok thanks, appreciate that. I just merged https://github.com/pulumi/registry/pull/2797 so we should be able to get this through now. 🤞

cnunciato (approved):

PR #2797, opened 2023-06-30
Just duplicate the package layout

cnunciato:
A recent change to the way we render our package pages made it mandatory for package authors to update their overview and install pages to use a new layout named `package`. Previously, these pages used the `overview` and `install` layouts, respectively, but those templates have been removed. 

Many authors aren't aware that they have to do this, however, so when they release a new update without having made these changes, the update either fails to render when published or fails to make it through the pipeline at all. This makes for a pretty poor experience as a package author, so this PR simply duplicates the `package` template so that authors can continue using the old layout names until they're able to cut over to the new ones.

To be clear, I definitely don't love having to do this -- but there isn't really another way to paper over it, and I feel bad leaving authors to hit it and blocking publishing because of it. So it might be good to consider pulling it in at least for a while to give folks some time to transition. See https://github.com/pulumi/registry/pull/2796#issuecomment-1615201739 for a bit more context.

### DIFF
--- a/scripts/ci/pull-request.sh
+++ b/scripts/ci/pull-request.sh
@@ -13,6 +13,8 @@ fi
 
 aws sts get-caller-identity
 
-./scripts/ci/validate-packages.sh
+# Commented temporarily to allow package authors to adopt the new layout naming.
+# https://github.com/pulumi/registry/pull/2797
+# ./scripts/ci/validate-packages.sh
 ./scripts/ci/build.sh
 ./scripts/ci/sync.sh


sean1588 (commented):
should we remove the validation that i added if we are going to commit this? these will never get used if we dont because any pr that tries to merge using one of the okd templates will fail the checks.

cnunciato:
@sean1588 If merging this makes sense to you too (feel free to push back on it of course), then yeah, I guess we would have to do that. 

Would be good to know whether this makes sense to you too, though, or if you can think of a better alternative. I thought about a couple of things to get around the duplication -- reading the file contents from the `package` template, explicitly overwriting the `layout` param somehow, using some other standard Hugo layout file as a fallback, etc.), but nothing quite works, in part because these files end up in the Hugo virtual filesystem, so we don't ever get access to them _as files_, unfortunately. It's a bummer all 'round, but I guess I'd rather take the hit myself than let users run into it. What do you think?

sean1588 (approved):

sean1588:
just updating this with our slack convo for context since I replied to you there...

> yeah im not too worried about the duplication aspect of it since this is more of a temporary solution until we can get the authors over to using the new layout. I guess my only concern with this is it will technically still allow the authors to use the old layout file names and I guess not sure how to fully prevent that going forward. But I think this is fine and yeah you can go ahead and remove the validation.

PR #2809, opened 2023-07-03
Fix some CockroachDB docs

cnunciato:
I meant to apply these fixes to https://github.com/pulumi/registry/pull/2798, but merged before I did so. I've also made them on the upstream in https://github.com/lbrlabs/pulumi-cockroach/pull/2.


scottslowe (approved):
LGTM, thank you

PR #2585, opened 2023-05-30
azure native v2 docs

sean1588:
**Add azure-native v2 docs**

This PR is not ready to be merged yet, as we are still making changes to the files here and need to resolve some build issues when it comes to generating the api docs for the v2 version of this package.

fixes https://github.com/pulumi/registry/issues/2569

PR #2846, opened 2023-07-12
Point Azure Native v2 from beta to GA

thomas11:
based on #2585


mikhailshilkov (approved):

thomas11:
@sean1588, eventually we want to replace v1 with v2 in the registry so that there's only one Azure Native again. Can you advise us on how to go about that?

PR #2851, opened 2023-07-12
hide azure v2 from list page since azure-native is now officially in v2

sean1588:
This will hide the separate azure-native-v2 package from the registry list page, since v2 is now officially GA and azure-native now has the v2 docs. The individual docs pages under the `/registry/packages/aure-native-v2` route will remain though so the old v2 routes will remain intact and external links will continue to function. I added an exception here to just not render it on the list page, since we need to keep the metadata file around in order for the individual docs pages for azure-native v2 to continue to get genned.

### DIFF
--- a/themes/default/layouts/partials/registry/package-list.html
+++ b/themes/default/layouts/partials/registry/package-list.html
@@ -82,7 +82,9 @@
             {{ $page := . }}
 
             {{ range (sort $allPackages "title") }}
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
LGTM, but could we get an issue filed to redirect requests for these V2 pages to their new home? Just so we aren't building and carrying around these pages unnecessarily for too long (assuming that sounds like the right thing to do to you also).

PR #2862, opened 2023-07-14
Remove azure v2 beta docs

mnlumi:
Remove the azure v2 beta docs. Now that the redirect is in place and confirmed to be working these can go away now.

### DIFF
--- a/themes/default/layouts/partials/registry/package-list.html
+++ b/themes/default/layouts/partials/registry/package-list.html
@@ -82,9 +82,7 @@
             {{ $page := . }}
 
             {{ range (sort $allPackages "title") }}
-                {{ if ne .name "azure-native-v2" }}
-                    {{ partial "registry/package" (dict "package" . "page" $page) }}
-                {{ end }}
+                {{ partial "registry/package" (dict "package" . "page" $page) }}
             {{ end }}
             <div class="no-results px-4">
                 <p>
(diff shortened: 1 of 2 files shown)


mnlumi:
This is dependent on https://github.com/pulumi/pulumi-hugo/pull/3158 being merged and fully propagated.

mnlumi:
@sean1588 There's several items marked as `<!-- HACK:... >` for azure-native-v2 titles. Should those be included here or handled separately?

```
themes/default/layouts/partials/registry/package-list.html
themes/default/layouts/registry/api.html
themes/default/layouts/registry/installation.html
themes/default/layouts/registry/list.html
themes/default/layouts/registry/overview.html
themes/default/layouts/registry/package.html
```

sean1588:
@mnlumi yeah those can be removed too... but we actually need to keep this in for a bit, so it doesn't break any external links that people may have to the old azure-native-v2 package route. I have an issue opened to put in a redirect for all the routes under /registry/packages/azure-native-v2 and redirect them to the azure-native package. The azure native v2 listing was removed from the registry's package list page though, so users coming to the registry should only see the azure native package. I just kept the sub pages around so it wouldn't break old links. So this can be done once those redirects have been established. I can take this from here though as there is a bit of work that was done to accomodate this in the docs repo as well that also needs to be cleaned up.

cc @mikhailshilkov @thomas11

cnunciato (approved):

Issue #2879, opened 2023-07-19
Serve Azure Native v1 docs


mnlumi:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Please make Azure Native v1 docs available (v1.104.0) in the registry.
- v1 docs are expected to remain for ~6 months
- v1 doc updates are understood to require manual updates and the expectation is that any further updates to v1 will be rare and only in urgent security related updates. There is no further development planned for v1.
- Search should prefer v2 and exclude v1.
- Azure v2 docs should have a link to v1 available so that users that navigate to v2 docs are able to find the v1 docs.
- It should be clear which version (v1, v2) of the docs are being viewed.
- v1 docs should include notifications that v2 is available and recommended.

cnunciato:
@sean1588 Let me know when you've settled on a route for the V1 docs, and I can push a commit onto the PR branch with the changes to filter out V1 results from search.

sean1588:
I have the route configured as /registry/packages/azure-native-v1, similar to how we were routing v2. Testing it out now, but should have a preview up for it soon for you all to test.

mikhailshilkov:
I have a few questions here which I don't really understand from the issue description
- What is our cost of hosting these pages? In my mind, Hugo has some limited capacity that prevented us from hosting full V2 docs, for example. Are we using that capacity here instead of more important docs? Should we spend it on full V2 docs?
- When we remove these docs in 6 months, what will happen with URLs? Will external links be correctly redirected?
- If users google for a specific resource, will we make sure they never land on v1 docs?
- Will v1 docs show up in the registry search and navigation?

I feel like this is a meaningful UX change that we can't easily revert and that we should not land in a rush. Can somebody write down a proposal and plan a call to review/discuss it?

cnunciato:
>If users google for a specific resource, will we make sure they never land on v1 docs?

@mnlumi @mikhailshilkov The PR above adds a `robots.txt` rule to tell search engines not to index these pages.

mnlumi:
Thanks @cnunciato @sean1588 for making these docs available!

Closing the loop from other conversations, when we are ready to deprecate this, we will ensure that the link for https://www.pulumi.com/registry/packages/azure-native-v1 has a redirect so we are not just dropping traffic to these pages.

PR #2882, opened 2023-07-19
add azure v1 docs

sean1588:
issue: https://github.com/pulumi/registry/issues/2879
This PR add the azure native v1 docs to the registry so they will still be accessible for those who are still using the v1 version of this package.

The [PR](https://github.com/pulumi/docs/pull/9505) in docs should be merged first before this goes in.

### DIFF
--- a/themes/default/layouts/partials/registry/package/package-alert.html
+++ b/themes/default/layouts/partials/registry/package/package-alert.html
@@ -78,11 +78,18 @@
             </p>
         </div>
     </div>
+{{ else if (eq $package "azure-native") }}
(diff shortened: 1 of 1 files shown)


cnunciato (approved):

cnunciato:
>The https://github.com/pulumi/docs/pull/9505 in docs should be merged first before this goes in.

@sean1588 Actually I think _this_ PR should be merged first, then the docs PR to pick up what's included here. (I think that's what we just discussed, so just reiterating here.

sean1588:
yup, sorry had to order mixed up in the pr description

PR #2963, opened 2023-08-02
Configure registry to deploy independently

sean1588:
part of: https://github.com/pulumi/registry/issues/3173
This PR configures registry to deploy independently to the test environment. This addresses phase 1 of the build separation work as outlined in the doc linked below. This will enable registry to deploy independently and manage its own infrastructure. Once a PR is merged, a workflow will get kicked off to build the registry docs and deploy it to an s3 bucket living behind a cloudfront cdn that registry manages. For now this is only taking place in the testing account. After we let this live for a bit in testing we can promote this to start deploying to the production account. This is already deployed to the testing environment (this pr just contains the code changes to commit). I have already ran the workflows against the test environment to apply these changes.

doc: https://docs.google.com/document/d/1b765fZyXhrzIhaEnNCC6ojYPrXfwfK86nfTAZwRYWOQ/edit

This PR does the following:

- **adds workflow to deploy to test environment** - Adds a new workflow to deploy to the test environment. This workflow has been configured to deploy to the `pulumi-testing-website` account in AWS. This will get kicked off whenever a PR is merged into this repo. 

- **adds a new pulumi program and stack for test environment so registry can manage its own infra**. This includes managing its own dedicated cloudfront distribution and manages its own buckets. The stack can be found in the pulumi cloud console here: https://app.pulumi.com/pulumi/registry/registry-testing. The docs cloudfront distribution will be updated to add this cloudfront distribution as an origin and serve all requests under the `/registry` route. See this  [PR](https://github.com/pulumi/docs/pull/9619).

- **moves the resourcedocsgen tool from the docs repository to registry (this repo).** The docs repo will no longer be responsible for genning the registry docs, so that tool is now moved here. It still exists in the docs repo at this point, since this will only deploy to the test environemnt when this PR is merged and the current production flow still relies on it being in the docs repo. When we switch this over to prod we can remove it from the docs repository.

- **build and deploys site to s3 bucket.** This repo will now produce its own hugo build and deploy the contents to s3 bucket.

- **host a copy of theme here** - this repo now contains a copy of theme and will produce its own assets build. There is some redundant styling to start with what currently exists in pulumi-hugo, but this is the first step in getting registry to manage its own styling independently. 

The [PR](https://github.com/pulumi/docs/pull/9619) in the docs repo adds a cloudfront origin to reference the registry from there. It also provisions a bucket to house the bundles.

testing env cdn endpoint: d298m5mqqr8wsc.cloudfront.net

Steps to move over to Prod (to be done once we are comfortable with the results in test):
1. create prod stack
2. move to using prod environment variables in workflow
3. open PR in pulumi/home to enable auth via OIDC to AWS production account for registry

### DIFF
--- a/scripts/await-in-progress.js
+++ b/scripts/await-in-progress.js
@@ -45,10 +45,8 @@ async function waitForInProgressRuns() {
 
     if (recent.length > 0) {
         const [ mostRecent ] = recent;
-        // console.log(`Waiting for ${mostRecent.html_url} to complete before continuing.`);
-        // await Promise.resolve(setTimeout(waitForInProgressRuns, 60000)); // One minute.
-        console.log("Temporarily exiting in test environemnt, until we find a way to batch.")
-        process.exit(0);
+        console.log(`Waiting for ${mostRecent.html_url} to complete before continuing.`);
+        await Promise.resolve(setTimeout(waitForInProgressRuns, 60000)); // One minute.
     } else {
         console.log("Continuing.");
     }


zchase (approved):

.github/workflows/testing-deploy.yml
```diff
@@ -0,0 +1,78 @@
...
+          go-version: 1.19.x
+
+      - uses: peaceiris/actions-hugo@v2
+        with:
+          hugo-version: '0.111.0'
+          extended: true
+
+      - name: Configure Git
+        run: |
+          git config --global url."https://${{ secrets.PULUMI_BOT_TOKEN }}:x-oauth-basic@github.com/pulumi/registry".insteadOf "https://github.com/pulumi/registry"
```
cnunciato, line 35: Same here, remove if not needed.

infrastructure/Pulumi.registry-testing.yaml
cnunciato, line 1: Nit, but could we name this `Pulumi.test.yaml` (or suchlike) to stick to conventions. (The whole repo is "registry", so adding the name here is somewhat redundant.)

themes/default/theme/postcss.config.js
```diff
@@ -0,0 +1,99 @@
...
+ * build process for our Sass files, since we run the resulting CSS through PostCSS
+ * at the end to do more transformations.
+ */
+
+module.exports = {
+    plugins: [
+        // TailwindCSS
+        require("tailwindcss")("./tailwind.config.js"),
+
+        // Apply vendor prefixes for CSS features that aren't
```
cnunciato, line 12: Should we just delete all this commented stuff?

cnunciato (commented):
* For non-production stacks, we should add a `robots.txt` directive to `Disallow: *`, as we do with PR previews, to prevent these pages from getting indexed.

* Having to double-commit all CSS and JS here is going to be a huge pain. Do we need to? I would've thought, since the bundles are available as `assets` via the pulumi-hugo module, that we could import and use them here, as we do in pulumi/docs, without having to include all the code as well. (We'd still need to run the same purgecss step we do in pulumi/docs, but other than that, unless Registry needs its own CSS and JS code -- which I don't think it does -- it seems like we should be able to inherit that stuff from pulumi-hugo. Or no?)

.github/workflows/pull-request.yml
```diff
@@ -30,17 +33,28 @@ jobs:
       - name: Check out branch
         uses: actions/checkout@v2
 
+      - name: Configure Git
+        run: |
+          git config --global url."https://${{ secrets.PULUMI_BOT_TOKEN }}:x-oauth-basic@github.com/pulumi/registry".insteadOf "https://github.com/pulumi/registry"
```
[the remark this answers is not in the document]
sean1588: maybe we do not. I was running into some git auth issues at one point when running the build and figured maybe I needed to include that for some reason. I'll pull it out and see if it still passes now.

themes/default/theme/postcss.config.js
sean1588: sure, we can delete it. I honestly pulled it from hugo and wasn't sure if there was a reason we were leaving it in there like for future reference or something.

sean1588:
> Having to double-commit all CSS and JS here is going to be a huge pain. Do we need to? I would've thought, since the bundles are available as assets via the pulumi-hugo module, that we could import and use them here, as we do in pulumi/docs, without having to include all the code as well. (We'd still need to run the same purgecss step we do in pulumi/docs, but other than that, unless Registry needs its own CSS and JS code -- which I don't think it does -- it seems like we should be able to inherit that stuff from pulumi-hugo. Or no?)

@cnunciato @zchase - that is what I did at first, but I thought during the code review on the call we landed on instead just including a copy of theme here and in the future we can prune out what CSS/JS parts do not belong so that registry will be one step closer to being completely independent.

sean1588:
> For non-production stacks, we should add a robots.txt directive to Disallow: *, as we do with PR previews, to prevent these pages from getting indexed.

For specific treatment of prod and non-prod stacks, we are already taking care of this at the top level with the robots.txt file in the hugo repo (i.e. `/registry` shares the same top level as the rest of the site.) So the testing environment will already contain that robots file at the root. However, since registry has its own CDN and is technically "reachable" on its own, I added a robots file here to disallow all in case the bots find our cdn endpoint. Let me know if that make sense.

cnunciato:
> @cnunciato @zchase - that is what I did at first, but I thought during the code review on the call we landed on instead just including a copy of theme here and in the future we can prune out what CSS/JS parts do not belong so that registry will be one step closer to being completely independent.

Yeah, I'm not sure the benefit of that would outweigh the cost, in practice -- at least not yet. When the Registry becomes a completely separate app of its own (as AI has, for example), there may be some benefit to that, but as of now, given it's pretty much only content and the UI all lives (and has been living) in pulumi-hugo, duplicating all that over here seems like it'd only create extra work. Correct me if I'm wrong though -- I may overlooking some reason it's necessary. But if not, I'd recommend leaving it out until we find we actually need it.

cnunciato:
>However, since registry has its own CDN and is technically "reachable" on its own, I added a robots file here to disallow all in case the bots find our cdn endpoint. Let me know if that make sense.

I think this makes sense, yeah -- so if you browsed directly to the Registry CDN endpoint, you'd see `/robots.txt` there with a `Disallow: *`, but this file would never be visible via www. Makes sense to me.

sean1588:
OK, @cnunciato, I addressed the comments and went ahead and pulled out theme. Let me know what you think.

cnunciato (approved):

PR #3015, opened 2023-08-15
feat: add pulumi-esxi-native provider to registry

edmondshtogu:


cnunciato (commented):
Hi @edmondshtogu -- reviewing this now. Couple things I'm seeing:

* Looks like this needs a quick rebase to resolve the conflict above ☝️ 
* One of our build steps seems to be failing [on this line in your `schema.json`](https://github.com/pulumiverse/pulumi-esxi-native/blob/e288af048ef84efdff2d5507a79c78f744fff9c3/provider/cmd/pulumi-resource-esxi-native/schema.json#L9) because `category/on-prem` is invalid: 
    ```
    invalid category tag category/on-prem
    ```
   I'd suggest changing that to `category/infrastructure`, but this can be any one of `cloud/{cloud|database|infrastructure|monitoring|network utility|vcs}` -- or you could also just omit it, as it's technically optional.
* The YAML example on the overview page seems to be missing. You can [remove `yaml` from this line](https://github.com/pulumiverse/pulumi-esxi-native/blob/e288af048ef84efdff2d5507a79c78f744fff9c3/docs/_index.md#L11) if you don't have a YAML example at hand. 

Otherwise, looks good! If you could fix these up and cut a new version tag (after making the docs change, so we can pick that up as well), we should be able to merge this. Thanks!

edmondshtogu:
Hi @cnunciato, thanks for reviewing the PR 🙌🏻 
I did remove the `yaml` example for now and also changed the tag to `category/infrastructure` from the `category/on-prem`

Did dump a new version as well.

Soon I will improve a few other things and release the v1.

cnunciato (approved):

cnunciato:
Thanks for your patience @edmondshtogu! 😅

PR #3034, opened 2023-08-23
Add aci provider to community packages list

danischm:


AaronFriel (commented):
@danischm do you think it would help if your index.md described what ACI is? My first thought was another cloud's "Container Instances" product, but it looks like this is Cisco ACI, a networking solution.

It'll be helpful for users of the provider, our registry, & Pulumi AI if we could add a link and description of what Cisco ACI is. This will also help with users finding this provider and not confusing it with another provider's container instances.

danischm:
It's more than a valid point. :-) I have added a section describing what Cisco ACI is and links.

AaronFriel (approved):

danischm:
Is there anything else I need to do to get it published? Do I need to publish a new release?

AaronFriel:
Ah, I think I see - the docs at v0.0.4 don't include your changes? To update your docs you should tag a new release and we'll pick that up.

PR #3127, opened 2023-09-09
Move robots.txt from layouts into local static

cnunciato:
In https://github.com/pulumi/registry/pull/2963, we added a `robots.txt` file to the repository to prevent Registry pre-prod environments from being indexed by Google. Unfortunately the location of this file (in `themes/default/layouts`) resulted in its being incorporated into the production pulumi.com build, which meant that as of yesterday at 4:50 PM PDT, any Googlebot crawls would have encountered this file and presumably blocked the whole site from being indexed.

Moving the file out of the Hugo module and into the root static folder should prevent it from being pulled into the production build.

From what I can tell, this has not resulted in any loss of results from the Google index, and our most recent content (yesterday's blog post) appears to be there:

image

... but for a time, this file was definitely publicly accessible. (I only noticed it when I went looking to check on AI Answers results in the Google Search console and saw they were being blocked.)

Will get a post-mortem scheduled for this for next week.


sean1588 (approved):
thanks so much for catching this @cnunciato!  So sorry about that, I'm glad it looks like you got to it before the bots did as this could have been pretty bad.

PR #3150, opened 2023-09-14
Deploy registry split to production account

sean1588:
fixes: https://github.com/pulumi/registry/issues/3173

This PR when merged will start deploying the registry infrastructure into the production account. It creates a production stack and configures the prod workflow to deploy.

Merging this PR will not actually cut over traffic to being served from the registry specific infratructure. We will need to update the cloudfront distibution in docs to get the origin added before traffic starts getting routed here.


zchase (approved):
Looks good to me. Just to confirm the next steps are?

1. Cut over traffic
2. Remove registry related stuff from `pulumi-hugo`

sean1588:
> Looks good to me. Just to confirm the next steps are?
> 
> 1. Cut over traffic
> 2. Remove registry related stuff from `pulumi-hugo`

yeah cutting over traffic would be next after we verify infra was successfully stood up to prod. 

as far as removing registry related stuff from pulumi-hugo, are you referring to like the styling in theme?

PR #3210, opened 2023-09-25
Regex on package name should only match until the first hyphen

ringods:
Community member @edmondshtogu [reported](https://pulumi-community.slack.com/archives/C037PV12W6L/p1695321462814079) that the package he contributed, pulumiverse/pulumi-esxi-native, was added to the community packages list, but was not picked up so far for publishing.

Here is a run where it was skipped:

https://github.com/pulumi/registry/actions/runs/6270503792/job/17028537875

It seems that this regex doesn't match correctly:

https://github.com/pulumi/registry/blob/8d9e1c7e3dc83691b4896ad01617704eeebeb15d/.github/workflows/generate-package-metadata.yml#L40

Matching the community repo name against this regex, you get this:

https://regex101.com/r/vc2dRk/1

`group2` get's the value `native` and is then used to check if there are open PRs for this package. At the time of writing, this is an open, but unrelated PR, containing the word `native` in the subject:

https://github.com/pulumi/registry/pull/3076

This PR contains a fix for the regex that no longer matches eagerly until the last hyphen in the repo name, but until the first one. Usually that matches `group1` to `pulumi-` and takes the rest as the package name:

https://regex101.com/r/SN1ACI/1

This is not the only package where the package name is reported incorrectly and would result in strange behaviour. Take [lbrlabs/pulumi-lbrlabs-eks](https://github.com/lbrlabs/pulumi-lbrlabs-eks) for example, the PRs to publish new versions are reported as `eks` instead of `lbrlabs-eks`:

* https://github.com/pulumi/registry/pull/3208
* https://github.com/pulumi/registry/pull/3193

This conflicts with our own EKS package, where pull requests are also reported as eks:

* https://github.com/pulumi/registry/pull/3121
* https://github.com/pulumi/registry/pull/3116

Because community packages are not automatically merged, an open PR from `lbrlabs/pulumi-lbrlabs-eks` currently prevents a new version of `pulumi/pulumi-eks` to be published.

### DIFF
--- a/.github/workflows/generate-package-metadata.yml
+++ b/.github/workflows/generate-package-metadata.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Get provider short name
         id: regex-prov
         run: |
-          if [[ "$REPO_SLUG" =~ (\/.*-)(.*?$) ]]; then
+          if [[ "$REPO_SLUG" =~ (\/[^-]*-)(.*?$) ]]; then
             echo "group2=${BASH_REMATCH[2]}" >> $GITHUB_OUTPUT
           else
             echo "Regex pattern match error."


edmondshtogu (approved):

sean1588 (approved):
thanks @edmondshtogu  for the report!
thank you for providing the fix @ringods  :star:

PR #3212, opened 2023-09-25
Revert "add registry specific sitemap"

sean1588:
Reverts pulumi/registry#3202


sean1588:
I am reverting this. I had paused the automerge workflow to verify this before it actually got deployed and found this registry sitemap is overwriting the current one at the root. 

http://www-production-pulumi-docs-origin-pr-9915-b3180add.s3-website.us-west-2.amazonaws.com/sitemap.xml

cnunciato:
@sean1588 The `sitemap.xml` file needs to live outside `themes/default`. If you put it at the root level, it won't be applied at pulumi/docs. (Sorry I didn't catch this in the review!)

sean1588:
> (Sorry I didn't catch this in the review!)

not your fault... I should have caught this also, good thing I decided to check before it went out since I had a suspicion. Yeah I just need to relocate the file.

PR #3215, opened 2023-09-25
Generate registry algolia search index in registry

sean1588:
This PR generates the registry algolia search index here. This will not do any updates to algolia or be consumed yet. This just starts uploading the index file which will later be consumed by a separate job that will merge this index with the docs one and  push the updates to algolia.

A lot of this code is from the algolia index generation in the docs repo. I removed the non registry related pieces and made a couple adaptations in order to source the package titles from the registry metadata files.

Here is what I changed:
- remove the reliance on the index.json file produced by hugo in docs. this was used soley for the purpose of grabbing the provider titles. I instead adapted this script into just pulling this directly from the package metadata files, since they already live here in this repo.
- removed genning of non registry indexes
- we do not make any update calls to algolia (this will be done by a job that pulls these files from the site)
- do not generate index settings here (we will continue to do that in the docs repo. I believe having two separate settings for each and then attempting to merge them up will be very hard to get right based on the structure of the settings object array)
- upload the generated `search-index.json` file to s3 at `/registry/search-index.json`

generated search index: http://registry--origin-pr-3215-245b5ccf.s3-website.us-west-2.amazonaws.com/registry/search-index.json

related docs [PR](https://github.com/pulumi/docs/pull/9928) to upload the docs specific indexes.

example:

genned object:
image

currently in aloglia production:
image

### DIFF
--- a/scripts/generate-search-index.sh
+++ b/scripts/generate-search-index.sh
@@ -9,7 +9,7 @@ source ./scripts/ci/common.sh
 # to a file named 'search-index.json'.
 node ./scripts/search/main.js "$1"
 
-registry_bucket="$(cat "$(origin_bucket_metadata_filepath)" | jq -r ".bucket")"
+destination_bucket="$(cat "$(origin_bucket_metadata_filepath)" | jq -r ".bucket")"
 
 # Upload the `search-index.json` file to S3 where it can be accessed by the update search index cron.
-aws s3 cp "search-index.json" "s3://${registry_bucket}/registry/search-index.json" --acl public-read --region "$(aws_region)"
+aws s3 cp "search-index.json" "s3://${destination_bucket}/registry/search-index.json" --acl public-read --region "$(aws_region)"


scripts/search/check-urls.js
cnunciato, line 1: I wouldn't have expected to find this code here (this file as well as the shell script), since I'm assuming we'll continue to do this check of the full Algolia index as a pulumi/docs job as we do today. Did you have something else in mind?

scripts/update-search-index.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Run the script that updates the Algolia search index. The value passed into this script denotes
+# the name of the index to be updated (e.g., 'preview' or 'production'). The index will be written
+# to a file named 'search-index.json'.
+node ./scripts/search/main.js "$1"
+
+registry_bucket="$(cat "$(origin_bucket_metadata_filepath)" | jq -r ".bucket")"
+
+# Upload the `search-index.json` file to S3 where it can be accessed by the update search index cron.
+aws s3 cp "search-index.json" "s3://${registry_bucket}/registry/search-index.json" --acl public-read --region "$(aws_region)"
```
cnunciato, line 15: Suggest a slight tweak to the name just to make it a bit clearer these don't all go to one bucket:

```suggestion
destination_bucket="$(cat "$(origin_bucket_metadata_filepath)" | jq -r ".bucket")"

# Upload the `search-index.json` file to S3 where it can be accessed by the update search index cron.
aws s3 cp "search-index.json" "s3://${destination_bucket}/registry/search-index.json" --acl public-read --region "$(aws_region)"
```

scripts/update-search-index.sh
cnunciato, line 1: I'd suggest we name this something else, like `generate-search-index`, so it's clearer it's not actually updating anything.

scripts/ci/push.sh
```diff
@@ -9,3 +9,4 @@ set -o errexit -o pipefail
 node ./scripts/await-in-progress.js
 ./scripts/ci/run-pulumi.sh update
 ./scripts/ci/make-s3-redirects.sh
+./scripts/update-search-index.sh
```
cnunciato, line 12: Could we do this before the S3 redirect step? Only because this can take a while.

It does make me wonder whether we could do this before the Pulumi run, though. Couldn't we?

scripts/search/check-urls.js
sean1588: no I didn't mean to keep these in. sorry, meant to remove these files!

scripts/update-search-index.sh
sean1588: agreed!

scripts/ci/push.sh
sean1588: yeah we definitely could. good call!

sean1588:
thanks @cnunciato for the feedback! All the suggestions make sense to me! I updated the files.

cnunciato (approved):

PR #3232, opened 2023-09-27
Update terminology for Event Store Cloud

lukehoban:


sean1588 (approved):

sean1588:
@lukehoban - this will update those docs for now, but the next time event store cloud publishes a release from the upstream, it will update to match the docs with what is in the upstream repo.  

The change will also need to be made there in the upstream in order for this not to happen: https://github.com/EventStore/pulumi-eventstorecloud/blob/main/docs/_index.md

lukehoban:
> The change will also need to be made there in the upstream in order for this not to happen: https://github.com/EventStore/pulumi-eventstorecloud/blob/main/docs/_index.md

Aha - yeah - good point.  I'll open a PR to that repo, but looks like we'll need to work with the partner on getting that merged.

PR #3249, opened 2023-09-29
Add registry sitemap

sean1588:
This PR adds the registry specific site map. The docs [PR](https://github.com/pulumi/docs/pull/9943) must be merged first, and workflows paused before this goes in. That PR will add the registry site map to the sitemap index file and stop importing registry altogether. If not, the sitemap here will overwrite the one in docs and looks like there is no way around this (at least from what I understand from the [hugo documentation](https://gohugo.io/templates/sitemap-template/) stating that this file need to be placed under `themes/default`).  So this needs to go out around the same time as the docs PR. Once both PRs are merged we will be able to turn back on the workflows. I also added an exception to filter out the `azure-native-v1` packages while I was in there since that is something we recently noticed.

http://registry--origin-pr-3249-00706221.s3-website.us-west-2.amazonaws.com/registry/sitemap.xml


themes/default/layouts/_default/sitemap.xml
cnunciato, line 1: Have you tried putting this in `layouts/_default` (from the root)? I'm pretty sure this should work there too (looks that way from a quick local test), which would mean we could merge this now, before the docs PR, to have the Registry sitemap already in place by the time the updated sitemap-index lands.

sean1588: yeah this works. Ialso  tested it out locally and ran a docs build importing from this commit and works fine.

cnunciato (approved):

PR #3258, opened 2023-10-02
Add ACL for registry cloudfront logging

sean1588:
This PR adds the ACL needed to give the `awslogsdelivery` account write access to the logs bucket. This has been done in the docs repo. Not sure if this is  necessarily "needed" as all the traffic routed here comes through the CloudFront CDN that docs manages, but probably doesn't hurt to have it.


cnunciato (approved):
Not a bad idea to have this also, yep. 👍

PR #3263, opened 2023-10-02
Add registry index json

sean1588:
Produce index.json like we are in the docs repo for the Hugo site. It turns out we actually do need this file in order to get hold of the top level pages in registry. I attempted to generate the index all the registry pages using the metadata files here, but actually ended up needing the overall site index.


sean1588:
@cnunciato - I'm going to go ahead and just push this through so we can get this rolling, since it doesn't affect anything regarding the deployment of the site or any of the content. It just affects the search-index.json file that gets produced which we are not officially using to push the indexes in production yet.

PR #3267, opened 2023-10-03
Add theme to registry

sean1588:
Add a copy of theme to registry. This will contain some duplicate stying with what is in the `pulumi-hugo` repo for the time being (i.e. there are some marketing and docs site styling in there). We should do some work to pull this out at some point. This however will allow us to build registry specific `css` and `js` bundles and remove the reliance on `theme` in the `pulumi-hugo` repository. 

I did the following here:
- Copy theme directory here
- I adjusted the bundle names to include registry in them for both javascript and js. see [webpack file](https://github.com/pulumi/registry/pull/3267/files#diff-3de65d62fedc98063eabef7d9c6b571f36eb6fd84735173775ae13115587b937R14) and [assets.html](https://github.com/pulumi/registry/pull/3267/files#diff-331d7ebf76a633b9637ff0a4e5ecea80b83c9c00150701667248a605d8711e38R12-R26)
- Added make targets for `build-assets`, `serve-all`, and `serve-assets`
- mount assets under registry in the config file
- updated ensure script to yarn install theme

This is deployed to the test environment and you can verify the bundle names there.

Once this PR gets merged, all CSS and JS updates that are made to target registry should be done here.


cnunciato (commented):
@sean1588 Can you also add a worklow to commit these asset bundles back to the master branch if they differ? Or even add this to the PR workflow to commit back to the current branch (which might be better, since we'll deploy on every commit to master)? Otherwise we'll have to do this manually and potentially ship bad CSS/JS.

sean1588:
> Or even add this to the PR workflow to commit back to the current branch (which might be better, since we'll deploy on every commit to master)

I could see this being annoying if it is slapping commits on PRs, especially if you have work that is in progress, because you will need to constantly be re-pulling locally in order to continue commiting in order to sync up the branch with the upstream so I went and added it to the push workflow.

PR #3279, opened 2023-10-04
Add link checker to registry

sean1588:
This will run the link checker against registry specific pages and use the registry specific sitemap to crawl.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,10 +7,13 @@
     "license": "MIT",
     "dependencies": {
         "@fullhuman/postcss-purgecss": "^4.0.3",
+        "@slack/web-api": "^5.12.0",
+        "broken-link-checker": "^0.7.8",
         "concurrently": "^6.0.0",
         "cssnano": "^5.0.8",
         "cypress": "^4.11.0",
         "glob": "^7.2.0",
+        "http-server": "^0.12.1",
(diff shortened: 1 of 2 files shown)


scripts/link-checker/check-links.js
```diff
@@ -0,0 +1,364 @@
...
+        "https://api.pulumi.com/",
+        "https://github.com/pulls?",
+        "https://github.com/pulumi/docs/edit/master",
+        "https://github.com/pulumi/docs/issues/new",
+        "https://github.com/pulumi/registry/edit/master",
+        "https://github.com/pulumi/registry/issues/new",
+        "https://github.com/pulumi/pulumi-hugo/edit/master",
+        "https://github.com/pulumi/pulumi-hugo/issues/new",
+        "https://www.linkedin.com/",
+        "https://linkedin.com/",
```
sean1588, line 201: we _could_ probably prune out this list if we really wanted, but not sure which ones specifically are linked from docs and which are linked from registry pages, so I just left most of them in.

cnunciato (approved):

scripts/link-checker/check-links.js
cnunciato: Yeah, we can do this later. Good callout though.

PR #3307, opened 2023-10-10
Revert "Publish Package Metadata awsx@v2.0.0"

mikhailshilkov:
Reverts pulumi/registry#3305


thomas11 (approved):
will probably clash with #3306 but we can sort that out later

PR #3340, opened 2023-10-16
Fix text in table wrapping

sean1588:
fixes: https://github.com/pulumi/docs/issues/9920

Moving this over from this [PR](https://github.com/pulumi/pulumi-hugo/pull/3415) in hugo, since registry has its own theme now.


sean1588:
going to go ahead and push this through. this PR was already approved when it was open against pulumi-hugo.

PR #3350, opened 2023-10-17
Fix Azure Native version guide for v2

AaronFriel:
Updates the API versioned Keyvault token and imports to one exported by Azure Native V2, fixes the Go imports to use the Go submodules at `github.com/pulumi/pulumi-azure-native-sdk`.


thomas11 (approved):
LGTM, thanks! The new API version exists and the Go import paths match current code samples.

PR #3366, opened 2023-10-20
Add a pull request template

cnunciato:
Given how many things must be done before a new package can be published (and meet the bar for quality), I thought it might help to have a PR template with a checklist to work by. Here's a suggestion -- feedback and suggestions welcome of course!

Along with this, we may want to tweak the `CODEOWNERS` file to require reviews from the Registry (AI?) and Docs teams where applicable.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,12 +10,13 @@ If this pull request adds a new package:
 - [ ] The package metadata file, if present, contains: 
   - [ ] a supported category (one of `Cloud`, `Infrastructure`, `Network`, `Database`, `Monitoring`, or `Utility`).
   - [ ] a description that explains what the package does.
-  - [ ] a valid logo URL to an SVG whose dimensions conform to the others in this repo.
+  - [ ] a valid logo URL that points to a PNG whose dimensions conform to the others in this repo (e.g., 100x100).
   - [ ] a version number prefixed with `v` that corresponds with a valid GitHub release.
-- [ ] The package repo contains an Overview doc that includes:
+- [ ] The package repo contains an Overview doc (`/docs/_index.md`) that includes:
   - [ ] a brief explanation of what the package is and what it does.
   - [ ] at least one representative example in all supported languages.
-- [ ] The package repo contains an Installation and Configuration doc that includes:
+  - [ ] a front-matter property for the `layout` set to `package`.
+- [ ] The package repo contains an Installation and Configuration doc (`/docs/installation-configuration.md`) that includes:
   - [ ] links to SDKs in all supported languages.
   - [ ] a copyable command for installing the resource plugin if necessary.
   - [ ] an example of configuring the provider with `pulumi config set`.


.github/pull_request_template.md
```diff
@@ -0,0 +1,24 @@
+## Description
+
+<!-- A brief description of the PR here. -->
+
+## Adding a new package?
+
+If this pull request adds a new package:
+
+- [ ] The package's schema URL is correct.
+- [ ] The package metadata file, if present, contains:
```
sean1588, line 10: I think this is good, but one thing to note is when it comes to the community contributions, I believe this is all automated, since all they do is open one PR to add the package to the package-list.json file. From that point forward the rest is an automated process. So the metadata file gets automatically generated and auto merged in these cases. So there might not be a way for us to catch these at PR time when it comes to community packages We could however validate these things using the schema file in the upstream repo since that is where the metadata file gets genned from when it comes to these cases.

.github/pull_request_template.md
```diff
@@ -0,0 +1,24 @@
...
+
+- [ ] The package's schema URL is correct.
+- [ ] The package metadata file, if present, contains: 
+  - [ ] a supported category (one of `Cloud`, `Infrastructure`, `Network`, `Database`, `Monitoring`, or `Utility`).
+  - [ ] a description that explains what the package does.
+  - [ ] a valid logo URL to an SVG whose dimensions conform to the others in this repo.
+  - [ ] a version number prefixed with `v` that corresponds with a valid GitHub release.
+- [ ] The package repo contains an Overview doc that includes:
+  - [ ] a brief explanation of what the package is and what it does.
+  - [ ] at least one representative example in all supported languages.
```
sean1588, line 17:
```suggestion
  - [ ] at least one representative example in all supported languages.
  - [ ] a front-matter property for the `layout` set to `package`.
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,24 @@
...
+
+If this pull request adds a new package:
+
+- [ ] The package's schema URL is correct.
+- [ ] The package metadata file, if present, contains: 
+  - [ ] a supported category (one of `Cloud`, `Infrastructure`, `Network`, `Database`, `Monitoring`, or `Utility`).
+  - [ ] a description that explains what the package does.
+  - [ ] a valid logo URL to an SVG whose dimensions conform to the others in this repo.
+  - [ ] a version number prefixed with `v` that corresponds with a valid GitHub release.
+- [ ] The package repo contains an Overview doc that includes:
```
sean1588, line 15:
```suggestion
- [ ] The package repo contains an Overview doc (`/docs/_index.md`) that includes:
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,24 @@
...
+  - [ ] a brief explanation of what the package is and what it does.
+  - [ ] at least one representative example in all supported languages.
+- [ ] The package repo contains an Installation and Configuration doc that includes:
```
sean1588, line 18:
```suggestion
- [ ] The package repo contains an Installation and Configuration doc (`/docs/installation-configuration.md`) that includes:
```

sean1588 (approved):
Thanks Christian, this is great! Just a couple suggestions if you want to take them. I think we could fairly easily automate a lot of these checks when a new package is submitted (e.g. the schema file url, metadata file, and existence of docs pages) but this should all be good for now. Though some of the stuff like reviewing the docs content is fairly subjective, so maybe it just makes sense to just do manual verification for all of these.

.github/pull_request_template.md
cnunciato: What I've been doing for these community-packages PRs -- and it's unfortunately pretty laborious -- is looking at the schema file and the repo and then checking that all of the necessary things (docs, logo, etc.) exist before approving the PR, for example here:

https://github.com/pulumi/registry/pull/3015

It's a painful process, but given the way we've got this set up now -- where the list-item PR comes through first, we merge that without much (any?) review, and then everything else after that that is automated and gets no review -- I'm not sure there's a much better option. At least not until we have some automation in place to check for the existence of some of these items before allowing a community-packages PR to be merged.

.github/pull_request_template.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Adding a new package?
+
+If this pull request adds a new package:
+
+- [ ] The package's schema URL is correct.
+- [ ] The package metadata file, if present, contains: 
+  - [ ] a supported category (one of `Cloud`, `Infrastructure`, `Network`, `Database`, `Monitoring`, or `Utility`).
+  - [ ] a description that explains what the package does.
+  - [ ] a valid logo URL to an SVG whose dimensions conform to the others in this repo.
```
ringods, line 13: Is this about the `logoURL` in the Pulumi schema file? If so, SVG is not supported as the icon for a .NET nuget package. If you do, the publishing will fail with this error:

`error: Response status code does not indicate success: 400 (Unsupported icon image format. Only PNG and JPG images are supported.).`

https://github.com/pulumiverse/pulumi-talos/actions/runs/6596863737/job/17923421848

cnunciato: Ah yes, PNG works. Thanks for pointing that out!

cnunciato, line 13:
```suggestion
  - [ ] a valid logo URL that points to a PNG whose dimensions conform to the others in this repo (e.g., 100x100).
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,25 @@
...
+## Adding a new package?
+
+If this pull request adds a new package:
+
+- [ ] The package's schema URL is correct.
+- [ ] The package metadata file, if present, contains: 
+  - [ ] a supported category (one of `Cloud`, `Infrastructure`, `Network`, `Database`, `Monitoring`, or `Utility`).
+  - [ ] a description that explains what the package does.
+  - [ ] a valid logo URL that points to a PNG whose dimensions conform to the others in this repo (e.g., 100x100).
+  - [ ] a version number prefixed with `v` that corresponds with a valid GitHub release.
```
ericrudder, line 14: [ ] The provider correctly identifies the source of creation being "Pulumi" (rather than TF)

@cnunciato You are welcome to re-word this, however I do want to make sure we get this right as many partners judge us by "impact" and this is how their dashboards light up.

joeduffy: Specifically, this is the `User-Agent` string.

PR #3397, opened 2023-10-25
Build assets into static/assets/registry

cnunciato:
Builds CSS and JS bundles into `static/registry/css` and `static/registry/js` and loads them from there. This change will allow us to remove the dependency on the exported `bundlesS3BucketName` from pulumi/docs and just serve the assets as regular static assets out of the same bucket as the website.

### DIFF
--- a/.github/workflows/testing-deploy.yml
+++ b/.github/workflows/testing-deploy.yml
@@ -1,8 +1,5 @@
 name: Build and deploy - test environment
 on:
-  pull_request:
-    branches:
-      - master
   push:
     branches:
       - master


themes/default/theme/webpack.config.js
```diff
@@ -13,7 +13,7 @@ module.exports = function (env, { mode }) {
         output: {
             filename: "[name]-registry.js",
             chunkFilename: "chunk-[id]-registry.js",
-            path: `${process.cwd()}/../assets/js`,
+            path: `${process.cwd()}/../assets/registry/js`,
```
sean1588, line 16: when the build runs, will it produce this under the `/registry` route though, or will it be under the root, i.e. `/assets/registry/js`? I'm guessing you tested it out locally though.

cnunciato: They'll end up at `/registry/css` and `/registry/js`, yeah. I did test locally so it _should_ be okay, but we'll run a test deployment to make sure nothing funny happens at the CDN level.

.github/workflows/testing-deploy.yml
```diff
@@ -1,5 +1,8 @@
 name: Build and deploy - test environment
 on:
+  pull_request:
```
sean1588, line 3: leaving an approval, just remember to remove this

sean1588 (approved):
:rocket: 

all looks good to me and works in test.

.github/workflows/testing-deploy.yml
```diff
@@ -1,5 +1,8 @@
...
+    branches:
+      - master
```
cnunciato, line 5:
```suggestion
```

PR #3496, opened 2023-11-10
Reorganize content to be more consistent across pages

toriancrane:
## Description

This PR reorganizes the content on the installation and config pages for the major providers to make them more consistent.

Fixes [this issue](https://github.com/pulumi/docs/issues/11567).


scottslowe:
I know the AWS Native provider doesn't support `skipMetadataApiCheck` (i.e., doesn't support authenticating via EC2 instance metadata), should we add a brief statement to that effect to the AWS Native installation and configuration page? Otherwise, the user will only know if that scroll all the way down to the bottom of the page to see that `skipMetadataApiCheck` isn't yet supported.

scottslowe (approved):
I love that we are bringing some consistency to these pages! Looks good!

PR #3552, opened 2023-11-23
Add examples shortcodes

cnunciato:
These shortcodes are purpose-built for Registry API docs examples, so they really belong here -- and their happen names collide with some names I'd like to be able to use in the docs. Adding them should allow me to replace them in pulumi-hugo.


cnunciato:
Actually -- I may not need to do exactly this. Closing for now.

PR #3719, opened 2024-01-03
Skip logo checks

### LINKED ISSUE #3718: Ignore failed logo checks
Our newly added link checker is currently failing on packages that don't have logo URLs. We should add an exception to tolerate these, since the UI currently handles them automatically (by [falling back to a placeholder icon](https://github.com/pulumi/registry/blob/6c308c3fe471b496a7430e321f0e081dc02584f9/themes/default/layouts/partials/registry/package/icon.html#L6)).

cnunciato:
Skips checking logo URLs with the link checker (which currently errors on packages that don't have `logoURLs`). Since we have [fallback code in the UI](https://github.com/pulumi/registry/blob/99825a368bddae5e772c9d935a2e907c114b8a31/themes/default/layouts/partials/registry/package/icon.html#L6) to handle these missing logos (we swap in placeholder icons client-side), the 404 notifications generated by the link checker are non-actionable, so this PR just bypasses them.

Fixes #3718.

### DIFF
--- a/scripts/link-checker/check-links.js
+++ b/scripts/link-checker/check-links.js
@@ -185,10 +185,12 @@ async function onComplete(brokenLinks) {
       although valid and publicly available, is reported as a broken link.
     - A number of synthetic illustrative links come from our examples/tutorials.
     - GitLab 503s for requests for protected pages that don't contain certain cookies.
+    -
 */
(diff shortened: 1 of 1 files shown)


scottslowe (approved):
Looks reasonable to me, approved

PR #3712, opened 2024-01-03
Remove unsupported note for `[allowed|forbidden]AccountIds`

sravaniv-3cortex:
## Description

Update the configuration options docs to reflect the support added in https://github.com/pulumi/pulumi-aws-native/pull/508 for `allowedAccountIds` and `forbiddenAccountIds`.

PR #3739, opened 2024-01-09
Open PR check should not match only on the short provider name

ringods:
The check for a new version of the `port` package detected there was an open PR:

https://github.com/pulumi/registry/actions/runs/7457082927/job/20292272734

This is not the case. It halted here because the `contains` check on `port` matches the word `unsupported` from this PR:

https://github.com/pulumi/registry/pull/3712

We consistently publish new provider releases with the prefix `Publish Package Metadata`. This PR updates the value to match to be `Publish Package Metadata <short provider name>` to filter out these false positives.

### DIFF
--- a/.github/workflows/generate-package-metadata.yml
+++ b/.github/workflows/generate-package-metadata.yml
@@ -46,7 +46,7 @@ jobs:
         env:
           REPO_SLUG: ${{ matrix.repoSlug }}
       - name: Check if there is already an open PR
-        if: contains(needs.generate-packages-list.outputs.prs, steps.regex-prov.outputs.group2 )  
+        if: contains(needs.generate-packages-list.outputs.prs, "Publish Package Metadata ${{steps.regex-prov.outputs.group2}}" )  
         id: skip-run
         run: echo "skip=1" >> $GITHUB_OUTPUT
       - name: Check out registry repo


.github/workflows/generate-package-metadata.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         env:
           REPO_SLUG: ${{ matrix.repoSlug }}
       - name: Check if there is already an open PR
-        if: contains(needs.generate-packages-list.outputs.prs, steps.regex-prov.outputs.group2 )  
+        if: contains(needs.generate-packages-list.outputs.prs, "Publish Package Metadata ${{steps.regex-prov.outputs.group2}}" )
```
sean1588, line 49: I wonder if it would make sense to abstract the prefix into a variable in the environment if possible so this will stay consistent with [line 91](https://github.com/pulumi/registry/pull/3739/files#diff-8bc697d3942ce672edeca3685a03be4b144f21c8b2c2dea50fb95cf4d6556d0fR91) and make it a little less prone to error down the road if any of this changes... though if it is not super straight forward to do this or takes up a lot of time, disregard this.

sean1588 (approved):
Thanks so much @ringods for doing this and all the contributions you make to keep our registry running well!

PR #3746, opened 2024-01-10
bump port version to 1.8.0

sean1588:
## Description

Manually bumping the version of the port package since there seems to be an issue with the packages job where it is not picking up the new version. This will force our build to build docs for 1.8.0.


dirien (approved):
LGTM!

PR #3890, opened 2024-02-01
deprecate packages from lbrlabs

jaxxstorm:
I will no longer be maintaining pulumi packages for these providers, so removing them from the registry


iwahbe (commented):
I'll take carrying this issue forward.

iwahbe:
I'm going to close in favor of https://github.com/pulumi/registry/pull/3890.

PR #3920, opened 2024-02-06
Remove note about partial previews

cnunciato:
I believe this note is no longer correct.


cnunciato:
Gah! Expected this to merge only after a review, but looks like it happened automatically. I'll look into supporting GitHub's auto-merge option on this repo for this purpose, as it may not be well suited to the current setup here.

PR #3984, opened 2024-02-17
Upgrade Cypress to latest, run tests without isolation

cnunciato:
Upgrades Cypress to the latest version and sets the [`testIsolation` option](https://docs.cypress.io/guides/core-concepts/test-isolation) to `false` for the API docs tests. I did this after noticing that it took [over 6 hours](https://github.com/pulumi/registry/actions/runs/7935714809) (!) to run these tests with isolation enabled, and figured since we don't actually _need_ pristine isolation between pageviews, we can run these tests a lot faster by turning it off. Added a comment with details and a link for more information.


sean1588 (approved):
makes sense to me 🚀

PR #4025, opened 2024-02-23
improve open pr matching for package updates

sean1588:
## Description

We have issues where package names collide with titles of open PRs. This causes the update to the package to be skipped in these circumstances. An example of this that has occurred recently concerns the port provider. The word `port` is found in an open PR with the word `Support` in the title. Since port is a substring of support, our job thinks there is already a PR open to update the provider. This PR does a quick fix to improve this a bit by checking that there is a leading whitespace before the provider name. I tested this to the best of my ability by running the workflow against my branch and with the added whitespace, the `Check if there is already an open PR` stage of the job the which sets a variable to skip for this package, `skip=1`, is now longer being set, since I have added the whitespace (i.e. the variable is not set therefore telling the job not to skip the package)

### DIFF
--- a/.github/workflows/generate-package-metadata.yml
+++ b/.github/workflows/generate-package-metadata.yml
@@ -38,7 +38,7 @@ jobs:
         id: regex-prov
         run: |
           if [[ "$REPO_SLUG" =~ (\/[^-]*-)(.*?$) ]]; then
-            echo "group2=${BASH_REMATCH[2]}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


ringods:
Related to #3739

cnunciato (approved):

iwahbe:
@sean1588 Is this PR still relevant?

The providers team had a similar problem, and solved it by only having automation respond to upgrade PRs from @pulumi-bot.

iwahbe:
I'm closing this PR, since it doesn't seem likely to merge. We can always resurrect it later if we want to merge it.

PR #4054, opened 2024-02-28
Fix the API docs section-ordering test

cnunciato:
Fixes a slight bug in the API docs section-ordering test. Currently, the test assumes that all pages contain the same set of sections, but it's possible and expected that some pages won't have everything. (Resource pages that don't contain any nested types, for example, won't have a Supporting Types section.) This change makes an adjustment to that test to ensure that on pages that don't contain every possible section, the sections that are presented are presented in the right order.

Also tacks on a fix to allow for alternate H1s, as not all resources conform to the same conventions.


sean1588 (approved):
Thanks Christian! this looks good to me!

PR #4057, opened 2024-03-01
Update managing-resources-with-server-side-apply.md

jkodroff:
Clarify the behavior of the patchForce annotation.

### DIFF
--- a/themes/default/content/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply.md
+++ b/themes/default/content/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply.md
@@ -149,7 +149,7 @@ Users may opt into the previous default Client-side Apply (CSA) mode by setting 
 
 With Server-Side Apply mode enabled, you can "upsert" a resource; that is, update a resource if it already exists, or create it if it does not yet exist.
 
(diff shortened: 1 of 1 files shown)


themes/default/content/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply.md
```diff
@@ -147,7 +147,9 @@ Users may opt into the previous default Client-side Apply (CSA) mode by setting
 
 ## Upsert a Resource
 
-With Server-Side Apply mode enabled, you can "upsert" a resource; that is, update a resource if it already exists, or create it if it does not yet exist. You can set the `pulumi.com/patchForce` annotation if you want to force an override for any conflicts with an existing version of the resource.
+With Server-Side Apply mode enabled, you can "upsert" a resource; that is, update a resource if it already exists, or create it if it does not yet exist.
+
+If the resource already exists, that default behavior is for any conflicting values to be resolved with the value from the patch being applied (that is, the new version of the resource). If you would prefer to resolve conflicts using the value of the existing resource, add the `pulumi.com/patchForce` annotation:
```
EronWright, line 152:
```suggestion
If the resource already exists, the default behavior is for any conflicting values to be resolved with the value from the patch being applied (that is, the new version of the resource). If you would prefer to resolve conflicts using the value of the existing resource, add the `pulumi.com/patchForce` annotation:
```

EronWright (changes requested):
The explanation seems wrong - `patchForce` is designed to force override on conflict.

PR #4066, opened 2024-03-04
Add e2e test bucket and policy for cross account access

sean1588:
## Description

This PR adds an s3 bucket to house the e2e test results and also adds the policy to enable cross account access to that bucket from the data-production aws account. I also exported the bucket name, `e2eTestsBucketName`, so it can be reference via stack reference. Once this is merged, I'll adapt the script to upload the test results to reference the bucket names from the stack outputs.


cnunciato (approved):
LGTM.

PR #4163, opened 2024-03-20
Limit the height of code blocks

sean1588:
## Description

This PR limits the height of example code blocks. With the new construtor syntax, the code block at times can get very long, taking up a lot of space on the page, so this limits the height. There were some other minor rendering changes made to the template structure to get the syntax highlighting to work, but I bundled those in with Zaid's PR. This PR is not dependent on that going in.

 This will also apply generally to all our code blocks, which I think is good, especially since we no longer collapse examples at the top of resource pages, this will help manage real estate on the page. This will limit their heights to 680px afterwhich the block will be scrollable.

Use this preview link as it contains built pages for the AWS provider: http://registry--origin-pr-4163-afeab115.s3-website.us-west-2.amazonaws.com/registry.

### DIFF
--- a/themes/default/theme/stencil/src/components/choosable/choosable.scss
+++ b/themes/default/theme/stencil/src/components/choosable/choosable.scss
@@ -7,6 +7,9 @@ pulumi-choosable {
         // rendered inline as well.
         &.active {
             display: inherit;
+            div.highlight {
+                max-height: 680px;
+            }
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
LGTM -- I'm assuming all that's needed here is the CSS and everything else is just pinning?

Issue #4202, opened 2024-03-26
`Coming soon!` misrenders with the new language chooser


iwahbe:
### What happened?

I was cleaning up our docs bug backlog, and I noticed this new issue ([here](https://www.pulumi.com/registry/packages/scaleway/api-docs/kubernetescluster/)):

Java is missing docs, so it has `Coming soon!`, but it's misrendered. In fact, it's java's render and then YAML's render, so YAML is empty (missing its own `Coming soon!` block).

<img width="736" alt="Screenshot 2024-03-26 at 10 55 25 AM" src="https://github.com/pulumi/registry/assets/22222529/84775428-eb9c-45e7-8ac5-6c59408f1278">
<img width="736" alt="Screenshot 2024-03-26 at 10 55 31 AM" src="https://github.com/pulumi/registry/assets/22222529/1597f46e-2733-489a-9179-9124e565ccb0">

### Example

https://www.pulumi.com/registry/packages/scaleway/api-docs/kubernetescluster/

### Output of `pulumi about`

NA

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

iwahbe:
I'm actually pretty sure I know the root cause here. I'll take this one.

PR #4232, opened 2024-03-28
Bump p/p version

guineveresaenger:
Followup to https://github.com/pulumi/registry/issues/4202 to propagate docsgen fixes in pu/pu.


sean1588:
thanks @guineveresaenger!! can we hold off on bumping this now though? I am waiting on getting some of the new constructor syntax formatting changes into pu/pu before we update the version, since right now it is in a state where we want to make some more adjustments before we ship it live.

iwahbe (approved):
@sean1588 I defer to you when we should bump the version. My 2 cents is that we can take the fix now and then bump again later.

sean1588:
going to close this assuming we will get this bumped to the latest in the next day. If not I'll reopen this and push this out there, because at there are many valuable fixes wrapped up in this release and may not be worth waiting.

PR #4307, opened 2024-04-10
Adjust api docs cypress tests

sean1588:
## Description

Making some adjustments to cypress tests to address some failures we are seeing. Some failures we are seeing are due to an issue in some of the logic of these tests. 

This fixes the supporting types check for the individual languages to ensure we skip this test when the resource does not contain supporting types. I also made an adjustment to the page title check. This check was failing because some of the resource titles seem to contain a trailing space causing there to be an extra space in between the resource name and the `|`, which causes this test to fail. I made the test a bit less strict since this is a minor issue, and honestly am not sure where this mysterious space is coming from on a couple of these one off pages.

### DIFF
--- a/cypress/e2e/api-docs.cy.js
+++ b/cypress/e2e/api-docs.cy.js
@@ -81,7 +81,10 @@ describe("Registry", () => {
                 });
 
                 it("has the correct page title", () => {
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
LGTM!

PR #4358, opened 2024-04-18
enable skipped tests

sean1588:
## Description

Enables some more skipped tests. These probably aren't perfect, but was quite difficult to get right since none of the content is nested in sections and ~everything seems to be a sibling of everything else at the top level, so I ended up just having to use css operators to verify the dom structure. This isn't perfect as far as testing accuracy, but hopefully good enough. I wasn't able to verify all pages locally since the electron process keeps crashing. I suspect we will run into the same issue in the pipeline, but we can try this out and see.

### DIFF
--- a/cypress/e2e/api-docs.cy.js
+++ b/cypress/e2e/api-docs.cy.js
@@ -164,9 +164,9 @@ describe("Registry", () => {
                         });
 
                         it.skip("renders a full example in all languages", () => {
-                          // We may disable this section for pages that fail to convert, so some
-                          // examples may not have this section once this is done.
-                          // https://github.com/pulumi/registry/issues/4320
+                            // We may disable this section for pages that fail to convert, so some
+                            // examples may not have this section once this is done.
+                            // https://github.com/pulumi/registry/issues/4320
                         });
                     });
 


cypress/e2e/api-docs.cy.js
```diff
@@ -151,22 +151,22 @@ describe("Registry", () => {
...
                         });
                     });
 
                     describe("Create section", () => {
-                        it.skip("renders a minimal example in all languages", () => {
-                            // Markup structure still TBD.
-                            // https://github.com/pulumi/pulumi/issues/14675
+                        it("renders a minimal example in all languages", () => {
+                          // renders the minimal example right after the `Constructor syntax` heading
+                          cy.get(container).find("#constructor-syntax + div > pulumi-chooser")
```
cnunciato, line 163:
```suggestion
                            // renders the minimal example right after the `Constructor syntax` heading
                            cy.get(container).find("#constructor-syntax + div > pulumi-chooser")
```

cypress/e2e/api-docs.cy.js
```diff
@@ -151,22 +151,22 @@ describe("Registry", () => {
...
                         });
 
                         it.skip("renders a full example in all languages", () => {
-                            // Markup structure still TBD.
-                            // https://github.com/pulumi/pulumi/issues/14675
+                          // We may disable this section for pages that fail to convert, so some
+                          // examples may not have this section once this is done.
+                          // https://github.com/pulumi/registry/issues/4320
```
cnunciato, line 169:
```suggestion
                            // We may disable this section for pages that fail to convert, so some
                            // examples may not have this section once this is done.
                            // https://github.com/pulumi/registry/issues/4320
```

cnunciato (approved):

PR #4400, opened 2024-04-25
Update GCP ESC configuration instructions

VenelinMartinov:
## Description

fixes https://github.com/pulumi/pulumi-gcp/issues/1815

The instructions on using ESC with GCP were misleading. It suggested using the `accessToken` provider config which causes issues for users once the credentials expire, since the token is stored in state. This then means the stack can no longer be updated even after changing the credentials.

Instead users should use env vars - this does not suffer from the same issues as the env vars are re-read each time, so expired credentials are never stored.

I've also deleted the section which suggests using config for everything - this suffers from the same issue of storing expired credentials in state.

## Adding a new package?

If this pull request adds a new package:

- [ ] The package's schema URL is correct.
- [ ] The package metadata file, if present, contains:
  - [ ] a supported category (one of `Cloud`, `Infrastructure`, `Network`, `Database`, `Monitoring`, or `Utility`).
  - [ ] a description that explains what the package does.
  - [ ] a version number prefixed with `v` that corresponds with a valid GitHub release.
  - [ ] links to SDKs in all supported languages.
  - [ ] a copyable command for installing the resource plugin if necessary.
  - [ ] an example of configuring the provider with `pulumi config set`.
  - [ ] an example of configuring the provider with environment variables.
- [ ] A CODEOWNER has reviewed the PR.
- [ ] A member of the @pulumi/docs team has reviewed all documentation.

### DIFF
--- a/themes/default/content/registry/packages/gcp/installation-configuration.md
+++ b/themes/default/content/registry/packages/gcp/installation-configuration.md
@@ -57,10 +57,10 @@ values:
           providerId: <your-provider-id>
           serviceAccount: <your-service-account>
   environmentVariables:
+    GOOGLE_OAUTH_ACCESS_TOKEN: ${gcp.login.accessToken}
     GOOGLE_PROJECT: ${gcp.login.project}
     GOOGLE_REGION: <your-region>
     GOOGLE_ZONE: <your-zone>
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
LGTM, thanks!

PR #4412, opened 2024-04-29
fix numbering in readme

susanev:
## description
was helping someone in community slack and noticed the numbering issue; also fixed some language while i was here

😎

### before
Screen Shot 2024-04-28 at 5 12 41 PM

### after
Screen Shot 2024-04-28 at 5 20 34 PM


lukehoban (approved):
Thanks @susanev! 🙌

PR #4663, opened 2024-05-31
Migrate to full use of tools/resourcedocsgen

guineveresaenger:
## Description

This pull request migrates functionality from https://github.com/pulumi/registrygen to `tools/resourcedocsgen` which is more actively maintained, in the aim of retiring `registrygen`.

### Before this change
(_to the best of my knowledge_)

- All providers were using `registrygen metadata` to generate metadata files, these are `Pulumi.yaml in the `themes/default/data` folder and `_index.md` and `configuration-installation.md` in the `themes/default/content` folder.
- Additionally, 3rd party providers were using a version check functionality to verify whether packages should be updated (on a cron-based Workflow).
- All providers are using `resourcedocsgen` to build API docs. This pull request does not change that.

### After this change
- All providers use `resourcedocsgen metadata`
- 3rd party providers use `resourcedocsgen pkgversion`
- Both commands are updated to contain the functionality of registrygen at this time, with once exception, see below.

It has been a constant source of confusion to maintainers on which tool to use when, as they're very similar and both come with unused code that looks like it's the real thing (but we use the other tool for that). I have personally wasted entirely too much time on this discrepancy.

After this pull request merges, we should be able to retire pulumi/registrygen.

I recommend reviewing these commits one by one:

1. [Migrate registrygen's metadata command to tools/resourcedocsgen](https://github.com/pulumi/registry/commit/5a129b2c73d30eca5f43002307c1d3f4c84885f9)
 Moves the current, stable functionality we've been using to `resourcedocsgen`. There are functionality and use updates.
 2. [Ignore the build resourcedocsgen binary](https://github.com/pulumi/registry/commit/f99d59f6f48414ecc4242292e78a2be7695a7f42) - you can ignore the gitignore
 3. [Special-case Pulumi for top-level config files not found](https://github.com/pulumi/registry/commit/cf3c30e9338e6d350ee1f41f26250b964e430540) This commit sends an error for 3rd party providers if the required docs files cannot be found. This will mostly help identify issues automatically upon first-time publication of a 3rd party provider to the pulumi registry. We continue to return `nil` for Pulumi providers, so that we continue to use the hard coded config docs files in the registry until we've implemented https://github.com/pulumi/registry/issues/4749 (work underway).
 4. [Update Go module name and update references](https://github.com/pulumi/registry/commit/ec95190db29bf1307c95d7ccd6c788051b9ea631)
 Follows Go module naming protocol. This is a cleanup from the migration from pulumi/docs.
 5. [Use resourcedocsgen for generating package metadata](https://github.com/pulumi/registry/commit/99a0b5919701b5987ddc6a15efab18fc8af7f09c) - Use new metadata command in CI
 6. [Transport pkgversion check to resourcedocsgen](https://github.com/pulumi/registry/commit/f76e20f49c53d027580a59eeed8d62f75bd9a1c5) - for 3rd party providers, import the pkgversion check as well
 7. [Replace pkgversion check](https://github.com/pulumi/registry/commit/f33fde09e649b7c1134381ab2ac6faea8f193b4c) - Use new pkgversion command in CI.


tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -420,6 +369,15 @@ func isComponent(keywords []string) bool {
...
 }
 
+func isFeaturedPackage(str string) bool {
+	for _, v := range featuredPackages {
+		if v == str {
+			return true
+		}
+	}
+	return false
+}
```
iwahbe, line 379: Nit: Can we just use https://pkg.go.dev/golang.org/x/exp/slices#Contains here?

```suggestion
func isFeaturedPackage(package string) bool {
	return slices.Contains(featuredPackages, package)
}
```

tools/resourcedocsgen/pkg/lookup.go
```diff
@@ -0,0 +1,196 @@
...
+	"sumologic":                           PackageCategoryMonitoring,
+	"tls":                                 PackageCategoryUtility,
+	"vault":                               PackageCategoryInfrastructure,
+	"venafi":                              PackageCategoryInfrastructure,
+	"vsphere":                             PackageCategoryCloud,
+	"wavefront":                           PackageCategoryMonitoring,
+	"yandex":                              PackageCategoryCloud,
+}
+
+// TODO[pulumi/pulumi#7813]: Remove this lookup once display name is available in
```
iwahbe, line 101: [This issue](https://github.com/pulumi/pulumi/issues/7813) was closed in Aug, 2021.

Can we create an issue here to clean this up?

tools/resourcedocsgen/pkg/lookup.go
```diff
@@ -0,0 +1,196 @@
...
+	"cloud":          PackageCategoryCloud,
+	"database":       PackageCategoryDatabase,
+	"infrastructure": PackageCategoryInfrastructure,
+	"monitoring":     PackageCategoryMonitoring,
+	"network":        PackageCategoryNetwork,
+	"utility":        PackageCategoryUtility,
+	"vcs":            PackageCategoryVCS,
+}
+
+var CategoryLookup = map[string]PackageCategory{
```
iwahbe, line 13: It looks like this is what our [`keywords`](https://github.com/pulumi/pulumi/blob/578e0937a96c594cc294cdf54deaa02c0328a3e5/pkg/codegen/schema/schema.go#L1858-L1865) field does. Can we just use that?

Can we create an issue to track the follow up if it doesn't already exist?

iwahbe (commented):
I love this!

tools/resourcedocsgen/pkg/lookup.go
guineveresaenger: I'll make an issue - I'd like to keep this PR scoped to the import of `registrygen`'s functionality as much as possible.

tools/resourcedocsgen/pkg/lookup.go
guineveresaenger: https://github.com/pulumi/registry/issues/4672

tools/resourcedocsgen/cmd/metadata.go
guineveresaenger: Sure thing!

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -369,8 +370,8 @@ func isComponent(keywords []string) bool {
 	return getTagFromKeywords(keywords, "kind/component") != nil
 }
 
-func isFeaturedPackage(package string) bool {
-	return slices.Contains(featuredPackages, package)
+func isFeaturedPackage(pkgName string) bool {
+	return git stslices.Contains(featuredPackages, pkgName)
```
iwahbe, line 374: Looks like something has gone wrong here...

It seems like there is an extra `git st` in there.

guineveresaenger: People like me need pre-commit hooks. 🤦

sean1588 (commented):
This is looking great @guineveresaenger ! Thanks so much for doing this!! I tested out these commands locally and outputs match what registrygen was doing before, so looks like the migration is working as expected! Just one thing to address.

cnunciato (commented):
Aside from Sean's comment, this all looks great to me. Thanks so much @guineveresaenger! I've been confused by the existence of both of these also.

.github/workflows/publish-provider-update.yml
```diff
@@ -22,13 +22,10 @@ jobs:
...
         uses: actions/checkout@v2
-      - name: Install Registrygen CLI
-        uses: jaxxstorm/action-install-gh-release@v1.2.0
-        with:
-          repo: pulumi/registrygen
       - name: Generate Package Metadata
         run: |
-          registrygen metadata --providerName=${{ env.PROVIDER_SHORT_NAME }} --repoSlug pulumi/pulumi-${{ env.PROVIDER_SHORT_NAME }} --schemaFile="${{ env.PROVIDER_SCHEMA_PATH }}" --version="${{ env.PROVIDER_VERSION }}" --publisher Pulumi
+          cd tools/resourcedocsgen && go run github.com/pulumi/registry/tools/resourcedocsgen \
+          resourcedocsgen metadata --providerName=${{ env.PROVIDER_SHORT_NAME }} --repoSlug pulumi/pulumi-${{ env.PROVIDER_SHORT_NAME }} --schemaFile="${{ env.PROVIDER_SCHEMA_PATH }}" --version="${{ env.PROVIDER_VERSION }}" --publisher Pulumi
```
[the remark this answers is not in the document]
guineveresaenger: oh! thank you for this catch!

sean1588 (approved):
LGTM 🚀

Issue #4696, opened 2024-06-07
New versions of (community) packages are not published


ringods:
### What happened?

I released v1.14 of the Scaleway provider yesterday and the new version PR was picked up.

https://github.com/pulumi/registry/pull/4694/files

Notice that the PR lists 3 new files to be added to the registry repo, rather than an update to the old metadata JSON file.

I merged it yesterday evening (CET). This morning, I see that the old version v1.12.1 is still listed in the registry:

https://www.pulumi.com/registry/packages/scaleway/

### Example

New files created by `resourcedocsgen` end up in the following locations:

https://github.com/pulumi/registry/tree/master/tools/resourcedocsgen/themes/default/content/registry/packages/scaleway
https://github.com/pulumi/registry/blob/master/tools/resourcedocsgen/themes/default/data/registry/packages/scaleway.yaml

But I guess somewhere else in the process the old locations are still used:

https://github.com/pulumi/registry/tree/master/themes/default/content/registry/packages/scaleway
https://github.com/pulumi/registry/blob/master/themes/default/data/registry/packages/scaleway.yaml

### Output of `pulumi about`

Not applicable

### Additional context

Could this have anything to do with the consolidation to the single `resourcedocsgen` tool?

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #4697, opened 2024-06-07
Fix package publishing job

sean1588:
## Description

fixes: https://github.com/pulumi/registry/issues/4696

There is an issue with the resourcedocsgen tool creating all the files under the `tools/resourcedocsgen` directory when updating the packages. This is because the tool is running in the context of the `tools/resourcedocsgen` dir, when before the consolidated of the tools, `registrygen`  was run from the root dir. This is happening now because we are doing a go run from the `tools/resourcedocsgen` dir so it is depositing the files under there, when really they need to be under the root.

This PR does the following:
- updates this to instead do a go build from the root and run the binary from the root so the files land in the right spot
- moved the yaml meta data files that were generated to the correct directory under themes/default/data/registry/packages so these pacakges can get updated (on merge of this PR, those api docs pages will get updated to the version specified)
-  removed the docs files (scaleway.yaml, scm.yaml) that were added to the wrong dir.  (On the next run of the community packages job, it should repull these files from the upstream and move these files to the right spot)


guineveresaenger (approved):
LGTM! Thank you for this catch.

PR #5052, opened 2024-07-29
Send urgent alerts to docs-ops-critical channel

sean1588:
## Description

part of: https://github.com/pulumi/devrel-team/issues/893

As a way to help differentiate more critical alerts that need to be addressed with immediacy from those that are less urgent, this will send the production build and deploy pipeline failures as well as any failures to update community packages to the #docs-ops critical channel. Let me know what you think.

### DIFF
--- a/.github/workflows/generate-package-metadata.yml
+++ b/.github/workflows/generate-package-metadata.yml
@@ -93,7 +93,7 @@ jobs:
       - name: Slack Notification
         uses: docker://sholung/action-slack-notify:v2.3.0
         env:
-          SLACK_CHANNEL: docs-ops
+          SLACK_CHANNEL: docs-ops-critical
           SLACK_COLOR: "#F54242"
           SLACK_MESSAGE: "generate package metadata failure in pulumi/registry repo :meow_sad:"
           SLACK_USERNAME: registrybot
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -74,7 +74,7 @@ jobs:
       - name: Slack Notification
         uses: docker://sholung/action-slack-notify:v2.3.0
         env:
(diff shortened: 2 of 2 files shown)


cnunciato:
I'm not sure I'd feel good about this, to be honest, as it implies that the messages that go go into docs-ops can be ignored, which needn't be the case -- we just haven't (to date) treated these notifications drop-everything urgent. I'd rather see us (specifically the on-call person_ treat the notifications that go into docs-ops as requiring review (e.g., they set their Slack notifications to notify of all messages) than create a new channel for "no, seriously" notifications.

If there's a higher level of notification needed than docs-ops, we should consider paging. Personally I don't think there is, yet, at least not with the events we're currently monitoring, but if there were, a Slack channel probably wouldn't be my choice.

cnunciato (changes requested):
I'm not sure I'd feel good about this, to be honest, as it implies that the messages that go into docs-ops can be ignored, which needn't be the case -- we just haven't (to date) treated those notifications drop-everything urgent. I'd rather see us (specifically the on-call person) treat the notifications that go into docs-ops as requiring immediate (or within-the-hour, or similar) review (e.g., they set their Slack notifications to notify of all messages) than create a whole new channel for "no, seriously" notifications.

If there's a higher level of notification needed than docs-ops, we should consider paging. Personally I don't think we're there yet, at least not with the events we're currently monitoring (i.e., build failures and the like), but if there were, a Slack channel probably wouldn't be my choice. IMO, a process tweak is all that should be necessary here.

iwahbe:
I strongly agree with @cnunciato's opposition to a doc-ops and doc-ops-critical channel. I'd much rather focus on making sure we only get notifications that we actually care about.

Issue #5129, opened 2024-08-08
Add the Tutorials tab


cnunciato:
We recently added a new Tutorials tab to the L2 navigation at pulumi.com/docs, so we need to do the same here as well.

PR #5130, opened 2024-08-08
Add the Tutorials tab

cnunciato:
Fixes https://github.com/pulumi/registry/issues/5129.


themes/default/layouts/partials/docs/search.html
```diff
@@ -1,25 +1,17 @@
 <div class="docs-type-nav-search">
     <div class="docs-type-nav">
-        {{ $mode := .Scratch.Get "mode" }}
```
cnunciato, line 3: I think it's fine to remove all this check because there's no longer a "docs" (or anything other than "registry") section in this repo.

sean1588: yeah should be fine, seems to be left over from when everything was together.

sean1588 (approved):
looks like the previews won't show the svg since it is over in the docs repo, but the ref does resolve on the site, pulumi.com/icons/graduation-cap.svg. 

Screen Shot 2024-08-08 at 4 30 59 PM

PR #5268, opened 2024-08-26
Hide installation config from nav when non existent

sean1588:
This PR enables a registry package to only contain only one file (`_index.md`) by hiding the Install & Config entry from the left nav when the `installation-configuration.md` file is not present.


guineveresaenger (approved):
This works for me locally - when there's only an `_index.md` file, it displays that file, and no nav for `installation-configuration`.

When both files exist, the nav and render are as before.

PR #5326, opened 2024-09-04
Add a script to apply temporary, targeted API docs fixes

### LINKED ISSUE #5095: Critical failure to render docs
### What happened?

A registry page has failed to render close to correctly.

My believe is that there is no level of broken provider docs that should result in this page.

### Example

Follow up on https://github.com/pulumi/pulumi-snowflake/issues/683.

### Output of `pulumi about`

NA

### Additional context

_No response_

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

### LINKED ISSUE #5136: Rancher2 Node Template docs are broken
### What happened?

Go to: https://www.pulumi.com/registry/packages/rancher2/api-docs/nodetemplate/

Looks like this: 

Image

### Example

n/a

### Output of `pulumi about`

n/a

### Additional context

_No response_

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

cnunciato:
Adds a script that runs during the build process to allow us to fix up random (major) API docs-generation issues while we await more permanent fixes. Also fixes one such issue.

Fixes #5095.
Fixes #5136.

### DIFF
--- a/scripts/apply-fixes.js
+++ b/scripts/apply-fixes.js
@@ -1,15 +1,43 @@
 const glob = require("glob");
 const fs = require("fs");
 
+/**
+ * This purpose of this script is to allow us to apply targeted, temporary fixes
+ * to API docs pages that need it, usually in response to a major rendering
+ * issue involving upstream provider docs.
+ *
+ * Any fix listed here should be accompanied by a comment explaining the issue
+ * as well as the fix, along with links to any related issues. Once those issues
+ * are resolved and the problem is no longer reproducible in production, the fix
+ * should be removed.
+ */
+
 glob.sync("themes/default/content/registry/packages/**/_index.md").forEach(path => {
 
-    // https://github.com/pulumi/registry/issues/5095
-    // Finds and replaces an errant set of backticks in the TF docs.
-    // We should be able to remove this once https://github.com/rancher/terraform-provider-rancher2/pull/1394 merges and is released.
+    /**
+     * Finds and replaces an errant set of backticks in the TF docs, breaking the layout
+     * of the page and rendering it unusable.
+     * https://github.com/pulumi/registry/issues/5095
+     *
+     * The script looks specifically for the following two-line pattern in the `nodetemplate`
+     * page of the `rancher2` docs:
+     *
+     * implements `fixNodeTemplateID()` that will update tfstate with proper id.
+     * ```
+     *
+     * The unmatched code fence above is what's breaking the page. The script just finds and
+     * removes this line.
+     *
+     * The following PR on the upstream provider fixes the issue:
+     * https://github.com/rancher/terraform-provider-rancher2/pull/1394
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -1191,6 +1191,14 @@ cssstyle@^3.0.0:
   dependencies:
     rrweb-cssom "^0.6.0"
 
+cypress-multi-reporters@^1.6.4:
```
cnunciato, line 1194: Getting this when I run `yarn` locally, so I'm assuming that means it should be checked in.

scripts/apply-fixes.js
```diff
@@ -0,0 +1,25 @@
...
+
+    // https://github.com/pulumi/registry/issues/5095
+    // Finds and replaces an errant set of backticks in the TF docs.
+    if (path === "themes/default/content/registry/packages/rancher2/api-docs/nodetemplate/_index.md") {
+        const content = fs.readFileSync(path, "utf8");
+        const lines = content.split("\n");
+        let precedingLine, nextLine = undefined;
+
+        lines.forEach((line, i) => {
+            if (line.endsWith("implements `fixNodeTemplateID()` that will update tfstate with proper id.")) {
```
sean1588, line 14: hmm honestly don't really like this. My concerns are around if the upstream changes we could incorrectly delete the following line that may end up causing just as much breakage especially if the line that follows happens to include html code or something important.  I don't really think this line approach is a good idea for that reason. Can we do something like check to make sure we have complete pairs of backticks and only remove last set of backticks if it is unclosed similar to what I am doing in the other PR I opened, except here it will only be isolated to this one file.

sean1588: ahh I see you checked the next lines below

scripts/apply-fixes.js
```diff
@@ -0,0 +1,25 @@
...
+                precedingLine = i;
+                nextLine = i + 1;
+            }
+        });
+
+        if (precedingLine && nextLine && lines[nextLine] === "") {
+            delete lines[nextLine];
```
sean1588, line 21:
```suggestion
if (lines[nextLine].includes("```") {
    delete lines[nextLine];
}
```

maybe something like this would suffice for now if i'm following the logic correctly here. At least this would stop us from accidentally deleting the wrong thing when/if the upstream changes.

scripts/apply-fixes.js
cnunciato: I figured it'd be extremely unlikely to encounter a situation in which these two lines would appear next to each other and be legit, but I went ahead and added a quick check to make sure there's also an add number of fences, just in case.

sean1588 (approved):

PR #5444, opened 2024-09-17
Document migrating aws.s3.Bucket to aws.s3.BucketV2

t0yv0:
## Description

Add a migration guide for moving from aws.s3.Bucket to aws.s3.BucketV2.

Fixes https://github.com/pulumi/home/issues/3630

## Adding a new package?

If this pull request adds a new package:

- [ ] The package's schema URL in this PR is correct.
- [ ] The package metadata file, if present, contains:
  - [ ] a supported category (one of `Cloud`, `Infrastructure`, `Network`, `Database`, `Monitoring`, or `Utility`).
  - [ ] a valid plugin download URL. See [Publish Your Package](https://www.pulumi.com/docs/using-pulumi/pulumi-packages/how-to-author/#publish-your-package).
  - [ ] a description that explains what the package does.
  - [ ] links to SDKs in all supported languages.
  - [ ] a copyable command for installing the resource plugin if necessary.
  - [ ] an example of configuring the provider with `pulumi config set`.
  - [ ] an example of configuring the provider with environment variables.
- [ ] The repository has:
  - [ ] a version tag prefixed with `v` that corresponds with a valid GitHub release and published package SDKs
- [ ] A CODEOWNER has reviewed the PR.
- [ ] A member of the @pulumi/docs team has reviewed all documentation.

### DIFF
--- a/themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
+++ b/themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
@@ -5,10 +5,9 @@ meta_desc: Practitioner level instructions for migrating from aws.s3.Bucket to a
 layout: package
 ---
 
-In the upcoming AWS Classic major release (v7), `aws.s3.Bucket` will be discontinued in favor of `BucketV2`. Users of
-`aws.s3.Bucket` resource are encouraged to migrate early. The migration requires a significant refactor to the source
-code and additional steps for avoiding data loss. This guide aims to cover all relevant scenarios with precise migration
-instructions.
+In the upcoming AWS Classic major release (v7), `aws.s3.Bucket` will be discontinued in favor of `BucketV2`. The
+migration requires a significant refactor to the source code and additional steps for avoiding data loss. This guide is
+for users opting to migrate early.
 
 To migrate existing `aws.s3.Bucket` resources to `aws.s3.BucketV2`:
 
@@ -16,16 +15,15 @@ To migrate existing `aws.s3.Bucket` resources to `aws.s3.BucketV2`:
    `BucketV2` inputs such as `policy` and `accelerationStatus` are to be replaced by side-by-side resources
    `aws.s3.BucketPolicy` and `aws.s3.BucketAccelerateConfigurationV2`.
 
-2. To avoid replacing the buckets and risking data loss, we recommend manually migrating the Bucket resources via
-   `pulumi import`
+2. To avoid replacing the buckets and risking data loss, migrate via `pulumi import`
 
-3. If replacing buckets is not a concern for your use case, consider a simpler migration procedure with `pulumi up`
+3. If replacing buckets is not a concern for your use case, use a simpler migration procedure via `pulumi up`
 
 ## Migrating with `pulumi import`
 
-This migration path is recommended in situations when replacing the bucket in the AWS account is not acceptable. After
-performing the steps your Pulumi program and state will be updated to track buckets using the new resource without
-executing any changes against the actual cloud account. The steps involve:
+This migration path is best for when replacing the bucket in the AWS account is not acceptable. After performing the
+steps your Pulumi program and state will be updated to track buckets using the new resource without executing any
+changes against the actual cloud account. The steps involve:
 
 - Find URNs for legacy Bucket Pulumi resources using `pulumi stack export`
 - Determine the actual bucket name(s)
@@ -3700,7 +3698,8 @@ for more information:
 The distinction between Bucket and BucketV2 originates from breaking changes introduced in the V4 release of the
 Terraform AWS Provider. In the Pulumi AWS provider projection BucketV2 represents the latest version of the upstream
 resource, while Bucket is maintained by Pulumi to enable backwards compatibility. We recommend using BucketV2 for new
-code and migrating old code to BucketV2 as it will remain the preferred long-term supported resource.
+code. Migrating existing old code to BucketV2 is not required in the V6 version of the provider, but has the benefits of
+aligning the code with the preferred long-term supported resource.
 
 See also:
 


t0yv0:
@interurban opening for early review, thanks a lot!

As an aside, I'm working to get examples formatted with the right `<chooser>` magic and I'm putting together a little script that updates the non-yaml sections automatically via `pulumi convert`. Would you all be interested in something like this or is there existing machinery to do this? Much appreciated.

t0yv0:
Looks like something with choosers inside lists is not happy:

/home/runner/work/registry/registry/themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md:
Line 2417: Ordered list item prefix [Expected: 1; Actual: 6; Style: 1/2/3].
Line 2420: Ordered list item prefix [Expected: 2; Actual: 7; Style: 1/2/3].
Line 2422: Ordered list item prefix [Expected: 3; Actual: 8; Style: 1/2/3].

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -0,0 +1,2423 @@
+---
+title: "BucketV2 Migration Guide"
+h1: "Migrating from aws.s3.Bucket to aws.s3.BucketV2"
+meta_desc: Practitioner level instructions for migrating from aws.s3.Bucket to aws.s3.BucketV2 resources
+layout: package
+---
+
+In the upcoming AWS Classic major release (v7), `aws.s3.Bucket` will be discontinued in favor of `BucketV2`. Users of
+`aws.s3.Bucket` resource are encouraged to migrate early. The migration is straightforward for simple use cases but
+requires some care for advanced configurations. Specifically `BucketV2` inputs such as `policy` and `accelerationStatus`
```
flostadler, line 10: nit: I'm not sure it's fair to call them advanced use cases. Things like enforcing secure transport or versioning are best practices for S3 (see https://docs.aws.amazon.com/AmazonS3/latest/userguide/security-best-practices.html).

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -0,0 +1,2423 @@
...
+are deprecated and using the requires migrating to side-by-side resources (see Removed inputs). This guide aims to cover
```
flostadler, line 11:
```suggestion
are deprecated and using them requires migrating to side-by-side resources (see Removed inputs). This guide aims to cover
```

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -0,0 +1,2423 @@
...
+all relevant scenarios with precise migration instructions.
+
+To migrate existing `aws.s3.Bucket` resources to `aws.s3.BucketV2`:
+
+1. Edit your Pulumi program sources to replace removed inputs (see below) with equivalent side-by-side resources.
+
+2. Perform `pulumi up` to replace (delete and re-create) the buckets in AWS.
```
flostadler, line 18: I think we should add some more warnings here to make it clear that this will delete all objects in the bucket and that they're not recoverable.

We might also add steps to make users check whether the buckets are empty or intentionally commit to cleaning the data.

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -0,0 +1,2423 @@
...
+
+If replacement is not acceptable, it is possible to perform a manual migration with `pulumi import` (see Avoiding
+replacement).
```
flostadler, line 21: We should be able to link to the section like this: `[Avoiding Replacement](#avoiding-replacement)`. Not 100% sure though whether Hugo supports that syntax

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -0,0 +1,2423 @@
...
+
+{{% /choosable %}}
+
+{{< /chooser >}}
+
+
+
+
+
+As a bonus, the policy can now more easily refer to the concrete name of the bucket.
```
flostadler, line 532: It might be worth explaining the benefit of not having cyclic deps in a bit more detail in the beginning. Gives some more insights into why this is a good thing for users.

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -0,0 +1,2423 @@
...
+
+Note that `dependsOn` on `BucketOwnershipControls` is required for Pulumi to correctly order the operations for this
+infrastructure. Consult the documentation on `aws.s3.BucketAclV2` for more fully worked examples of configuring ACL.
+
+### hostedZoneId input
+
+This will only be available as an output. If your program specified this as an input, simply remove it. The input was
+ignored by prior versions of the provider and was exposed in error.
+
+### other inputs
```
flostadler, line 1978: How were the props with explicit instructions chosen? I guess they're the most used ones?

I'm wondering if we should have a single example with all options (might not be possible if some settings are mutually exclusive) and show how to migrate that instead.

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -0,0 +1,2423 @@
...
+
+
+## Avoiding replacement
+
+In situations when replacing the bucket in the AWS account is not acceptable, it is possible to perform a manual
+migration that changes Pulumi program and state to track buckets using the new resource without executing any changes
+against the actual cloud account. While the details will vary depending on your use case, this procedure generally
+involves the following steps:
+
+- Find URNs for legacy Bucket Pulumi resources using `pulumi stack export`
```
flostadler, line 2003: We should be able to extract those with jq, right? Could be a nice addition to add that here

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
t0yv0: It's been failing builds. Looks like this only works in GitHub markdown. @interurban might have some pointers here.

t0yv0:
@flostadler I went ahead and took a larger bucket example for a spin through the import-based migration procedure. It worked but highlighted a few extra steps, this was useful.

For alternative upgrade procedures.

I have filed https://github.com/pulumi/pulumi-aws/issues/4471 as this seems to have regressed.

I've also had a slack conversation with a user trying this today and filed https://github.com/pulumi/pulumi-aws/issues/4470 - sounds like we could do a bit more possibly, depending on interest here.

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
t0yv0: It's a difficult sell. I tried adding a Context section at the bottom, but not sure it belongs, perhaps this sentence does not belong either. Primary motivation for users to do this is to get onto a supported version because they have to.

I've checked the style of https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-4-upgrade#s3-bucket-refactor FWIW it's very technical to-the-point and does not explain any design considerations as to why.

georgberky:
> I've also had a slack conversation with a user trying this today and filed [pulumi/pulumi-aws#4470](https://github.com/pulumi/pulumi-aws/issues/4470) - sounds like we could do a bit more possibly, depending on interest here.

The confusing things for me were to find the old migration guide and
- the full explanation being in a video instead of in text
- the guide not being too clear that you have to declare a v2 resource _next to_ the v1 resource instead of replacing v1 with v2 even though they have the same Pulumi resource name
- the whole mechanism not working anymore
- an IaC tool trying to delete my resource (luckily failing because the old bucket wasn't empty, but this doesn't raise confidence)

t0yv0:
Thanks for your feedback @georgberky that's a good point; we don't edit old blog content but we have a way to marking it obsolete and pointing to new material, I'll make sure we do this to https://www.pulumi.com/blog/announcing-v5.0.0-of-the-pulumi-aws-provider/#can-i-migrate-from-awss3bucket-to-awss3bucketv2

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -0,0 +1,3701 @@
...
+code and additional steps for avoiding data loss. This guide aims to cover all relevant scenarios with precise migration
+instructions.
+
+To migrate existing `aws.s3.Bucket` resources to `aws.s3.BucketV2`:
+
+1. Edit your Pulumi program sources to replace removed inputs with equivalent side-by-side resources. Specifically
+   `BucketV2` inputs such as `policy` and `accelerationStatus` are to be replaced by side-by-side resources
+   `aws.s3.BucketPolicy` and `aws.s3.BucketAccelerateConfigurationV2`.
+
+2. Perform `pulumi up` to replace the buckets in AWS. **WARNING**: replacing the buckets will delete and re-create them,
```
corymhall, line 19: I'm wondering if we should flip this and mention replacing buckets as the alternative. I think 99% of buckets won't be able to be replaced so we should just go with that as the default.

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -0,0 +1,3701 @@
...
+
+{{< chooser language "typescript,python,go,csharp,java,yaml" >}}
+
+{{% choosable language typescript %}}
+
+```typescript
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+
+const myBucket = new aws.s3.Bucket("my-bucket", {serverSideEncryptionConfiguration: {
```
corymhall, line 540:
```suggestion
const myBucket = new aws.s3.Bucket("my-bucket", {
  serverSideEncryptionConfiguration: {
```

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -0,0 +1,3701 @@
...
+| websiteDomain            | aws.s3.BucketWebsiteConfigurationV2        |
+| websiteEndpoint          | aws.s3.BucketWebsiteConfigurationV2        |
+
+
+## Avoiding replacement
+
+In situations when replacing the bucket in the AWS account is not acceptable, it is possible to perform a manual
+migration that changes Pulumi program and state to track buckets using the new resource without executing any changes
+against the actual cloud account. While the details will vary depending on your use case, this procedure generally
+involves the following steps:
```
corymhall, line 2093: I think as far as manual steps go this is not bad and as good as we can do 👍🏻

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
t0yv0: You're right.. that seems to be really true. I'll incorporate.

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
t0yv0: I have auto-generated all examples in non-YAML languages from YAML using `pulumi convert`. Prefer leaving like this for now as it's easier to edit.

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -16,17 +16,28 @@ To migrate existing `aws.s3.Bucket` resources to `aws.s3.BucketV2`:
    `BucketV2` inputs such as `policy` and `accelerationStatus` are to be replaced by side-by-side resources
    `aws.s3.BucketPolicy` and `aws.s3.BucketAccelerateConfigurationV2`.
 
-2. Perform `pulumi up` to replace the buckets in AWS. **WARNING**: replacing the buckets will delete and re-create them,
-   losing any data stored in these buckets. If this is not acceptable, consider using `pulumi import` to migrate
+2. Perorm this is not acceptable, consider using `pulumi import` to migrate
```
corymhall, line 19: Maybe something like

"In order to avoid resource replacement, consider using `pulumi import` to manually migrate resources"

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -695,9 +1625,16 @@ resources:
 
 
 
+## Migrating with `pulumi up`
 
+If you have reliable backups or test account scenarios where deleting and re-creating buckets is acceptable, a simpler
```
corymhall, line 1630: 👍🏻

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
t0yv0: Hmm, I think I have spellechecker in Emacs but I missed this, let me double-check.

cnunciato:
@t0yv0 @mjeffryes Correct me if I'm wrong, but wouldn't it possible to write a tool to take care of this for users automatically? This seems like an awful lot of manual (and potentially risky) work to distribute to all of our users (times all of their buckets) for something that looks, on the surface, to be readily amenable to automation. Have we considered that?

mjeffryes:
It might be possible, to automate this, but TBH, we're not 100% convinced that upstream is going to follow through on pushing everyone to the side-car resources, so it doesn't make sense to invest in an automation for a migration that may not be needed.

The purpose of these docs is just to have a reference of what's currently required to migrate. We're not telling users they should migrate at this time.

themes/default/content/registry/packages/aws/how-to-guides/bucketv2-migration.md
```diff
@@ -0,0 +1,3708 @@
+---
+title: "BucketV2 Migration Guide"
+h1: "Migrating from aws.s3.Bucket to aws.s3.BucketV2"
+meta_desc: Practitioner level instructions for migrating from aws.s3.Bucket to aws.s3.BucketV2 resources
+layout: package
+---
+
+In the upcoming AWS Classic major release (v7), `aws.s3.Bucket` will be discontinued in favor of `BucketV2`. Users of
+`aws.s3.Bucket` resource are encouraged to migrate early. The migration requires a significant refactor to the source
```
mjeffryes, line 9: Maybe saying "Users of `aws.s3.Bucket` resource are encouraged to migrate early." is too aggressive. We don't actually want to push existing users to move yet I think.

interurban (approved):

Issue #4902, opened 2024-07-11
Strata Cloud Manager Provider - Make the Provider Searchable by More Terms


phillipedwards:
Currently, [Strata Cloud Manager](https://www.pulumi.com/registry/packages/scm/) provider in the pulumi registry is searchable by one or more of: strata, cloud, manager.

As the product of Strata Cloud Manager is owned by Palo Alto, is a Prisma product, and is often referred to by the term "scm" I'd like the provider to be searchable by all of those terms.

Concretely, if someone types one of the following in the registry "search" bar, I'd like this provider to be found:
- strata | cloud | manager
- palo | alto
- prisma
- scm

cnunciato:
Also document how to add these keywords.

thoward:
@interurban can you follow up with @mjeffryes and/or @iwahbe on this?

iwahbe:
I havn't dug into search at all. It sounds like we need a general keywork/tag mechanism, and that [what we have](https://github.com/pulumi/pulumi-scm/blob/0254ffa3e5e71f01cd9f7ba7686061fd9dbab129/provider/resources.go#L66-L71) isn't doing it.

PR #5506, opened 2024-09-25
update title to make SCM package more findable

sean1588:
fixes: https://github.com/pulumi/registry/issues/4902


interurban (approved):

iwahbe:
This is not a durable change, and will be blown away the next time that scm publishes a new update. I think the durable fix will be to write an update to `.DisplayName` in the scm package.

iwahbe:
I'm going to close this out, since this is the wrong place to make this kind of change. If we want to make this change, it needs to be in the provider repo.

PR #5544, opened 2024-09-27
Check dir exists for cloud before copying tutorials

sean1588:
This is causing the tutorials job to fail since the aws-native tutorials have been removed. I left it in the listing, just in case we decide to bring them back or if this is something temporary. I instead added a check to verify that the directory exists (meaning there are tutorials for that cloud provider, if not it would not create the dir) and to skip if it doesn't.

### DIFF
--- a/scripts/ci/mktutorial.sh
+++ b/scripts/ci/mktutorial.sh
@@ -19,7 +19,8 @@ for cloud in "aws-apigateway" "aws" "aws-native" "classic-azure" "azure" "gcp" "
     mkdir -p ./themes/default/content/registry/packages/$dest
     mkdir -p ./themes/default/content/registry/packages/$dest/how-to-guides
 
-    # Check if the directory exists before copying
+    # Check if the directory exists before copying.
+    # (meaning there are tutorials for that cloud provider, if not the dir would not get created)
     if [ -d "$TUTORIAL_OUT/tutorials/$cloud" ]; then
         cp $TUTORIAL_OUT/tutorials/$cloud/* ./themes/default/content/registry/packages/$dest/how-to-guides/
     fi


scripts/ci/mktutorial.sh
```diff
@@ -18,6 +18,11 @@ for cloud in "aws-apigateway" "aws" "aws-native" "classic-azure" "azure" "gcp" "
 
     mkdir -p ./themes/default/content/registry/packages/$dest
     mkdir -p ./themes/default/content/registry/packages/$dest/how-to-guides
-    cp $TUTORIAL_OUT/tutorials/$cloud/* ./themes/default/content/registry/packages/$dest/how-to-guides/
+
+    # Check if the directory exists before copying.
+    # (meaning there are tutorials for that cloud provider, if not the dir would not get created)
+    if [ -d "$TUTORIAL_OUT/tutorials/$cloud" ]; then
+        cp $TUTORIAL_OUT/tutorials/$cloud/* ./themes/default/content/registry/packages/$dest/how-to-guides/
```
thoward, line 25: LGTM. My only thought is 'will something downstream break if the directory isn't there after this runs?'.

e.g. will there be a 404 that trips up the 404 checker, or some such. but likely the only way we'll know that is by shipping this and seeing what happens.

thoward (approved):

scripts/ci/mktutorial.sh
sean1588: that is a good question you. I don't believe the 404 checker itself checks for registry pages. From looking at the code here though, I don't believe the pages will actually be removed, since I only see a command to copy files to the location in our repo. So it looks like this was all configured to either add or update existing tutorial files. It doesn't look like it will remove any that are no longer existing. I'm guessing we will want to remove, since there was probably a reason they removed them from the examples repo in the first place. Then we'll need to figure out if  redirect them or just go and fix internal broken links that may crop up when the pages are removed.

PR #5603, opened 2024-10-03
Move sitemap to correct location

sean1588:
fixes: https://github.com/pulumi/registry/issues/5604
Moves the registry sitemap to the correct location as defined per the site-index file.


komalali (approved):

sean1588:
hmm something is off here, http://registry--origin-pr-5603-514baf68.s3-website.us-west-2.amazonaws.com/registry/sitemap.xml does not resolve in the preview link, but does resolve when I run it locally. Let me do some more investigation.

cnunciato (approved):

PR #5653, opened 2024-10-08
Add cdnHost to fix refresh loop

cnunciato:
Similar to https://github.com/pulumi/docs/pull/13072, this fix adds a setting to the Segment cookie-consent configuration to fix a refresh loop.


cnunciato:
Closing in favor of #5647.

PR #5692, opened 2024-10-14
Relax lint rules for index files

guineveresaenger:
## Description

This pull request amends the markdown lint script to allow for files matching `/packages\/[a-z0-9]*\/_index.md/` to have slightly more relaxed linting rules compared to other markdown files in the bridge.

The following rules are set to `false` for files such as `registry/themes/default/content/registry/packages/auth0/_index.md`:
1.  Multiple top-level headings in the same document [MD025](https://github.com/DavidAnson/markdownlint/blob/main/doc/md025.md)
2. Ordered list item prefix [MD029](https://github.com/DavidAnson/markdownlint/blob/main/doc/md029.md)
3. Spaces inside code span elements [MD038](https://github.com/DavidAnson/markdownlint/blob/main/doc/md038.md) 

These files, for many bridged providers, are now generated from upstream and we do not have the same control over their content as we do for other files.

Part of https://github.com/pulumi/ci-mgmt/issues/1100.


scripts/lint/lint-markdown.js
```diff
@@ -13,6 +13,14 @@ const path = require("path");
 const FRONT_MATTER_REGEX = /((^---\s*$[^]*?^---\s*$)|(^\+\+\+\s*$[^]*?^(\+\+\+|\.\.\.)\s*$))(\r\n|\r|\n|$)/m;
 const AUTO_GENERATED_HEADING_REGEX = /###### Auto generated by ([a-z0-9]\w+)[/]([a-z0-9]\w+) on ([0-9]+)-([a-zA-z]\w+)-([0-9]\w+)/g;
 
+/**
+ * REGEX for matching top-level provider index files only. Example:
+ *
+ * registry/themes/default/content/registry/packages/auth0/_index.md
+ *
+ */
+const PACKAGE_INDEX_REGEX = /packages\/[a-z0-9]*\/_index.md/
```
sean1588, line 22:
```suggestion
const PACKAGE_INDEX_REGEX = /packages\/[a-z0-9\-]*\/_index.md/
```

slight bug with the regex, since it is failing to match package names that have a hyphen in them, like `aws-native` and `azure-native`. Other than that though it is all working well!

sean1588 (approved):
thanks so much @guineveresaenger! This all works well and looks good to me. Only that one minor regex adjustment.

scripts/lint/lint-markdown.js
guineveresaenger: good call, tysm!

PR #5703, opened 2024-10-15
Adds Copilot to /registry

foot:
## Description

Supports the use case:

- User opens https://www.pulumi.com/registry/ and sees copilot

Notes:

- Follows a similar pattern in /docs
- The entire `/registry` path is forwarded to this stack / repo so it has to have its own set of configuration for copilot etc
- PRs that intro'd Copilot in docs:
  - https://github.com/pulumi/docs/pull/12905
  - https://github.com/pulumi/docs/pull/13101

### DIFF
--- a/.github/workflows/testing-deploy.yml
+++ b/.github/workflows/testing-deploy.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - master
-      - simon/add-copilot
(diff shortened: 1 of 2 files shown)


sean1588 (approved):
Thank you @foot !! This all looks great to me 🚀

PR #5869, opened 2024-11-01
Do not clone the registry on build

iwahbe:
As currently set up, running `resourcedocsgen docs registry` clones the registry (even though `resourcedocsgen` is always called from the registry), then optionally checks out a specific commit (how we run in CI), then builds docs there, then copies them over.

We don't have to do any of this. By assuming that `resourcedocsgen` is always run in the registry root at the version, targeting the version of the registry we are checked out as (which is always true in CI), we can save ourselves a full clone of the registry and make developing the registry that much simpler.

This change also makes possible to build and develop the registry in a fork or without internet, since building no longer requires a clone and checkout of a commit from `github.com/pulumi/registry`.

### DIFF
--- a/scripts/ci/build.sh
+++ b/scripts/ci/build.sh
@@ -37,7 +37,7 @@ popd
 
 REGISTRY_COMMIT="$(git_sha_short)"
 printf "Generating API docs from registry commit %s...\n\n" "${REGISTRY_COMMIT}"
-resourcedocsgen docs registry --commitSha "${REGISTRY_COMMIT}" \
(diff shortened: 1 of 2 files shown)


guineveresaenger (approved):
This looks reasonable to me.
I was curious as to why this was written in this way in the first place.

https://github.com/pulumi/registry/pull/2963 moved this code from `pulumi/docs`

https://github.com/pulumi/docs/pull/6788#pullrequestreview-805161124 specifically calls out whether we need this.

I think this is fine to remove. Thank you!

Issue #5033, opened 2024-07-26
Docs markup broken


nafg:
[themes/default/content/registry/packages/random/api-docs/randompassword/_index.md](https://www.pulumi.com/registry/packages/random/api-docs/randompassword/)

![Image](https://github.com/user-attachments/assets/dfd3f74d-7a80-44d8-a714-3eca27648962)

PR #5877, opened 2024-11-03
Pre-render content

iwahbe:
This *might* categorically fix issues like #5033.


iwahbe:
As it turns out, it does not fix such issues.

PR #5879, opened 2024-11-04
Bump pulumi/pulumi dependency

iwahbe:
## Description

Bump the version of pulumi/pulumi used in `resourcedocsgen` and the version of Go used everywhere.

This PR is stacked on [iwahbe/do-not-clone-registry-on-build](https://github.com/pulumi/registry/tree/iwahbe/do-not-clone-registry-on-build).


tools/mktutorial/go.mod
```diff
@@ -1,11 +1,18 @@
 module github.com/pulumi/docs/tools/mktutorial
 
-go 1.13
```
mjeffryes, line 3: "What year is it?"

mjeffryes (approved):

PR #5898, opened 2024-11-05
Run Markdown lint in parallel

iwahbe:
We run a markdown lint pass on each PR build, which takes about 2m20s to run. This
comprises about 7% of the total CI time needed to validate a contributing PR. We can and
should run this in parallel with the "Build and deploy preview" job.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -8,6 +8,23 @@ jobs:
   resourcedocsgen:
     uses: ./.github/workflows/check-resourcedocsgen.yml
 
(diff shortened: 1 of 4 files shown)


iwahbe:
> [!NOTE]
> It looks like the markdown lint job got much slower. That's because it was moved from the expensive `pulumi-service-ubuntu-22.04-16core` runner to the free `ubuntu-latest` provided by GitHub. Because it runs in parallel, the overall CI run is faster.

t0yv0 (approved):

PR #5900, opened 2024-11-05
Lint `tools/mktutorial`

iwahbe:
Applies the same go linter to `mktutorial`.


.github/workflows/check-go.yml
iwahbe, line 1: `check-resourcedocsgen.yml` was renamed to `check-go.yml`, and then parameterized to apply to arbitrary components.

mjeffryes (approved):

PR #5908, opened 2024-11-06
Refactor cmd/metadata.go for clarity

iwahbe:
## Description

This PR is best reviewed commit by commit:

- 243e0a7fcfac267771ee4146c4ee2a276a8221c7 moves `repoSlug` parsing into `{owner}/{name}` into it's own type.
- 1bd3c2365e6c345d326d298d78af1ef552237f41 makes `mainSpec` a file local variable.
- 0c8d46c15badd8ae9b9bc11c673ccfd14037b252 moves schema fetching into it's own set of functions.
- 4b50e4022949cd90fa90e479179edf1c710902b3 rewrites the boolean logic for determining what is a component vs a native provider vs neither.
- fbaae16ed927c8d9ba7645bed74655d45ac78d86 simplifies the publisher logic into a switch statement.

☝️ are pure refactors, and will not change behavior.

- 9e6554fc56b1cb28d21cd409b5a7f9549a740738 changes the GH repo slug to require input in the form `a/b` where both `a` and `b` are non-empty. Previously `/` would have been accepted.

### DIFF
--- a/tools/resourcedocsgen/cmd/metadata.go
+++ b/tools/resourcedocsgen/cmd/metadata.go
@@ -150,10 +150,9 @@ func PackageMetadataCmd() *cobra.Command {
 				publisherName = publisher
 			case mainSpec.Publisher != "":
 				publisherName = mainSpec.Publisher
-			case repoSlug.owner != "":
-				publisherName = cases.Title(language.Und, cases.NoLower).String(repoSlug.owner)
 			default:
-				return errors.New("unable to determine package publisher")
+				contract.Assertf(repoSlug.owner != "", "repoSlug.owner is non-empty by construction")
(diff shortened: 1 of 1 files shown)


t0yv0:
Would it be possible to put the codebase under test before touching it? I do not see any test files at all around this small codebase. It does not look like a lot is happening here so perhaps the testing itself does not need to be particularly complicated.

iwahbe:
> Would it be possible to put the codebase under test before touching it? I do not see any test files at all around this small codebase. It does not look like a lot is happening here so perhaps the testing itself does not need to be particularly complicated.

Yes. I have rebased on top of https://github.com/pulumi/registry/commit/f99cca5233da567ee6f5a2b6ac9ac68f3e38f049, which adds tests for this functionality.

t0yv0 (approved):

PR #5909, opened 2024-11-06
Add a basic test for resourcedocsgen

iwahbe:
## Description

Add some tests to the `resourcedocsgen metadata` command before modifying it.


tools/resourcedocsgen/cmd/metadata_test.go
```diff
@@ -0,0 +1,53 @@
...
+		"--packageDocsDir", pacakgeDocsDir,
+	})
+	require.NoError(t, cmd.Execute())
+	t.Run("random.yaml", func(t *testing.T) {
+		t.Parallel()
+		autogold.ExpectFile(t, readFile(t, filepath.Join(metadataDir, "random.yaml")))
+	})
+	t.Run("_index.md", func(t *testing.T) {
+		t.Parallel()
+		autogold.ExpectFile(t, readFile(t, filepath.Join(pacakgeDocsDir, "_index.md")))
```
t0yv0, line 45: There are some tricks that sometimes make the golden files have fewer escapes:

```
Use your own formatting (JSON, etc.): Make your got value of type autogold.Raw("foobar"), and it will be used as-is for .golden files (not allowed with inline tests.)
```

t0yv0 (approved):

PR #5915, opened 2024-11-07
Fix broken links in how-to-guides

mjeffryes:
## Description

We recently removed some azure classic examples from the examples repo, so several of these generated guides no longer point to anything. This change removes those guides. (I've chosen this over pinning an old version of the examples from the repo because we would not be keeping the content up to date any more with new provider releases.)

I also removed a stale link from one of the AWS tutorial. (Docker moved/removed the example it was based on, but the rest of the guide should be fine.)


sean1588 (approved):
Thanks @mjeffryes ! yeah this makes sense to me. Thanks for removing them!

PR #5917, opened 2024-11-07
Concurrent registry docs generation

iwahbe:
## Description

Speed up CI by running docs generation in parallel.

On my machine, this decreases the time it takes to build the docs from 6m30s to 1m30s, a 4 fold speedup.

In CI, we see:
| Concurrent go routines | CI time (minutes) |
|--|--|
| 8 | 25 |
| [`NumCPU()`](https://github.com/pulumi/registry/pull/5918) | 23 |
| [1](https://github.com/pulumi/registry/pull/5919) | 30 |

This PR depends on https://github.com/pulumi/pulumi/pull/17711.

---

I'm planning on writing some tests for `resourcedocsgen docs registry` in a separate PR. I'll merge those and then rebase this PR so the area under change is tested. I'd still appreciate a review on this PR now. I don't imagine that this code will change once tests are in (unless it doesn't work).

### DIFF
--- a/tools/resourcedocsgen/cmd/docs/registry.go
+++ b/tools/resourcedocsgen/cmd/docs/registry.go
@@ -16,7 +16,6 @@ package docs
 
 import (
 	"encoding/json"
-	stderrors "errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -25,7 +24,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
-	"sync"
 
 	"github.com/spf13/cobra"
 
@@ -38,6 +36,7 @@ import (
 	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/pulumi/registry/tools/resourcedocsgen/pkg"
+	concpool "github.com/sourcegraph/conc/pool"
 )
 
 func getRepoSlug(repoURL string) (string, error) {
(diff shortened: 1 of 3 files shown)


danielrbradley:
What's the memory usage like per process and therefore overall?

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -126,31 +129,71 @@ func genResourceDocsForAllRegistryPackages(registryRepoPath, baseDocsOutDir, bas
...
+
+			docsOutDir := filepath.Join(baseDocsOutDir, metadata.Name, "api-docs")
+			err = genResourceDocsForPackageFromRegistryMetadata(metadata, docsOutDir, basePackageTreeJSONOutDir)
+			if err != nil {
+				addError(errors.Wrapf(err, "generating resource docs using metadata file info %s", f.Name()))
+				return
+			}
 
-		glog.Infof("=== completed %s ===", f.Name())
+			glog.Infof("=== completed %s ===", f.Name())
```
blampe, line 188: (This is my first time seeing this custom codegen logic so take this with a grain of salt.)

AFAICT we always generate docs for each metadata file we visit? If that's true we should try to short-circuit if the metadata hasn't changed and outputs already exist, e.g. by tracking a hash or mtime somewhere.

If we turn `genResourceDocsForPackageFromRegistryMetadata` into a self-contained binary we could probably handle all of this with `make -j`.

iwahbe:
> What's the memory usage like per process and therefore overall?

I havn't profiled. We havn't had any memory pressure from docs generation in the past, and I didn't observe memory pressure on my machine. The only memory pressure this repo has generated has been from hugo builds, and this is a step before that.

tools/resourcedocsgen/cmd/docs/registry.go
blampe: > We need to since we don't store the output in repo.

Let's check it in then! No need to re-compute docs that haven't changed.

> If/when we implement caching, this will still generate wins when the cache key changes (say when we bump the pulumi/pulumi version).

For sure, the regenerate-everything case will always exist. I'm thinking more about the local dev loop where incremental builds are important.

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -126,31 +129,71 @@ func genResourceDocsForAllRegistryPackages(registryRepoPath, baseDocsOutDir, bas
...
+	// Because there are a fixed number of hallpasses (maxConcurrency), and each
+	// process needs a hallpass to run, we limit effective concurrency to the number
+	// of available hallpasses.
+	maxConcurrency := runtime.NumCPU()
+
+	var hallpass struct{}
+	monitor := make(chan struct{}, maxConcurrency)
+	for i := 0; i < maxConcurrency; i++ {
+		monitor <- hallpass
+	}
```
danielrbradley, line 155: These low-level concurrency approaches make me very nervous as, in my experience, it's very easy to unintentionally and subtly break things. Not sure on what the recommended libs are in Go world, but a quick search showed up [sourcegraph/conc](https://github.com/sourcegraph/conc) which would allow us to write this process as follows:

```go
func mapStream(
    in chan int,
    out chan int,
    f func(int) int,
) {
    s := stream.New().WithMaxGoroutines(10)
    for elem := range in {
        elem := elem
        s.Go(func() stream.Callback {
            res := f(elem)
            return func() { out <- res }
        })
    }
    s.Wait()
}
```

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -126,31 +129,71 @@ func genResourceDocsForAllRegistryPackages(registryRepoPath, baseDocsOutDir, bas
...
+			}(<-monitor)
+			glog.Infof("=== starting %s ===\n", f.Name())
+			glog.Infoln("Processing metadata file")
+			metadataFilePath := filepath.Join(registryPackagesPath, f.Name())
+
+			b, err := os.ReadFile(metadataFilePath)
+			if err != nil {
+				addError(errors.Wrapf(err, "reading the metadata file %s", metadataFilePath))
+				return
+			}
```
danielrbradley, line 173: Would be nice to not have the core process of file operations etc intermingled with the concurrency concerns. Perhaps some refactoring to pull out a single method that the concurrency-related code can call into would suffice.

danielrbradley (commented):
Overall approach of applying concurrency makes sense, but it would be nice to have a more robust concurrency abstraction that's harder to break.

Also, not sure what the debugging story is around this tool. Expect it would be nice to be able to run this single-threaded if needing to debug and step through any parts.

danielrbradley (approved):
Much happier with the concurrency code! Thanks!

PR #5920, opened 2024-11-07
Do not limit GO's GC during Hugo builds

iwahbe:
Since CI does a full build of the registry, I'm trying to validate if we can run a registry build without manually upping GC usage. It looks like we can.


mjeffryes (approved):
Looks like it worked pretty well. Not a big wall clock time impact, but should let us make better use of the resources of the runner.

iwahbe:
It looks like this change alone took build times down from 30m to 20m!

PR #5931, opened 2024-11-08
Initial tests for `resourcedocsgen docs registry`

iwahbe:
Add integration tests for the `resourcedocsgen docs registry some-package` command.


tools/resourcedocsgen/cmd/docs/registry.go
iwahbe, line 1: The changes to this file are necessary to test without messing with chdir (which interferes with autogold).

tools/resourcedocsgen/cmd/docs/registry_test.go
```diff
@@ -0,0 +1,58 @@
...
+	cmd := resourceDocsFromRegistryCmd()
+	cmd.SetArgs([]string{
+		"random", /* pkgName */
+		"--baseDocsOutDir", baseDocsOutDir,
+		"--basePackageTreeJSONOutDir", basePackageTreeJSONOutDir,
+		"--registryDir", registryDir,
+	})
+	require.NoError(t, cmd.Execute())
+	t.Run("docs", func(t *testing.T) { util.AssertDirEqual(t, baseDocsOutDir) })
+	t.Run("tree", func(t *testing.T) { util.AssertDirEqual(t, basePackageTreeJSONOutDir) })
```
iwahbe, line 57: Each call to `util.AssertDirEqual` must be in a separate test, since the test path is used to generate files.

iwahbe (commented):
There isn't that much code change here, but this puts the results of docs generation under test, which means there is a checked-in docs folder (random).

t0yv0 (approved):

PR #5934, opened 2024-11-08
Increase bucket sync parallelism

iwahbe:
Increase the concurrency limit when syncing to s3.

| Limit | Sync time (minutes) |
|--|--|
| 10 (default) | 15 |
| 16 (`nproc`) | 10 |
| 32 (`nproc * 2`) | 5 |

This PR leaves the concurrency limit at `nproc * 2`.

### DIFF
--- a/scripts/ci/sync.sh
+++ b/scripts/ci/sync.sh
@@ -47,7 +47,7 @@ aws s3api put-public-access-block --bucket $destination_bucket --public-access-b
 aws s3api put-bucket-ownership-controls --bucket $destination_bucket --ownership-controls="Rules=[{ObjectOwnership=ObjectWriter}]"
 aws s3api put-bucket-acl --bucket $destination_bucket --acl bucket-owner-full-control --acl public-read
 
-aws configure set default.s3.max_concurrent_requests "$(nproc --all)" # Default is 10
+aws configure set default.s3.max_concurrent_requests "$(( "$(nproc --all)" * 2))" # Default is 10
 aws s3api put-bucket-tagging --bucket $destination_bucket --tagging "TagSet=[{$(aws_owner_tag)}]" --region "$(aws_region)"
 
 # Make the bucket an S3 website.


tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -35,6 +36,7 @@ import (
 	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/pulumi/registry/tools/resourcedocsgen/pkg"
+	concpool "github.com/sourcegraph/conc/pool"
```
t0yv0, line 39: Any chance this can stick to stdlib? Perhaps no big deal.

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -126,31 +128,36 @@ func genResourceDocsForAllRegistryPackages(registryRepoPath, baseDocsOutDir, bas
 		return errors.Wrap(err, "reading the registry packages dir")
 	}
 
+	pool := concpool.New().WithErrors().WithMaxGoroutines(runtime.NumCPU())
 	for _, f := range metadataFiles {
-		glog.Infof("=== starting %s ===\n", f.Name())
-		glog.Infoln("Processing metadata file")
-		metadataFilePath := filepath.Join(registryPackagesPath, f.Name())
+		f := f
+		pool.Go(func() error {
```
t0yv0, line 134: Seems interesting. There is always the danger that something was not fully isolated and concurrent execution will surface bugs but worth a shot.

t0yv0 (approved):

tools/resourcedocsgen/cmd/docs/registry.go
iwahbe: This is from #5917. The only changes unique to this PR are in scripts files. I'll rebase before merging to make it clearer.

Issue #3883, opened 2024-01-31
Move the docs for Pulumi-managed providers into their respective repos


cnunciato:
Today, we check the Overview and Installation/Configuration pages for Pulumi-managed providers into the Registry repo directly, whereas for community-managed providers, we source them from their upstream repos' `docs` folders. We should have our provider documentation managed in the same way, so those docs can be versioned (and follow the same ingestion process) as our community-managed providers.

PR #5965, opened 2024-11-12
Mark fetched `_index.md` and `installation-configuration.md` as such

iwahbe:
This will make it much easier to move the source of truth for these docs into their respective upstream repos.

I'm assuming that all of our `_index.md` and `configuration-installation.md` files use YAML front-matter. Running a check-metadata pass with this build will tell us definitively.

This makes https://github.com/pulumi/registry/issues/3883 much easier.

### DIFF
--- a/themes/default/content/registry/packages/aci/_index.md
+++ b/themes/default/content/registry/packages/aci/_index.md
@@ -1,4 +1,6 @@
 ---
+# WARNING: this file was fetched from https://raw.githubusercontent.com/netascode/pulumi-aci/v0.0.6/docs/_index.md
+# Do not edit by hand unless you're certain you know what you are doing!
 title: ACI
 meta_desc: Provides an overview of the Cisco ACI Provider for Pulumi.
 layout: overview
--- a/themes/default/content/registry/packages/aci/installation-configuration.md
+++ b/themes/default/content/registry/packages/aci/installation-configuration.md
@@ -1,4 +1,6 @@
 ---
+# WARNING: this file was fetched from https://raw.githubusercontent.com/netascode/pulumi-aci/v0.0.6/docs/installation-configuration.md
+# Do not edit by hand unless you're certain you know what you are doing!
 title: ACI Installation & Configuration
 meta_desc: Information on how to install the ACI provider.
 layout: installation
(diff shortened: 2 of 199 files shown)


iwahbe:
Running this safely will show that the PR is safe:

```sh
for p in themes/default/data/registry/packages/*; do         
repo="$(yq .repo_url $p)"
version="$(yq .version $p)"
schemaFile="$(yq .schema_file_path $p)"
./tools/resourcedocsgen/resourcedocsgen metadata --repoSlug "${repo#https://github.com/}" --version "$version" --schemaFile "$schemaFile"
done
```

iwahbe:
It is indeed safe. Trialing a metadata update unfortunately shows 38 places where we have manually updated files that should be auto-generated. I'll open a separate issue to address those.

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -208,12 +208,28 @@ func PackageMetadataCmd() *cobra.Command {
...
+				content = bytes.ReplaceAll(content, []byte("\r\n"), []byte("\n"))
+
+				if rest, ok := bytes.CutPrefix(bytes.TrimLeft(content, "\n\t\r "), []byte("---\n")); ok {
+					content = append([]byte(`---
+# WARNING: this file was fetched from `+requiredFilePath+`
+# Do not edit by hand unless you're certain you know what you are doing!
+`), rest...)
+				} else {
+					return fmt.Errorf(`expected file %s to start with YAML front-matter ("---\n"), found leading %q`,
+						requiredFilePath, string(content))
```
guineveresaenger, line 229: That could be a _lot_ of content; would it make sense to truncate?

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -208,12 +208,28 @@ func PackageMetadataCmd() *cobra.Command {
...
+				content, err := readRemoteFile(requiredFilePath, repoSlug.owner)
 				if err != nil {
 					return err
 				}
+				if len(content) == 0 {
+					continue
+				}
+
+				// Normalize end of line representation
+				content = bytes.ReplaceAll(content, []byte("\r\n"), []byte("\n"))
```
guineveresaenger, line 220: Let's note this in the PR description as well, since it's orthoganal.

guineveresaenger (approved):
Nice! I like the version in the repo path; good call.

tools/resourcedocsgen/cmd/metadata.go
iwahbe: Yes, yes it would.

I had it truncated earlier, but changed it when I was debugging another issue. Thanks for catching that!

PR #5990, opened 2024-11-13
Fix underline on hover for guide count badges.

### LINKED ISSUE #5942: how-to badge should not underline on hover
## overview
underline hover styles should not apply to count badges.

## screenshot
Screen Shot 2023-07-21 at 4 15 09 PM

thoward:
fixes #5942 (see original issue for context)

### DIFF
--- a/themes/default/theme/src/scss/docs/_main-nav.scss
+++ b/themes/default/theme/src/scss/docs/_main-nav.scss
@@ -36,6 +36,10 @@ nav.main-nav {
             text-underline-offset: 6px;
             color: map-get($blue, 750);
         }
+        
+        span.count-badge:hover {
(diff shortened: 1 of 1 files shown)


mjeffryes:
Unfortunately, it doesn't look like this works in the preview. Apparently you can't opt out of text decoration in a child span (https://stackoverflow.com/questions/8681301/text-decoration-for-link-and-span-inside-this-link), so to avoid this underline, we have to either move the count out of the link or change the text decoration of the anchor tag to none and use a span to add the underline to just the name and not the count. (I tend to lean toward the first option.)

thoward:
ooo that's a tricky gotcha! cancelling this PR

PR #5991, opened 2024-11-14
Generated warning for metadata files

iwahbe:
## Description

- 7cf22cafe21586a3e2cbb5ffb31dcfda68e358a2 actually adds the warning.
- 2264c9f766c81997d3722256f629381b4a65bab6 updates some (but not all) metadata files with the warning.


guineveresaenger (approved):
The logic looks fine to me, but there's a bit of context missing - which packages are left _without_ this warning, and why?Is it a matter of waiting until they all percolate, or are there packages that don't use resourcedocsgen, or something else entirely? Thank you!

Issue #5948, opened 2024-11-11
Include package published in api docs URLs


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

API doc URLs look like this:

```
https://www.pulumi.com/registry/packages/{package-name}/api-docs/{module}/{lowercase-resource-or-function-name}/
```

For example:

```
https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketv2/
```

This URL scheme limits future improvements like name-spaced packages or versioned packages. We should transition the new URL structure to a more flexible setup:

```
https://www.pulumi.com/registry/packages/{source}/{owner}/{pacakge}/{version=latest}/api-docs/{kind=r|d}/{module}/{resource}
```

AWS S3 BucketV2 would now be:

```
https://www.pulumi.com/registry/packages/pulumi/pulumi/aws/latest/api-docs/r/s3/bucketv2/
```

This will allow us to deconflict packages from different sources (like OpenTofu or CDK constructs) and native Pulumi packages, as well as deconflict packages from different publishers but with the same name. This makes space for https://github.com/pulumi/registry/issues/5846, which we know we want to tackle in the future.

PR #5995, opened 2024-11-14
Add a Published version API

iwahbe:
This is a very simple API, just allowing other repositories to get the latest version of a provider. The API is listed as v0, marking it as experimental and without backwards comparability guarantees. I don't expect this API to survive the [IA transition](https://github.com/pulumi/registry/issues/5948) the registry will undergo.

### DIFF
--- a/themes/default/content/registry/api/v0/_content.gotmpl
+++ b/themes/default/content/registry/api/v0/_content.gotmpl
@@ -1,0 +1,10 @@
+{{ range $package := .Site.Data.registry.packages }}
+  {{ $.AddResource (dict
+    "content" (dict
+      "mediaType" "application/json"
+      "value" (dict "published-latest" $package.version | jsonify)
+    )
+    "kind" "page"
+    "path" (printf "packages/%s/version" $package.name)
+  )}}
+{{ end }}


guineveresaenger (commented):
I'm not sure I entirely understand the use case.

Is there a reason we wouldn't be able to query the version found in `themes/default/data/registry/packages/<provider>.yaml` ?

iwahbe:
> I'm not sure I entirely understand the use case.

The use case is a into-registry pipeline being able to check what version a provider is currently at, to determine if an update should be sent to the registry.
 
> Is there a reason we wouldn't be able to query the version found in `themes/default/data/registry/packages/<provider>.yaml` ?

AFAIK Hugo doesn't present the data folder in the generated site.

guineveresaenger:
Sure, but why would we need to query an endpoint on the generated site? Couldn't we query this repo? I suppose I'm not sure how sophisticated this pipeline needs to be, but it's been my impression that these files are absolutely meant to be used for version update reconciliation.

iwahbe:
> Sure, but why would we need to query an endpoint on the generated site? Couldn't we query this repo? I suppose I'm not sure how sophisticated this pipeline needs to be, but it's been my impression that these files are absolutely meant to be used for version update reconciliation.

I'm trying to move to a model where the registry presents an API distinct from it's implementation details. If we want to move our metadata files from `themes/default/data/registry/packages/` to `data/registry/packages/` (where they should be) or change the way they are represented, I don't want that to break other systems. Likewise, if the registry decides to remove the experimental API, then I want to be able to do so without changing the layout of the repo.

iwahbe:
I revisited this, moving the entire API into Hugo instead of `resourcedocsgen`.

iwahbe:
> Your site preview for commit [443ace0](https://github.com/pulumi/registry/commit/443ace020ad8b3ffd4ea407d39ab1814264456b8) is ready! 🎉
> 
> http://registry--origin-pr-5995-443ace02.s3-website.us-west-2.amazonaws.com/registry.

To "use" the API, you would run

```console
$ curl http://registry--origin-pr-5995-443ace02.s3-website.us-west-2.amazonaws.com/registry/api/v0/packages/azure/version
{"version":"v6.9.0"}
```

guineveresaenger (approved):
Discussed offline and LGTM.

Thank you for the smaller change as well.

PR #6077, opened 2024-11-26
fix(docs) remove yaml example for hcloud provider

PRIHLOP:
## Description

This PR removes the yaml example from the documentation because the HCloud provider doesn't currently support the yaml language.

### DIFF
--- a/themes/default/content/registry/packages/hcloud/_index.md
+++ b/themes/default/content/registry/packages/hcloud/_index.md
@@ -8,7 +8,7 @@ layout: package
 ---
 ## Installation
 
-The hcloud provider is available as a package in all Pulumi languages:
+The hcloud provider is available as a package in all Pulumi languages(except yaml):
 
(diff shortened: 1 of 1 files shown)


iwahbe:
@PRIHLOP pulumi-hcloud should support YAML. Why do you think that it does not?

PRIHLOP:
I tried to work with YAML without results. I also tried to convert from YAML to Python, and as a result, I got a clear `__main__.py` file only with import options and without any instructions. Also, I checked the [repository](https://github.com/pulumi/pulumi-hcloud) and didn't find any mentions of YAML in it, no tests, and nothing about YAML.

iwahbe:
YAML works fine for me. This stood up without issue:

```yaml
name: dev-yaml
runtime: yaml
resources:
  network:
    type: hcloud:Network
    properties:
      ipRange: "10.0.1.0/24"
```

The way Pulumi Providers and Pulumi YAML are set-up, all modern providers work with YAML, without the providers needing to opt-in to working with YAML.

If you want, you can file an issue that some resource isn't working in Pulumi YAML in the pulumi-hcloud repo.

PR #6078, opened 2024-11-26
fix(docs) set correct repository path for registry pages

PRIHLOP:
## Description

Registry pages have their repository now and don't need to redefine the repo path.


iwahbe (approved):
This looks good to me.

PR #6097, opened 2024-11-27
Enable remote publishing

iwahbe:
This PR adds a repository dispatch target to accept incoming docs PRs for providers without host repositories.

Unfortunately, I'm not able to test dispatch events without them being present on the default branch. For risk mitigation:
- The edited file does not effect community providers *at all*.
- There are minimal changes to the existing job.

### DIFF
--- a/.github/workflows/publish-provider-update.yml
+++ b/.github/workflows/publish-provider-update.yml
@@ -1,17 +1,48 @@
 env:
   GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
+  # The path within the Pulumi provider repo to the schema file.
   PROVIDER_SCHEMA_PATH: ${{ github.event.client_payload.schema-path }}
+  # The short-name of the provider, used in URL names. This is effectively the provider's
+  # identity in the Pulumi registry right now.
   PROVIDER_SHORT_NAME: ${{ github.event.client_payload.project-shortname }}
+  # The git tag that the provider is released with.
+  #
+  # This is only meaningful when the provider is based on a GitHub repo.
   PROVIDER_VERSION: ${{ github.event.client_payload.ref }}
+  # A public URL which hosts the provider's Pulumi schema.
+  #
+  # This URL must be durable, as it will be required available to build the registry.
+  PROVIDER_SCHEMA_URL: ${{ github.event.client_payload.schema-url }}
+  # A public URL which hosts the provider's  _index.md file.
+  #
+  # This URL must be durable, as it will be required available to build the registry.
+  PROVIDER_INDEX_URL:  ${{ github.event.client_payload.index-url }}
 
 name: provider docs build
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-provider-update.yml
```diff
@@ -35,12 +65,46 @@ jobs:
...
           provider_version: ${{ env.PROVIDER_VERSION }}
+
+  build-remote-provider-docs:
+    if: ${{ github.event.event_type == remote-provider }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate Inputs
+        run: |
+          if ![ "${{ env.PROVIDER_VERSION }}     = "" ]; then echo "PROVIDER_VERSION is not respected; exit 1;    fi
+          if ![ "${{ env.PROVIDER_SCHEMA_PATH }} = "" ]; then echo "PROVIDER_VERSION is not respected; exit 1;    fi
```
guineveresaenger, line 76:
```suggestion
          if ![ "${{ env.PROVIDER_SCHEMA_PATH }} = "" ]; then echo "PROVIDER_SCHEMA_PATH is not respected; exit 1;    fi
```

Also I don't speak fluent bash but that doesn't look right to me with the placement of the quotes.

I also can't tell without context what are we asserting. If I ran into this error, I wouldn't know what do do next.

.github/workflows/publish-provider-update.yml
```diff
@@ -1,17 +1,47 @@
...
+      # An update for a provider that is hosted in a Pulumi repo.
+      #
+      # Required inputs are:
+      # - PROVIDER_SHORT_NAME
+      # - PROVIDER_VERSION
+      #
+      # Optional inputs are:
+      # - PROVIDER_SCHEMA_PATH
       - resource-provider
+      # An update for a provider that is hosted opaquely.
```
guineveresaenger, line 34: This is a new term - presumbably referring to third-party and dynamic providers? Could we make this clear?

Or just say "hosted elsewhere".

guineveresaenger (approved):
let's try it out!

.github/workflows/publish-provider-update.yml
iwahbe: > Also I don't speak fluent bash but that doesn't look right to me with the placement of the quotes.

There were some missing quotes. 

> I also can't tell without context what are we asserting. If I ran into this error, I wouldn't know what do do next.

These operate mostly as asserts. I've made them a bit clearer.

.github/workflows/publish-provider-update.yml
iwahbe: I've changed the event name to `push-provider-update`.

iwahbe:
It's unfortunately impossible to test registry dispatch GH actions that are not on the default branch, so I'm going to merge and then start testing. I'd be very surprised if this works first time, so bear with me on fixes.

PR #6209, opened 2024-12-11
Update checklist for community package contributions

mjeffryes:
## Description

We want all new packages added to the registry to support all core Pulumi languages and to have an identified maintainer contact. This PR adds those items to the PR checklist.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -18,7 +18,7 @@ If this pull request adds a new package, please verify/document:
   - [ ] at least one representative example in all supported languages (i.e. TypeScript, Python, Go and C#).
   - [ ] a front-matter property for the `layout` set to `package`.
 - [ ] The package repo contains an Installation and Configuration doc (`/docs/installation-configuration.md`) that includes:
-  - [ ] links to SDKs in all supported languages (ie. should link to published SDKs in NPM, PyPi and NuGet).
+  - [ ] links to SDKs in all supported languages (i.e. should link to published SDKs in NPM, PyPi and NuGet).
   - [ ] a copyable command for installing the resource plugin if necessary.
   - [ ] an example of configuring the provider with `pulumi config set`.
   - [ ] an example of configuring the provider with environment variables.


.github/pull_request_template.md
```diff
@@ -14,14 +15,14 @@ If this pull request adds a new package:
   - [ ] a valid logo URL that points to a PNG whose dimensions conform to the others in this repo (e.g., 100x100).
 - [ ] The package repo contains an Overview doc (`/docs/_index.md`) that includes:
   - [ ] a brief explanation of what the package is and what it does.
-  - [ ] at least one representative example in all supported languages.
+  - [ ] at least one representative example in all supported languages (ie. ts/js, python, go and c#).
```
guineveresaenger, line 18: What about Java? this would still retain a discrepancy between all official Pulumi languages and what's required here.

guineveresaenger, line 18:
```suggestion
  - [ ] at least one representative example in all supported languages (i.e. TypeScript, Python, Go and C#).
```

.github/pull_request_template.md
```diff
@@ -14,14 +15,14 @@ If this pull request adds a new package:
...
   - [ ] a front-matter property for the `layout` set to `package`.
 - [ ] The package repo contains an Installation and Configuration doc (`/docs/installation-configuration.md`) that includes:
-  - [ ] links to SDKs in all supported languages.
+  - [ ] links to SDKs in all supported languages (ie. should link to published SDKs in NPM, PyPi and NuGet).
```
guineveresaenger, line 21:
```suggestion
  - [ ] links to SDKs in all supported languages (i.e. should link to published SDKs in NPM, PyPi and NuGet).
```

guineveresaenger (changes requested):
Should we require Java? Why/why not?

guineveresaenger (approved):

mjeffryes:
Hmmm... I'm inclined to give package maintainers a pass on Java because it's represents a pretty tiny slice of users and we know from experience that publishing packages for Java is a pain. I'd really like to be able to offer publishing support for package authors so they would not need to jump through hoops for even NPM, PyPi and NuGet. Hopeful that we'll get there one day.

iwahbe (commented):
This is a major pain to authors. Why don't we allow them to use parameterized providers instead of published SDKs?

mjeffryes:
> This is a major pain to authors. Why don't we allow them to use parameterized providers instead of published SDKs?

I don't quite follow you here? I assume you mean letting them just include docs for how to use with local package support? Two concerns with that:
1) Doesn't work for native providers or components
2) If they are explicitly bridging a TF provider, the APIs likely won't align; I'd rather just advertise the local packages based docs for all languages

mjeffryes:
> I keep forgetting that [pulumi/pulumi#17056](https://github.com/pulumi/pulumi/issues/17056) isn't in yet. I'll come back to this proposal when [pulumi/pulumi#17056](https://github.com/pulumi/pulumi/issues/17056) is closed.

Sounds good! I agree it would be way better not to have to force publishing woes on everyone who wants to add a package to the registry!

PR #6225, opened 2024-12-13
Allow missing metadata files

iwahbe:
This is necessary to allow Any Terraform Providers to be added to the registry, since they are currently built without `installation-configuration.md` URLs.

Stacked on #6235


tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -215,22 +213,36 @@ func PackageMetadataCmd() *cobra.Command {
...
 			}{
-				{"_index.md", indexFileURL},
-				{"installation-configuration.md", installationConfigurationURL},
+				{"_index.md", indexFileURL, true},
+				{"installation-configuration.md", installationConfigurationURL, false},
+			}
+
+			// If the URL is directly specified for any remote file, then it
+			// must be specified for all remote files.
+			var explicitURLs bool
```
guineveresaenger, line 226: It looks like we only need this because we added separate inputs for index and install file URLs.

Is it possible to simplify this and continue to keep using the `packageDocsDir`? We're not planning to have the two files exist in different folders, do we? (I may be missing a reason here to have the option to input the explicit file names rather than the docs dir.)

Then this turns in to a simple "if file exist" check further down for installation-configuration.

guineveresaenger (approved):
one suggestion which I think would make the logic a bit more straightforward

tools/resourcedocsgen/cmd/metadata.go
iwahbe: > It looks like we only need this because we added separate inputs for index and install file URLs.

Yes. We added support for index and installation configuration files by URL in https://github.com/pulumi/registry/pull/5896. #5896 added separate URLs for each file, to avoid making any assumptions about the file source (beyond a `GET` returning a file).

> Is it possible to simplify this and continue to keep using the `packageDocsDir`?

I don't know what you mean. We are continuing to use `packageDocsDir`. The URL flags (`--schemaFileURL`, `--indexFileURL` and `--installationConfigurationFileURL`) all specify where `resourcedocsgen` should _get_ the file from. `packageDocsDir` specifies  where to _put_ the downloaded docs[^1] file.

> I may be missing a reason here to have the option to input the explicit file names rather than the docs dir.

We are using full URLs instead of a docs dir because the URLs impose less assumptions about where the docs are located, while still being precise. Under the URL scheme, we allow users to not specify a file, but if they do, then we make sure it exists.

Hypothetically, we could allow users to specify a path prefix (like `my-cdn.aws.com/514738/docs/`) and then do existence checks on `${prefix}/_index.md` and `${prefix}/installation-configuration.md`. We don't do it that way now. I decided not to do it that way because:
1. It Creates an implicit contract for which paths exist at the prefix. The user needs to know that if they put content at that path, then something will happen.
2. Its error prone. A user might be confused why their docs were dropped at `${prefix}/installation-configuraton.md`[^2], where as explicit URLs will error.

[^1]: The schema file is not stored in repo.
[^2]: "configuraton" is misspelled. It's missing the last "i".

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -119,17 +119,15 @@ func PackageMetadataCmd() *cobra.Command {
...
 				if err != nil {
 					return errors.Wrapf(err, "parsing repo url %q from schema", mainSpec.Repository)
 				}
+
 				if r.Hostname() != "github.com" {
 					return fmt.Errorf("mainSpec.Repository host must be from GH, found %q", r.Hostname())
 				}
 
-				if err := repoSlug.Set(r.Path); err != nil {
+				if err := repoSlug.Set(strings.Trim(r.Path, "/")); err != nil {
```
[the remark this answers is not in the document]
iwahbe: It fixes a different bug. I'll spin it out into it's own PR.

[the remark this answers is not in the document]
iwahbe: PR is https://github.com/pulumi/registry/pull/6235.

iwahbe:
@lunaris I took a cleanup pass that I think should address your clarity concern. PTAL

tools/resourcedocsgen/cmd/metadata.go
guineveresaenger: I saw you added a field to the file struct instead, that does make it more clear.

Still wanted to reply - 

> packageDocsDir specifies where to put the downloaded docs file

Ah, my bad - I should have said, "can we continue to use the assumption that these files are sourced from a `url/docs/` folder that the user provides" (this then maps to the structure of `packageDocsDir`). It is what we ask of third party providers today.

> Hypothetically, we could allow users to specify a path prefix

Yeah, that was my thought. I was coming down on the side of "we have this contract already today",  since we do ask third-party provider authors to place these files in a `docs` folder.

I'm not sure that odd things would happen to all files in this folder - if there's docs files that don't have the two expected names, nothing will occur.
I also believe in our ability to send a good error message when verifying the existence of expected docs.

guineveresaenger (approved):
let's merge - thank you for adding tests, that is awesome. 🎉

PR #6246, opened 2024-12-17
Split metadata into 2 different commands

iwahbe:
#6225 shows the increasing complexity of the `resourcedocsgen metadata`, and the accompanying difficulty of understanding how different flags interact.

A large part of the difficulty stems from the fact that there are really two commands:

1. Extract metadata from a GitHub repo
2. Extract metadata from a series of URLs

This PR separates the two commands into 2 actual sub-commands:

```console
$ resourcedocsgen metadata from-github
$ resourcedocsgen metadata from-urls
```

Stacked on top of https://github.com/pulumi/registry/pull/6225

### DIFF
--- a/tools/resourcedocsgen/cmd/metadata.go
+++ b/tools/resourcedocsgen/cmd/metadata.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/blang/semver"
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
@@ -72,15 +73,12 @@ func packageMetadataFromURLsCmd(metadataDir, packageDocsDir *string) *cobra.Comm
 		SilenceUsage: true,
 	}
 
-	// Inputs (required)
+	// Required Flags
 
 	providerName := cmd.Flags().String("providerName", "", "The name of the provider e.g. aws, aws-native. "+
 		"Required when there is no schemaFile flag specified.")
-	schemaURL := cmd.Flags().String("schemaFileURL", "",
-		`The URL from which the schema can be retrieved.
-
-schemaFileURL takes precedence over schemaFile.`)
-	indexURL := cmd.Flags().String("indexFileURL", "",
+	schemaFileURL := cmd.Flags().String("schemaFileURL", "", `The URL from which the schema can be retrieved.`)
+	indexFileURL := cmd.Flags().String("indexFileURL", "",
 		`The URL from which the docs/_index.md file can be retrieved.`)
 
 	contract.AssertNoErrorf(stderrors.Join(
@@ -89,13 +87,13 @@ schemaFileURL takes precedence over schemaFile.`)
 		cmd.MarkFlagRequired("indexFileURL"),
 	), "impossible")
 
-	// Inputs (optional)
+	// Optional Flags
 	installationConfigurationURL := cmd.Flags().String("installationConfigurationFileURL", "",
 		`The URL from which the docs/installation-configuration.md file can be retrieved.`)
 
 	cmd.PreRunE = func(cmd *cobra.Command, args []string) error {
 		var errs []error
-		if *indexURL == "" {
+		if *indexFileURL == "" {
 			errs = append(errs, stderrors.New(`missing URL for required file "_index.md"`))
 		}
 
@@ -103,25 +101,26 @@ schemaFileURL takes precedence over schemaFile.`)
 	}
 
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
-		// Inputs are now cooked, so dereference them.
+		// By the time this function runs, input *string values have been set by
+		// cobra.
 		//
 		// They are guaranteed to be non-nil.
 
 		providerName := *providerName
-		schemaURL := *schemaURL
-		indexURL := *indexURL
+		schemaFileURL := *schemaFileURL
+		indexFileURL := *indexFileURL
 		installationConfigurationURL := *installationConfigurationURL
 		metadataDir := *metadataDir
 		packageDocsDir := *packageDocsDir
 
 		// Get the schema.
 
-		mainSpec, err := readRemoteSchemaFile(schemaURL, "")
+		mainSpec, err := readRemoteSchemaFile(schemaFileURL, "")
 		if err != nil {
 			return errors.WithMessage(err, "unable to read remote schema file")
 		}
 
-		err = writePackageMetadata(mainSpec, providerName, schemaURL, metadataDir, time.Now())
+		err = writePackageMetadata(mainSpec, providerName, schemaFileURL, metadataDir, time.Now())
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


iwahbe:
I have manually verified that the `from-github` component works for all packages currently in the registry with this script:

```sh
go build -C ./tools/resourcedocsgen . && for p in themes/default/data/registry/packages/*; do
repo="$(yq .repo_url "$p")"
version="$(yq .version "$p")"
publisher="$(yq .publisher "$p")"
if ! yq .schema_file_path "$p" --exit-status 2>1 > /dev/null; then
    schemaFile="$(yq .schema_file_url "$p" | sed -e "s#^https://raw.githubusercontent.com/.*/$version/##")"
else
    schemaFile="$(yq .schema_file_path "$p")"
fi
GITHUB_TOKEN=$(gh auth token) ./tools/resourcedocsgen/resourcedocsgen metadata from-github \
            --repoSlug "${repo#https://github.com/}" \
            --version "$version" \
            --schemaFile "$schemaFile" \
            --publisher "$publisher"
echo Processed "$p"
done
```

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -47,264 +45,315 @@ var featuredPackages = []string{
...
+		return nil
+	}
 
 	return cmd
 }
 
+func writePackageMetadata(
+	spec *schema.PackageSpec, providerName, schemaURL, metadataDir string, updatedOn time.Time,
+) error {
+	// Validate the schema for usage.
```
guineveresaenger, line 301: I think this comment could be more specific,

"A schema needs to have X and Y for the registry to publish it successfully."

Otherwise the code is pretty self-explanatory here.

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -558,3 +578,45 @@ func inferPublishDate(repo repoSlug, version string) (time.Time, bool, error) {
 
 	return commit.Commit.Author.Date, true, nil
 }
+
+func readDocsFile(url string) ([]byte, error) {
```
guineveresaenger, line 582: It looks like we're doing more than reading here.

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -47,264 +45,315 @@ var featuredPackages = []string{
...
-	cmd.Flags().BoolVar(&component, "component", false,
-		"Whether or not this package is a component and not a provider")
-	cmd.Flags().StringVar(&metadataDir, "metadataDir", "themes/default/data/registry/packages",
-		"The location to save the metadata - this will default to the folder "+
-			"structure that the registry expects (themes/default/data/registry/packages)")
-	cmd.Flags().StringVar(&packageDocsDir, "packageDocsDir", "",
-		"The location to save the package docs - this will default to the folder "+
-			"structure that the registry expects (themes/default/content/registry/packages)")
+
+	// Apply complex defaults
```
guineveresaenger, line 194: What is complex about it?

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -47,264 +45,315 @@ var featuredPackages = []string{
...
+			mainSpec.Publisher = publisher
+		}
+
+		err = writePackageMetadata(mainSpec, providerName, schemaURL, metadataDir, publishedDate)
+		if err != nil {
+			return errors.Wrap(err, "generating package metadata")
+		}
+
+		if packageDocsDir == "" {
+			// if the user hasn't specified an packageDocsDir, we will default to
```
guineveresaenger, line 251:
```suggestion
			// if the user hasn't specified a packageDocsDir, we will default to
```

guineveresaenger (commented):
I like this in general. I think the comments could have a bit more detail, and there's some text that no longer makes sense.

tools/resourcedocsgen/cmd/metadata.go
iwahbe: Complex was the wrong word. I have changed it to:

```go
// Apply default values that depend on other inputs.
```

tools/resourcedocsgen/cmd/metadata.go
iwahbe: I was hoping that the code itself would serve to explain. I'll add a list of validations to the comment.

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -47,264 +45,315 @@ var featuredPackages = []string{
...
+	if spec.Name != providerName {
+		validationErrors = append(validationErrors, fmt.Errorf(
+			"--providerName doesn't match the schema name: %q != %q", providerName, spec.Name))
+	}
+
+	if spec.Version == "" {
+		validationErrors = append(validationErrors, fmt.Errorf("schema for %q has no version", spec.Name))
+	}
+
+	category, err := getPackageCategory(spec)
```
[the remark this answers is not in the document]
iwahbe: It's used to display "Type" and "Use Case" in the registry:
Screenshot 2025-01-06 at 4 40 00 PM

We still use it.

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -47,264 +45,315 @@ var featuredPackages = []string{
...
+
+		contract.Assertf(version != "", "version is a required field")
+		mainSpec.Version = version
+
+		mainSpec.Repository = "https://github.com/" + repoSlug.String()
+
+		// If the schema for this package does not have the displayName, and it
+		// does have a pre-baked title, use that title.
+		if mainSpec.DisplayName == "" {
+			// Eventually all of Pulumi's own packages will have the displayName
```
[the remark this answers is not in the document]
iwahbe: Not much progress has been made, but some has. I changed "will" to "should".

tools/resourcedocsgen/cmd/metadata.go
lunaris: We could even pull validation out into its own helper function then this (what we are validating) would naturally appear as the doc-comment for that function

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -47,264 +45,315 @@ var featuredPackages = []string{
...
-			remoteFiles := []struct {
-				name     string
-				required bool
-				url      string
-			}{
-				{"_index.md", true, urlSource.indexURL()},
-				{"installation-configuration.md", false, urlSource.installationConfigurationURL()},
-			}
+	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		// Inputs are now cooked, so dereference them.
```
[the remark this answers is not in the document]
iwahbe: I changed the comment around this. Clarify stages is what I was trying to do in https://github.com/pulumi/registry/pull/6246#discussion_r1902191788.

tools/resourcedocsgen/cmd/metadata.go
iwahbe: I added a comment describing what the function expects. `readDocsFile` is conceptually a wrapper around `readRemoteFile`, hence the name.

guineveresaenger (approved):

PR #6252, opened 2024-12-17
Start on title lookup removal

iwahbe:
This PR makes some progress on pulumi/pulumi#7813, documenting which title lookups have an associated display name (and which PR added the display name).

### DIFF
--- a/tools/resourcedocsgen/pkg/lookup.go
+++ b/tools/resourcedocsgen/pkg/lookup.go
@@ -112,21 +112,26 @@ var CategoryLookup = map[string]PackageCategory{
 	"yandex":                              PackageCategoryCloud,
 }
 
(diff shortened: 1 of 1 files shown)


tools/resourcedocsgen/pkg/lookup.go
```diff
@@ -112,21 +112,22 @@ var CategoryLookup = map[string]PackageCategory{
 	"yandex":                              PackageCategoryCloud,
 }
 
-// TODO[pulumi/pulumi#7813]: Remove this lookup once display name is available in
-// the Pulumi schema.
+// TitleLookup is a map pf package name to the desired display name
```
thomas11, line 115:
```suggestion
// TitleLookup is a map of package name to the desired display name
```

thomas11 (approved):

PR #6270, opened 2024-12-18
Corrections to the MSI auth instructions for Azure Native

thomas11:
## Description

The instructions to configure MSI authentication in Azure Native contained two inaccuracies:

1. The user needs to turn it on explicitly.
2. Configuring custom endpoints is not possible anymore since pulumi/pulumi-azure-native#3630 and pulumi/pulumi-azure-native#3654.

### DIFF
--- a/themes/default/content/registry/packages/azure-native/installation-configuration.md
+++ b/themes/default/content/registry/packages/azure-native/installation-configuration.md
@@ -297,14 +297,11 @@ as shown above except use `clientCertPath`/`ARM_CLIENT_CERTIFICATE_PATH` instead
 
 ### Authenticate using Managed Service Identity (MSI)
 
-MSI is [configured in Azure](https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview).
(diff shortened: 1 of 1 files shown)


mjeffryes (approved):
I think @iwahbe has been trying to transition these docs to all live in the provider repos and just ingested by registry builds; you might check with him if we could do that for AZ native too. 


> Configuring custom endpoints is not possible anymore
Is this going to be perceived as a regression by our users? Is there no way to support it with the new libraries?

thomas11:
> > Configuring custom endpoints is not possible anymore
> 
> Is this going to be perceived as a regression by our users? Is there no way to support it with the new libraries?

I wasn't able to find a way to support this with the new libraries. However, I found that to be ok because it's not necessary anymore. The new library automatically handles the cases where users had to use custom endpoints before: https://github.com/Azure/azure-sdk-for-go/blob/sdk/azidentity/v1.8.0/sdk/azidentity/managed_identity_client.go#L143

iwahbe (approved):

iwahbe:
We are moving these docs out of the registry, but we havn't done that for AZ native yet. I don't have a problem with editing this file until it is moved.

PR #6466, opened 2025-01-21
Parse templates once

iwahbe:
We have a global set of templates, this PR moves us to parse the templates once at `init` time and wraps each template in function. This puts us in a place to start strongly typing template access.

### DIFF
--- a/tools/resourcedocsgen/pkg/docs/gen.go
+++ b/tools/resourcedocsgen/pkg/docs/gen.go
@@ -28,7 +28,7 @@ import (
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/pulumi/pulumi/pkg/v3/codegen/pcl"
-	. "github.com/pulumi/registry/tools/resourcedocsgen/pkg/docs/templates"
+	"github.com/pulumi/registry/tools/resourcedocsgen/pkg/docs/templates"
 
 	"github.com/golang/glog"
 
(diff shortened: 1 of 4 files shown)


lunaris (approved):
Nice! I have a few comments, which may already be in the works/things you have thought of:

* I feel like you could push the param separator stuff down and just have a "parameters (plural) template". If it's common for most the languages, have some private stuff in the templates package that takes the unique bits (individual param rendering, the loop that inserts separators) and returns the plural template so that callers can simplify?
* Probably makes sense to keep this PR light, but feels like an opportunity to pull out a "language to template factory" function later on that captures the `switch lang` duplicated in a few places.
* While we're pulling out this interface, I think it's a great opportunity to document the public API that we've just built -- that is, the stuff in the `templates` package. I think also a package-level doc comment that captures the great essence from this PR would be useful for people returning to it later!

iwahbe:
> Nice! I have a few comments, which may already be in the works/things you have thought of:
> 
> * I feel like you could push the param separator stuff down and just have a "parameters (plural) template". If it's common for most the languages, have some private stuff in the templates package that takes the unique bits (individual param rendering, the loop that inserts separators) and returns the plural template so that callers can simplify?
> * Probably makes sense to keep this PR light, but feels like an opportunity to pull out a "language to template factory" function later on that captures the `switch lang` duplicated in a few places.
> * While we're pulling out this interface, I think it's a great opportunity to document the public API that we've just built -- that is, the stuff in the `templates` package. I think also a package-level doc comment that captures the great essence from this PR would be useful for people returning to it later!

I've merged with just documentation improvements. This is the first of a couple of cleanup PRs.

PR #6478, opened 2025-01-22
Refactor: Add structured language types

iwahbe:
Clean up codegen by using an enum type for docs generation. This required some surprising changes, since existing languages lists included `"csharp", "go", "java", "javaargs", "python", "pythonargs", "typescript", "nodejs", "javascript"` and `"yaml"`. This required adding 60 lines of new generation code. Other additions are either small changes or the code in `language`.

This is a pure refactor. Test outputs are unchanged and users will observe *zero* difference.

---

This PR can be reviewed commit by commit. Each commit leaves the repo in a valid state with tests passing.


tools/resourcedocsgen/pkg/docs/gen_method.go
```diff
@@ -104,7 +105,7 @@ func (mod *modContext) genMethod(r *schema.Resource, m *schema.Method) methodDoc
 
 	docInfo := dctx.decomposeDocstring(f.Comment, resourceSnippetLanguages)
 	args := methodDocArgs{
-		Title: title(m.Name, ""),
+		Title: strings.Title(m.Name), //nolint:staticcheck // Maintain usage of strings.Title for backwards compatibility.
```
lunaris, line 108: Does the deleted `title` method just call `strings.Title`? Or is there a possible change of behaviour here?

tools/resourcedocsgen/pkg/docs/utils.go
```diff
@@ -158,3 +159,24 @@ type noCopy struct{}
 
 // Lock is a type hint for go vet. It need not be called.
 func (*noCopy) Lock() {}
+
+// Convert a markdown language tag to a [language.Language] if there is a corresponding
+// language. If no language exists, then _, false will be returned.
+func convertMarkdownLanguage(s string) (language.Language, bool) {
```
lunaris, line 165: I feel like I'd stick this in `language` as something like `FromMarkdownLanguage` maybe

tools/resourcedocsgen/pkg/util/language/language.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Language is a valid docsgen language. The default value is not valid.
+type Language struct{ tag string }
+
+var (
+	CSharp     = Language{"csharp"}
+	Go         = Language{"go"}
+	Java       = Language{"java"}
+	Python     = Language{"python"}
+	Typescript = Language{"nodejs"}
```
lunaris, line 35: Nit: I feel like we should just call this constant `NodeJS`/`Nodejs` -- it's fairly ubiquitous at this point and even in this PR the template files call it Node even though the constant is TypeScript.

tools/resourcedocsgen/pkg/util/language/language.go
```diff
@@ -0,0 +1,78 @@
...
+	case Typescript:
+		varName = "Typescript"
+	case YAML:
+		varName = "YAML"
+	default:
+		varName = "Language{}"
+	}
+
+	return "language." + varName
+}
```
lunaris, line 78: Is this used anywhere? Maybe GitHub is hiding a diff from me but I can't see it.

tools/resourcedocsgen/pkg/util/language/set.go
```diff
@@ -0,0 +1,89 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package language
+
+import "iter"
```
lunaris, line 17: I'm not convinced we're getting a whole ton of value out of this over an array of languages/strings. In the use cases where we want all the languages, I think an array is fine and means we don't need `Iter()` etc. For the single use of `Has` we have that I think replaces `StringSet`, it looks like we could just write `if foo == "aws-native"` instead? While we might need flexibility later on, it doesn't seem like we do now/have for a while, and changing that lets us delete two whole files and avoid writing a ton of functionality. Even if we did want the set, I feel like we might still want to consider a more literal representation, unless we have good reasons for the bitset representation that I'm possibly missing?

tools/resourcedocsgen/pkg/util/language/language.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// When-ever a new language is added above, it needs to be added here.
+var langOrder = [...]Language{
+	CSharp,
+	Go,
+	Typescript,
+	Python,
+	YAML,
+	Java,
+}
```
lunaris, line 49: Another thing it feels like we could kill (and in the process simplify things should anyone ever need to add a new language) if we get rid of `Set`.

tools/resourcedocsgen/pkg/util/language/language.go
```diff
@@ -0,0 +1,78 @@
...
+// Package language exposes an enumeration of docs generation target languages:
+//
+// - [CSharp]
+// - [Go]
+// - [Java]
+// - [Python]
+// - [Typescript]
+// - [YAML]
+//
+// For a list of all languages, use [ListAll].
```
lunaris, line 24: Probably needs a tweak since it looks like we ended up calling it `All`.

tools/resourcedocsgen/pkg/util/language/language.go
iwahbe: I'm fine changing the identified to `NodeJS`. Language lists (depending on where they were) used both (`nodejs` and `typescript`).

tools/resourcedocsgen/pkg/util/language/language.go
iwahbe: `GoString` is the function invoked by the `%#v` symbol with `fmt.*` functions. It's used in some `panic`s, but never invoked during normal runtime. It's nice to have when debugging.

It allows `fmt.Printf("Language: %#v\n", language.YAML)` to print `language.YAML` instead of `language.Language{tag: "yaml"}`.

tools/resourcedocsgen/pkg/util/language/language.go
iwahbe: We won't really be able to kill this, since we need to have *a* deterministic iteration order for languages.

tools/resourcedocsgen/pkg/util/language/set.go
iwahbe: I think that `Set` has better copy semantics then `map[Language]struct{}`, which is why I chose it. I personally prefer that `All()` doesn't need to allocate, but its a needless performance optimization.

Given that `language.Set` always contains every language, I can rip it out. I'd rather have something like `language.Iter()` then `language.List()`, since it provides a cleaner `for lang := range language.Iter()` (no extra `_, `).

It was more important pre-refactor, when lots of functions were passing around lists of languages where they needed to apply.

tools/resourcedocsgen/pkg/docs/utils.go
iwahbe: I'd rather keep `package language` concerned only with defining languages (a tightly constrained and simple task), and keep `convertMarkdownLanguage` in `package docs`. How to convert from markdown tags to a Pulumi language isn't 100% obvious. 

I won't change the behavior of `convertMarkdownLanguage` since this is a refactor PR, but it seems very plausible that `convertMarkdownLanguage("javascript")` should return `(language.NodeJS, true)`. It doesn't now.

tools/resourcedocsgen/pkg/util/language/language.go
```diff
@@ -0,0 +1,92 @@
...
+	Python = Language{"python"}
+	NodeJS = Language{"nodejs"}
+	YAML   = Language{"yaml"}
+)
+
+// All providers a deterministic iteration of all valid languages.
+func All() iter.Seq[Language] { return all }
+
+// all is the implementation of [All].
+func all(yield func(Language) bool) {
```
flostadler, line 45: IMO a lazy iterator is a bit overkill here and makes the code in other parts a bit more complex. E.g. the hard-coding of the number of languages.

Feel free to disregard as it's just a nit, but I feel like just exposing the  langOrder array directly would be fine.

tools/resourcedocsgen/pkg/docs/gen.go
```diff
@@ -482,11 +467,15 @@ func (mod *modContext) withDocGenContext(dctx *Context) *modContext {
 // false or `overlaySupportedLanguages` is empty/nil, it returns the default list of supported languages. Otherwise, it
 // filters the `overlaySupportedLanguages` to ensure that they are a subset of the default supported languages.
 func (dctx *Context) getSupportedLanguages(isOverlay bool, overlaySupportedLanguages []string) []string {
+	contextSupportedLanguages := make([]string, 0, 6 /* the number of supported languages */)
```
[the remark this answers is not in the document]
flostadler: Ah, just saw it further down, `language.All()` is an iterator

flostadler (approved):
LGTM

PR #6506, opened 2025-01-24
Refactor: simplify render params

iwahbe:
Stacked on top of https://github.com/pulumi/registry/pull/6478.

---

We currently have 4 *mostly* identical ways to render `[]formalParam` into a string. This PR factors these implementations into a single function:

https://github.com/pulumi/registry/blob/9a42e211f0388791fb4b812decd99eaef88c5d39/tools/resourcedocsgen/pkg/docs/gen_method.go#L322

This improves clarity and reduces LOC.


iwahbe:
@lunaris I'm tagging you here since you suggested this in an earlier PR.

flostadler (approved):
LGTM

PR #6529, opened 2025-01-27
Refactor: Move module names into `*modContext`

iwahbe:
This commit removes a dependency from `*modContext` to it's embedded `*Context`. It also makes the logic slightly simpler.

It is a pure refactor. There is no change in behavior.


flostadler (approved):
🚀

Issue #6593, opened 2025-01-28
Document Output-based Functions in Java


meagancojocar:
### Problem description
On https://www.pulumi.com/registry/packages/aws/api-docs/ec2/getvpc/

we have: 
```
public static CompletableFuture<GetVpcResult> getVpc(GetVpcArgs args, InvokeOptions options)
// Output-based functions aren't available in Java yet
```

But they are available! need to re-gen the docs

mjeffryes:
Look like we need to edit the template probably?

https://github.com/pulumi/registry/blob/39d3df5acf0e8a8e3da30f948391ecf898e116b4/tools/resourcedocsgen/pkg/docs/templates/function.tmpl#L89

thoward:
Updated template.

PR #6596, opened 2025-02-05
Adds a signature for Java Output-based functions to the function template.

thoward:
Fixes https://github.com/pulumi/registry/issues/6593 and https://github.com/pulumi/pulumi-java/issues/1264

(see issues for context)

Output for example referenced in https://github.com/pulumi/registry/issues/6593

image

Output for example referenced in https://github.com/pulumi/pulumi-java/issues/1264

image


sean1588 (approved):
Thanks @thoward !! This looks amazing!!!

PR #6625, opened 2025-02-07
Add file targets for `bin/resourcedocsgen` and docs content

iwahbe:
This change simplifies the process for building docs for a provider to a single command:

```sh
make content/registry/packages/<package_name>
```

Because we are using proper file targets with make, this means we get efficient rebuilds.

This introduces a pinned dependency on `iwahbe/helpmakego`. No developer action is needed to consume the new makefiles. The tool is downloaded and installed locally on first use.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,20 +36,36 @@ build: build-assets
 
 HUGO_SERVE := hugo serve --buildDrafts --buildFuture --ignoreVendorPaths="github.com/pulumi/registry/**/*" | grep -v -e 'WARN .* REF_NOT_FOUND'
 
-# bin/resourcedocsgen is considered to depend on all files in the tools/resourcedocsgen directory.
+HELPMAKEGO_VERSION := v0.1.0
+HELPMAKEGO := bin/${HELPMAKEGO_VERSION}/helpmakego
+
+# Ensure that `helpmakego` is installed at ${HELPMAKEGO} before it is used to resolve targets.
 #
-# Caveat: This does not account for `replace` directives in `go.mod`.
-bin/resourcedocsgen: $(shell find tools/resourcedocsgen | sed 's@ @\\ @g')
-	mkdir -p bin
+# This has the side effect of ensuring that the `bin` directory is present.
+_ := $(shell if ! [ -x ${HELPMAKEGO} ]; then \
+	GOBIN=$(shell pwd)/bin/${HELPMAKEGO_VERSION} go install github.com/iwahbe/helpmakego@${HELPMAKEGO_VERSION}; \
+	fi \
+)
+
+bin/resourcedocsgen: $(shell ${HELPMAKEGO} tools/resourcedocsgen)
 	go build -C tools/resourcedocsgen -o ../../bin ./...
+
+bin/mktutorial: $(shell ${HELPMAKEGO} tools/mktutorial)
+	go build -C tools/mktutorial -o ../../bin ./...
 
 # Generate the API docs for `content/registry/packages/<pkg>`, where <pkg> is the name of
 # the package to be handed off to resourcedocsgen.
-content/registry/packages/%: bin/resourcedocsgen
+.SECONDEXPANSION: # Needed for content/registry/packages/% to have dependencies that reference %.
+api-docs/%: .make/content/registry/packages/$$*/api-docs
+.make/content/registry/packages/%/api-docs: bin/resourcedocsgen \
+				themes/default/data/registry/packages/%.yaml \
+				$$(wildcard themes/default/content/registry/packages/%/*)
 	bin/resourcedocsgen docs registry \
 		--baseDocsOutDir ./content/registry/packages \
 		--basePackageTreeJSONOutDir ./static/registry/packages/navs \
 		$*
+	@mkdir -p "$(@D)"
+	@touch $@
 
 # Run hugo locally, ignoring REF_NOT_FOUND warnings
 .PHONY: serve
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ cypress/screenshots
 tools/resourcedocsgen/resourcedocsgen
 ctrf/
 bin
+.make
--- a/README.md
+++ b/README.md
@@ -83,20 +83,15 @@ Optionally, use `make serve-all` to enable hot reloading of both the pages and C
 
 #### Generating API docs for packages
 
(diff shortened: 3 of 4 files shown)


Makefile
```diff
@@ -36,16 +36,27 @@ build: build-assets
 
 HUGO_SERVE := hugo serve --buildDrafts --buildFuture --ignoreVendorPaths="github.com/pulumi/registry/**/*" | grep -v -e 'WARN .* REF_NOT_FOUND'
 
-# bin/resourcedocsgen is considered to depend on all files in the tools/resourcedocsgen directory.
-#
-# Caveat: This does not account for `replace` directives in `go.mod`.
-bin/resourcedocsgen: $(shell find tools/resourcedocsgen | sed 's@ @\\ @g')
+HELPMAKEGO_VERSION := v0.1.0
+HELPMAKEGO := bin/${HELPMAKEGO_VERSION}/helpmakego
+${HELPMAKEGO}:
```
danielrbradley, line 41: I think this should depend on `bin` too

Makefile
```diff
@@ -36,16 +36,27 @@ build: build-assets
...
+	GOBIN=$(shell pwd)/bin/${HELPMAKEGO_VERSION} go install github.com/iwahbe/helpmakego@${HELPMAKEGO_VERSION}
+
+bin:
 	mkdir -p bin
+
+bin/resourcedocsgen: ${HELPMAKEGO} $(shell ${HELPMAKEGO} tools/resourcedocsgen) | bin
```
danielrbradley, line 47: Not a big fan of order-only prereqs .. they make things look kinda complex. The slightly more hacky immediate variable approach we use in providers is a little easier as you don't then need to remember to add `| bin` to all your binary builds:

```
_ := $(shell mkdir -p bin)
```

Makefile
```diff
@@ -36,16 +36,27 @@ build: build-assets
...
+	go build -C tools/mktutorial -o ../../bin ./...
+
 # Generate the API docs for `content/registry/packages/<pkg>`, where <pkg> is the name of
 # the package to be handed off to resourcedocsgen.
-content/registry/packages/%: bin/resourcedocsgen
+.SECONDEXPANSION: # Needed for content/registry/packages/% to have dependencies that reference %.
+content/registry/packages/%: bin/resourcedocsgen \
+				themes/default/data/registry/packages/$$*.yaml \
+				themes/default/content/registry/packages/$$*/installation-configuration.md \
+				themes/default/content/registry/packages/$$*/_index.md
```
danielrbradley, line 59: Could we re-write this more cleanly using `find`? Also, I don't _think_ this is supported on Mac's version of make.

iwahbe: > Could we re-write this more cleanly using `find`?

No. `$$*` references the `%` in `content/registry/packages/%`, it's not a glob expansion. `content/registry/packages/%` depends on the `bin/resourcedocsgen` and 3 other files with known paths. We don't need to use `find`.

> Also, I don't _think_ this is supported on Mac's version of make.

The version of `make` that ships with my mac supports `.SECONDEXPANSION`.

danielrbradley: 1. Use `basename` to get the last component of the current target.
2. Interpolate the basename inside the call to `find`

Makefile
iwahbe: It doesn't need to. `go` will create directories as needed.

Makefile
iwahbe: It's also worth doing this for `helpmakego`.

Makefile
iwahbe: Can you show what the `find` call would look like? I really don't think that we want to use `find` here.

danielrbradley: Just tested this locally and the secondary expansion works fine on mac.

A few other thoughts here ...
1. Instead of using `$$*`, I think it's clearer to use the simpler `%` form.
2. Instead of listing the specific files within the content directory, we could use a wildcard to be a bit more robust:

    ```
    $$(wildcard themes/default/content/registry/packages/%/*)
    ```

Given we're outputting multiple files and it's not clear which will be written last, then we _should_ use a sentinel file that's only touched after all files are generated. If an error occurred half way then the target would still be marked as up-to-date. 

This somewhat complicates things as you generally want a phony alias for usability, but that doesn't work for wildcards .. so then you'd need to use variables to indicate the provider to build .. which actually might make all of this less magic and a little easier to follow, though leaves questions around how to generate all of them at once while maintaining the up-to-date targets design.

It's hard to make makefiles simple :/

iwahbe: - I didn't know you could use `%` in dependencies, I'll definitely make that change. 
- I can use wildcard here. using `$$(wildcard ...)` adds that back in.
- Let's use the directory as the "sentinel" file. That way life stays simple:
    ```make
    .SECONDEXPANSION: # Needed for content/registry/packages/% to have dependencies that reference %.
    content/registry/packages/%/api-docs: bin/resourcedocsgen \
    				themes/default/data/registry/packages/%.yaml \
    				$$(wildcard themes/default/content/registry/packages/%/*)
    	bin/resourcedocsgen docs registry \
    		--baseDocsOutDir ./content/registry/packages \
    		--basePackageTreeJSONOutDir ./static/registry/packages/navs \
    		$*
    	@touch $@
  ```

This works well testing locally. We can't do a perfect job since Make doesn't think in directories, but this gets us very close. AFAIK this works correctly as long as you don't manually update the contents of `content/registry/packages/<pkg-name>/api-docs`.

danielrbradley: Oh no ... don't do the directory .. that's worse that the file as that always has to be created before any files are written, so it'll be even more likely to result in a failed target being marked as up-to-date!

What about a sentinel file in the folder ...

```make
.SECONDEXPANSION: # Needed for content/registry/packages/% to have dependencies that reference %.
content/registry/packages/%/api-docs/.make: bin/resourcedocsgen \
				themes/default/data/registry/packages/%.yaml \
				$$(wildcard themes/default/content/registry/packages/%/*)
	bin/resourcedocsgen docs registry \
		--baseDocsOutDir ./content/registry/packages \
		--basePackageTreeJSONOutDir ./static/registry/packages/navs \
		$*
	@touch $@
```

iwahbe: The folder works great for all but the first attempt, since it's timestamp is never updated by default.

I moved to a phony target: `.make/content/registry/packages/%/api-docs`. This way we can `.gitignore` a single directory and there is no change of the `.make` files winding up in a hugo build.

danielrbradley (approved):

PR #6640, opened 2025-02-10
Fix FunctionArgsOutputVersion for Java

### LINKED ISSUE #6595: Possibly wrong type name suffix for Java Output-based function documentation generation
In [this function](https://github.com/pulumi/registry/blob/39d3df5acf0e8a8e3da30f948391ecf898e116b4/tools/resourcedocsgen/pkg/docs/gen_function.go#L249C21-L249C31), if you call the "FunctionArgsOutputVersion" from the template, you'll get a typename called "...InvokeArgs" which is similar to the C# function signature, but incorrect for Java, which uses the same type for both the Promise-wrapped function signature and the Output-wrapped function signature.

flostadler:
## Description

Java uses the same args type for both the regular and output versions of the function. This corrects docsgen to set the correct `FunctionArgsOutputVersion`. Additionally it fixes java emitting broken output versions when no output version should be generated.

Note: The actual docs shouldn't change because we've worked around this issue previously by using the regular `FunctionArgs` even for the output version in Java.

Fixes #6595

### DIFF
--- a/tools/resourcedocsgen/pkg/docs/templates/function.tmpl
+++ b/tools/resourcedocsgen/pkg/docs/templates/function.tmpl
@@ -86,7 +86,7 @@ func </span>{{ index .FunctionName lGo }}Output<span class="p">(</span>{{ htmlSa
 <div>
 <pulumi-choosable type="language" values="java">
 <div class="highlight"><pre class="chroma"><code class="language-java" data-lang="java"><span class="k">public static CompletableFuture<{{ template "linkify_param" (index .FunctionResult lJava) }}> </span>{{ index .FunctionName lJava }}<span class="p">(</span>{{ htmlSafe (index .FunctionArgs lJava) }}<span class="p">)</span>
-<span class="k">public static Output<{{ template "linkify_param" (index .FunctionResult lJava) }}> </span>{{ index .FunctionName lJava }}<span class="p">(</span>{{ htmlSafe (index .FunctionArgsOutputVersion lJava) }}<span class="p">)</span>
+{{ if (index .HasOutputVersion lJava) }}<span class="k">public static Output<{{ template "linkify_param" (index .FunctionResult lJava) }}> </span>{{ index .FunctionName lJava }}<span class="p">(</span>{{ htmlSafe (index .FunctionArgsOutputVersion lJava) }}<span class="p">)</span>{{- end }}
 </code></pre></div>
 </pulumi-choosable>
 </div>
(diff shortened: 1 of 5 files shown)


danielrbradley (approved):

flostadler:
Something seems to be wrong with the template, looking why it's broken now

tools/resourcedocsgen/pkg/docs/templates/function.tmpl
```diff
@@ -85,8 +85,8 @@ func </span>{{ index .FunctionName lGo }}Output<span class="p">(</span>{{ htmlSa
...
-<div class="highlight"><pre class="chroma"><code class="language-java" data-lang="java"><span class="k">public static CompletableFuture<{{ template "linkify_param" (index .FunctionResult lJava) }}> </span>{{ index .FunctionName lJava }}<span class="p">(</span>{{ htmlSafe (index .FunctionArgs lJava) }}<span class="p">)</span>
-<span class="k">public static Output<{{ template "linkify_param" (index .FunctionResult lJava) }}> </span>{{ index .FunctionName lJava }}<span class="p">(</span>{{ htmlSafe (index .FunctionArgs lJava) }}<span class="p">)</span>
+<div class="highlight"><pre class="chroma"><code class="language-java" data-lang="java"><span class="k">public static CompletableFuture<{{ template "linkify_param" (index .FunctionResult lJava) }}> </span>{{ index .FunctionName lJava }}<span class="p">(</span>{{ htmlSafe (index .FunctionArgs lJava) }}<span class="p">)</span>{{ if (index .HasOutputVersion lJava) }}
+<span class="k">public static Output<{{ template "linkify_param" (index .FunctionResult lJava) }}> </span>{{ index .FunctionName lJava }}<span class="p">(</span>{{ htmlSafe (index .FunctionArgsOutputVersion lJava) }}<span class="p">)</span>{{ end }}
```
iwahbe, line 89: Do we believe that `.FunctionArgs` is always the same as `.FunctionArgsOutputVersion`, and that's why the docs won't change?

Can we construct a test case where they are different to put the `.FunctionArgs` -> `.FunctionArgsOutputVersion` change under test?

flostadler: Yes, both `FunctionArgs` and `FunctionArgsOutputVersion` are only written to in a single place by using the ret value of `mod.genFunctionArgs(f, funcNameMap, true/false /*outputVersion*/)`. 

The existing tests (e.g. `TestGeneratePackage`) already verify that we're generating the same output and do not regress. 

What would we gain here with a synthetic test that simulates a situation that cannot happen in practice?

iwahbe: If we are sure that they are always the same, we can omit the test.

iwahbe (approved):

PR #6708, opened 2025-02-19
Ignore `ParameterNotFound` error when cleaning up SSM parameters

flostadler:
## Description

The bucket cleanup started failing after https://github.com/pulumi/registry/pull/6656 landed because AWS SSM returns an error when trying to delete a non-existent parameter (sad). It seems that some of the parameters got manually cleaned up.
This change ignores the 404s and treats them as a successful operations.

Fixes https://github.com/pulumi/registry/issues/6680


iwahbe (approved):
Looks good.

sean1588:
thanks @flostadler!!

PR #6720, opened 2025-02-20
Update the instructions for adding a new package to the registry

iwahbe:
The new instructions render like this in "Write" mode, but don't get rendered in the display mode (which we don't want 95% of the time):

Screenshot 2025-02-20 at 10 43 33 PM

---

I have also added new instructions for maintainers, which should be more actionable and clearer. I'm always happy to have suggestions here.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,28 +1,39 @@
-## Description
+<!--
 
-<!-- A brief description of the PR here. -->
+        +--------------------------------------------------------+
+        | If you are adding a new package to the Pulumi Registry |
+        +--------------------------------------------------------+
 
-## Adding a new package?
+        Please make sure that you have:
 
-If this pull request adds a new package, please verify/document:
+        - [ ] Released your package with a version prefixed with `v` (e.g. `v0.1.0`). The
+              part after the leading `v` must be valid semver 2.0.
 
-- [ ] The primary maintainer contact for this package (please indicate if the provider is sponsored by a company or community maintained)
-- [ ] The package's schema URL in this PR is correct.
-- [ ] The package metadata file, if present, contains:
-  - [ ] a supported category (one of `Cloud`, `Infrastructure`, `Network`, `Database`, `Monitoring`, or `Utility`).
-  - [ ] a valid plugin download URL. See [Publish Your Package](https://www.pulumi.com/docs/using-pulumi/pulumi-packages/how-to-author/#publish-your-package).
-  - [ ] a description that explains what the package does.
-  - [ ] a valid logo URL that points to a PNG whose dimensions conform to the others in this repo (e.g., 100x100).
-- [ ] The package repo contains an Overview doc (`/docs/_index.md`) that includes:
-  - [ ] a brief explanation of what the package is and what it does.
-  - [ ] at least one representative example in all supported languages (i.e. TypeScript, Python, Go and C#).
-  - [ ] a front-matter property for the `layout` set to `package`.
-- [ ] The package repo contains an Installation and Configuration doc (`/docs/installation-configuration.md`) that includes:
-  - [ ] links to SDKs in all supported languages (i.e. should link to published SDKs in NPM, PyPi and NuGet).
-  - [ ] a copyable command for installing the resource plugin if necessary.
-  - [ ] an example of configuring the provider with `pulumi config set`.
-  - [ ] an example of configuring the provider with environment variables.
-- [ ] The repository has:
-  - [ ] a version tag prefixed with `v` that corresponds with a valid GitHub release and published package SDKs (in NPM, PyPi, and NuGet)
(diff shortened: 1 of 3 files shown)


iwahbe:
> > The new instructions render like this in "Write" mode, but don't get rendered in the display mode (which we don't want 95% of the time)
> 
> Not sure I follow why we wouldn't want them rendered in display mode most of the time. I think there is some value having it render when displayed since it can also can serve as a check for the contributors themselves to reference to make sure they have met those standards and have all the required things we check for when opening the PR.

I don't want them rendered most of the time for 2 reasons:
1. Because they are not relevant to most PRs, and so don't need to be seen by PR reviewers.
2. Because I want to force maintainers to add the checklist as a comment on the PR manually, then work through it. I don't want the PR author to be able to check things off the PR.

☝️ is related to another constraint: I want to make the check-list more self-explanatory, and thus longer, and needing to fit that into the PR template is becoming painful.

.github/pull_request_template.md
```diff
@@ -1,28 +1,27 @@
...
-- [ ] A member of the @pulumi/docs team has reviewed all documentation.
+<!--
+
+        +--------------------------------------------------------+
+        | If you are adding a new package to the Pulumi Registry |
+        +--------------------------------------------------------+
+
+        Please make sure that you have:
+
+        - [ ] Released your package with a version prefixed with `v` (e.g. `v0.1.0`)
```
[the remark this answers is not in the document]
iwahbe: We do. This check is part of writing metadata into the registry:

https://github.com/pulumi/registry/blob/93c005c243e2445f943a4608917b24ab6f27c7ae/tools/resourcedocsgen/cmd/metadata.go#L319-L321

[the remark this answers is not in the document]
flostadler: Then I think we should explicitly mention that here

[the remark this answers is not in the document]
iwahbe: I agree. The PR has been updated.

docs/adding-a-new-package.md
```diff
@@ -0,0 +1,54 @@
...
+  If the provider is maintained by a company, Pulumi needs a contact person at the maintaining company.
+
+- [ ] The package will generate accurate documentation:
+
+  1. Check out the PR under review and run:
+
+  ```sh
+  $ make bin/resourcedocsgen
+  $ ./bin/resourcedocsgen metadata from-github \
+          --repo-slug '<repoSlug>' \
```
flostadler, line 20:
```suggestion
          --repoSlug '<repoSlug>' \
```

docs/adding-a-new-package.md
```diff
@@ -0,0 +1,54 @@
...
+
+  - [ ] The registry renders a valid logo for the new provider.
+
+- [ ] Hand-written docs are complete and accurate:
+
+  - [ ] Validate that you can install the new provider with the instructions found in `/docs/installation-configuration.md`.
+
+  Maintainers should run the `pulumi plugin install resource <name> <version> --server <pluginDownloadURL>` command specified in the `/docs/installation-configuration.md` and see a provider be downloaded.
+
+  - [ ] `/docs/installation-configuration.md` contains a link to the published SDK in each language (i.e. TypeScript, Python, Go and C#).
```
flostadler, line 45: What about Java? I think now that it's GA we should also add it to the list.

flostadler: Publishing java is hard enough that we do not require users to do that

.github/pull_request_template.md
```diff
@@ -1,28 +1,28 @@
...
-- [ ] A member of the @pulumi/docs team has reviewed all documentation.
+<!--
+
+        +--------------------------------------------------------+
+        | If you are adding a new package to the Pulumi Registry |
+        +--------------------------------------------------------+
+
+        Please make sure that you have:
+
+        - [ ] Released your package with a version prefixed with `v` (e.g. `v0.1.0`). The
```
flostadler, line 9: Shouldn't we also call out the other requirements? I.e. that they've:
- published SDKs for Node.js, Python, Go, C#
- link to the SDKs in their installation-configuration
- show an example of how to use the provider in the above mentioned languages

flostadler (commented):
I've used the maintainer checklist for a provider. Was really helpful!
But I think we should not remove some of the requirements for publishing from the PR template. Otherwise users don't know all the things they need to provider/do

.github/pull_request_template.md
iwahbe: Done.

flostadler (approved):
LGTM!

docs/adding-a-new-package.md
ringods: For Pulumiverse, I would like to publish Java packages, but I'm blocked on this:

https://github.com/pulumi/pulumi-java/issues/1359

PR #6793, opened 2025-03-05
Set `staging: false` to work around pulumi/pulumi-aws#5229

iwahbe:
This is a pre-requisite step to upgrading the AWS provider to a modern version. This change should be a no-op since the default value of `staging` is already false.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -213,6 +213,7 @@ const baseCacheBehavior = {
 // https://docs.aws.amazon.com/general/latest/gr/aws_service_limits.html#limits_cloudfront
 const distributionArgs: aws.cloudfront.DistributionArgs = {
     enabled: true,
+    staging: false, // Attempt to work around https://github.com/pulumi/pulumi-aws/issues/5229
 
     // We only specify one origin for this distribution: the S3 content bucket.
     origins: [


infrastructure/index.ts
```diff
@@ -213,6 +213,7 @@ const baseCacheBehavior = {
 // https://docs.aws.amazon.com/general/latest/gr/aws_service_limits.html#limits_cloudfront
 const distributionArgs: aws.cloudfront.DistributionArgs = {
     enabled: true,
+    staging: false, // Attempt to work around https://github.com/pulumi/pulumi-aws/issues/5229
```
danielrbradley, line 216: Could you elaborate a little on what's not currently working and why this fixes it?

iwahbe: We are on AWS v5, and would like to be on v6 (latest and greatest, etc.). Just upgrading (attempted in https://github.com/pulumi/registry/pull/6730) tries to replace the distribution due to https://github.com/pulumi/pulumi-aws/issues/5229. Setting `staging: false` proactively is the suggested work-around.

danielrbradley: Ah gotya - so this property is added, and is required in v6, so we'll add it now and it shouldn't change anything in v5 .. but lets us also move to v6.

danielrbradley (approved):

infrastructure/index.ts
iwahbe: Hopefully.

PR #7057, opened 2025-04-07
Enable registry to source metadata from pulumi registry APIs

sean1588:
## Enable registry to source metadata from registry APIs. 

Part of: https://github.com/pulumi/pulumi-service/issues/26019

Enables resourcedocsgen to source metadata from package endpoints. Will follow up with pulling schema from endpoints once available.

This PR enables `resourcedocsgen` to pull package metadata from the registry apis instead of sourcing from the data files in the registry repo.

There is a boolean flag, `--use-api`,  I added to enable building from the APIs. By default it will continue to use the filesystem. This will allow us to rollback easily if needed. We can flip this and test it out once the registry service contains the packages. Currently I am seeing an empty list of packages being returned, so I am assuming this has yet to be done. So far I am "guessing" what would be returned just based on inspecting the code in pulumi service. We can wait to get this in until then.

For right now, we should expect the same end result with these changes as it is still configured to use the metadata files on the file system. See previews for confirmation.

### DIFF
--- a/tools/resourcedocsgen/cmd/docs/registry.go
+++ b/tools/resourcedocsgen/cmd/docs/registry.go
@@ -15,12 +15,12 @@
 package docs
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 	"net/url"
-	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -36,6 +36,7 @@ import (
 	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/pulumi/registry/tools/resourcedocsgen/pkg"
+	"github.com/pulumi/registry/tools/resourcedocsgen/pkg/registry/svc"
 	concpool "github.com/sourcegraph/conc/pool"
 )
 
@@ -132,46 +133,30 @@ func getSchemaFileURL(metadata pkg.PackageMeta) (string, error) {
 	return fmt.Sprintf("https://raw.githubusercontent.com/%s/%s/%s", repoSlug, metadata.Version, schemaFilePath), nil
 }
 
-func getRegistryPackagesPath(repoPath string) string {
-	return filepath.Join(repoPath, "themes", "default", "data", "registry", "packages")
-}
-
-func genResourceDocsForAllRegistryPackages(registryRepoPath, baseDocsOutDir, basePackageTreeJSONOutDir string) error {
-	registryPackagesPath := getRegistryPackagesPath(registryRepoPath)
-	metadataFiles, err := os.ReadDir(registryPackagesPath)
+func genResourceDocsForAllRegistryPackages(
+	ctx context.Context,
+	provider svc.PackageMetadataProvider,
+	baseDocsOutDir, basePackageTreeJSONOutDir string,
+) error {
+	metadataList, err := provider.ListPackageMetadata(ctx)
 	if err != nil {
-		return errors.Wrap(err, "reading the registry packages dir")
+		return errors.Wrap(err, "listing package metadata")
 	}
 
 	pool := concpool.New().WithErrors().WithMaxGoroutines(runtime.NumCPU())
-	for _, f := range metadataFiles {
-		f := f
+	for _, metadata := range metadataList {
 		pool.Go(func() error {
-			glog.Infof("=== starting %s ===\n", f.Name())
-			glog.Infoln("Processing metadata file")
-			metadataFilePath := filepath.Join(registryPackagesPath, f.Name())
-
-			b, err := os.ReadFile(metadataFilePath)
-			if err != nil {
-				return errors.Wrapf(err, "reading the metadata file %s", metadataFilePath)
-			}
-
-			var metadata pkg.PackageMeta
-			if err := yaml.Unmarshal(b, &metadata); err != nil {
-				return errors.Wrapf(err, "unmarshalling the metadata file %s", metadataFilePath)
-			}
-
+			glog.Infof("=== starting %s ===\n", metadata.Name)
 			docsOutDir := filepath.Join(baseDocsOutDir, metadata.Name, "api-docs")
 			err = genResourceDocsForPackageFromRegistryMetadata(metadata, docsOutDir, basePackageTreeJSONOutDir)
 			if err != nil {
-				return errors.Wrapf(err, "generating resource docs using metadata file info %s", f.Name())
+				return errors.Wrapf(err, "generating resource docs using metadata file info %s", metadata.Name)
 			}
 
-			glog.Infof("=== completed %s ===", f.Name())
+			glog.Infof("=== completed %s ===", metadata.Name)
 			return nil
 		})
 	}
-
 	return pool.Wait()
 }
 
@@ -179,6 +164,8 @@ func resourceDocsFromRegistryCmd() *cobra.Command {
 	var baseDocsOutDir string
 	var basePackageTreeJSONOutDir string
 	var registryDir string
+	var useAPI bool
+	var apiURL string
 
 	cmd := &cobra.Command{
 		Use:   "registry [pkgName]",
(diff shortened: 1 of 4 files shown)


flostadler:
@sean1588 Can you sync with @iwahbe on how we plan to publish to the registry service?

While your change adds the necessary bits to pull from the registry service, Ian is working on the publishing side (which will still be driven by `pulumi/registry`). Given that the two are heavily intertwined we should align the implementation as well.

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -39,6 +40,76 @@ import (
...
+
+// RegistryAPIProvider implements PackageMetadataProvider using the Pulumi API
+// to retrieve package metadata.
+type RegistryAPIProvider struct {
+	apiURL string
+}
+
+// PackageMetadata represents the API response structure for package metadata
+// from the Pulumi Registry API.
+// TODO: import this from the pulumi-service if possible.
```
iwahbe, line 66: These will eventually be available in the `"github.com/pulumi/pulumi/sdk/v3/go/common/apitype"` package. I don't think this repo will ever be able to see in the service repo.

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -39,6 +40,76 @@ import (
...
+type PackageListResponse struct {
+	Packages []PackageMetadata `json:"packages"`
+}
+
+// NewFileSystemProvider creates a new FileSystemProvider
+func NewFileSystemProvider(registryDir string) *FileSystemProvider {
+	return &FileSystemProvider{
+		registryDir: registryDir,
+	}
+}
```
iwahbe, line 94: Nit: This should return a `PackageMetadataProvider` and `FileSystemProvider` should be private. Likewise with `NewAPIProvider`.

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -39,6 +40,76 @@ import (
...
+
+// contains checks if a string is in a slice
+func contains(slice []string, item string) bool {
+	for _, s := range slice {
+		if s == item {
+			return true
+		}
+	}
+	return false
+}
```
iwahbe, line 111: This function already exists in the `"slices"` package:https://pkg.go.dev/slices#Contains. We should just use that.

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -136,85 +207,87 @@ func getRegistryPackagesPath(repoPath string) string {
...
 	if err != nil {
-		return errors.Wrap(err, "reading the registry packages dir")
+		return errors.Wrap(err, "listing package metadata")
 	}
 
 	pool := concpool.New().WithErrors().WithMaxGoroutines(runtime.NumCPU())
-	for _, f := range metadataFiles {
-		f := f
+	for _, metadata := range metadataList {
+		metadata := metadata
```
iwahbe, line 221: We havn't needed to do this since go 1.22 (https://tip.golang.org/doc/go1.22#language).
```suggestion
```

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -39,6 +40,76 @@ import (
 	concpool "github.com/sourcegraph/conc/pool"
 )
 
+// PackageMetadataProvider is an interface for providers that retrieve package metadata
+// from either the filesystem directory or the Pulumi Registry API.
+type PackageMetadataProvider interface {
+	// GetPackageMetadata returns metadata for a specific package
+	GetPackageMetadata(pkgName string) (*pkg.PackageMeta, error)
+	// ListPackageMetadata returns metadata for all packages
+	ListPackageMetadata() ([]*pkg.PackageMeta, error)
```
iwahbe, line 49: Unless these can be nil (with a nil error), can we just return the structs directly?
```suggestion
	GetPackageMetadata(pkgName string) (pkg.PackageMeta, error)
	// ListPackageMetadata returns metadata for all packages
	ListPackageMetadata() ([]pkg.PackageMeta, error)
```

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -39,6 +40,76 @@ import (
 	concpool "github.com/sourcegraph/conc/pool"
 )
 
+// PackageMetadataProvider is an interface for providers that retrieve package metadata
+// from either the filesystem directory or the Pulumi Registry API.
+type PackageMetadataProvider interface {
```
iwahbe, line 45: Both `GetPackageMetadata` and `ListPackageMetadata` should take and use a `context.Context` for network operations.

iwahbe, line 45: Let's put `PackageMetadataProvider` and it's implementors into their own file.

I wouldn't be sad if it was their own package.

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -234,6 +307,115 @@ func resourceDocsFromRegistryCmd() *cobra.Command {
...
+
+	if len(response.Packages) == 0 {
+		return nil, errors.Errorf("no package found with name %s", pkgName)
+	}
+
+	if len(response.Packages) > 1 {
+		return nil, errors.Errorf("multiple packages found with name %s", pkgName)
+	}
+
+	return convertAPIPackageToPackageMeta(response.Packages[0])
```
iwahbe, line 391: Nit: I'm a big fan of switch statements here, but it's 100% optional:
```suggestion
	switch len(response.Packages) {
	case 0:
		return nil, errors.Errorf("no package found with name %s", pkgName)
	case 1:
		return convertAPIPackageToPackageMeta(response.Packages[0])
	default:
		return nil, errors.Errorf("multiple packages found with name %s", pkgName)
	}
```

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -234,6 +307,115 @@ func resourceDocsFromRegistryCmd() *cobra.Command {
...
+		return nil, errors.Errorf("unexpected status code %d when fetching package list", resp.StatusCode)
+	}
+
+	var response PackageListResponse
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return nil, errors.Wrap(err, "decoding API response")
+	}
+
+	metadataList := make([]*pkg.PackageMeta, 0, len(response.Packages))
+	for _, apiPkg := range response.Packages {
```
iwahbe, line 412: This is a paginated API, so this might need to call GET more then once.

tools/resourcedocsgen/cmd/docs/registry.go
sean1588: oh cool!! thank you!

tools/resourcedocsgen/cmd/docs/registry.go
sean1588: i really like this pattern!

tools/resourcedocsgen/cmd/docs/registry.go
sean1588: ahh good call! will adjust!

sean1588:
@iwahbe I have addressed your comments.

tools/resourcedocsgen/pkg/registry/svc/metadata.go
```diff
@@ -0,0 +1,256 @@
...
+		Featured:      apiPkg.IsFeatured,
+		Native:        slices.Contains(apiPkg.PackageTypes, "native"),
+		Component:     slices.Contains(apiPkg.PackageTypes, "component"),
+		PackageStatus: pkg.PackageStatus(apiPkg.PackageStatus),
+		SchemaFileURL: apiPkg.SchemaURL,
+		Version:       apiPkg.Version,
+		Title:         apiPkg.Title,
+		UpdatedOn:     apiPkg.CreatedAt.Unix(),
+	}, nil
+}
```
iwahbe, line 115: This function never errors and always returns a non-null pointer. We can just have it return `pkg.PackageMeta`.
```suggestion
func convertAPIPackageToPackageMeta(apiPkg PackageMetadata) pkg.PackageMeta {
	return pkg.PackageMeta{
		Name:          apiPkg.Name,
		Publisher:     apiPkg.Publisher,
		Description:   apiPkg.Description,
		LogoURL:       apiPkg.LogoURL,
		RepoURL:       apiPkg.RepoURL,
		Category:      pkg.PackageCategory(apiPkg.Category),
		Featured:      apiPkg.IsFeatured,
		Native:        slices.Contains(apiPkg.PackageTypes, "native"),
		Component:     slices.Contains(apiPkg.PackageTypes, "component"),
		PackageStatus: pkg.PackageStatus(apiPkg.PackageStatus),
		SchemaFileURL: apiPkg.SchemaURL,
		Version:       apiPkg.Version,
		Title:         apiPkg.Title,
		UpdatedOn:     apiPkg.CreatedAt.Unix(),
	}
}
```

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -132,89 +133,69 @@ func getSchemaFileURL(metadata pkg.PackageMeta) (string, error) {
...
-}
-
-func genResourceDocsForAllRegistryPackages(registryRepoPath, baseDocsOutDir, basePackageTreeJSONOutDir string) error {
-	registryPackagesPath := getRegistryPackagesPath(registryRepoPath)
-	metadataFiles, err := os.ReadDir(registryPackagesPath)
+func genResourceDocsForAllRegistryPackages(
+	provider svc.PackageMetadataProvider,
+	baseDocsOutDir, basePackageTreeJSONOutDir string,
+) error {
+	ctx := context.Background()
```
iwahbe, line 140: Let's plumb the context through here.
```suggestion
func genResourceDocsForAllRegistryPackages(
	ctx conxtext.Context,
	provider svc.PackageMetadataProvider,
	baseDocsOutDir, basePackageTreeJSONOutDir string,
) error {
```

iwahbe (commented):
The code looks good. This will need unit tests to merge.

iwahbe:
We should be able to run a live test of this now. The live registry should have all the providers.

CamSoper:
@sean1588 Closing as stale. Feel free to reopen, but I'm planning on revisiting this myself for the March hackathon.

PR #7064, opened 2025-04-08
Refine contribution process

danielrbradley:
Use issues rather than PRs as the main mechanism for adding packages as this allows requesting a dynamically bridged provider or component to be listed as well as native or statically bridged providers.

This deficiency was found while trying to show a customer how to add their Terraform provider to the registry via dynamic bridging and realising it's not currently possible.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-provider.yml
+++ b/.github/ISSUE_TEMPLATE/new-provider.yml
@@ -5,15 +5,22 @@ labels:
   - needs-triage
   - kind/package
 body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for your interest in adding a new provider to the Pulumi Registry! Please fill out the following information to help us understand your request.
+
+        If you are the maintainer of the provider, please also complete the [Pulumi Registry Onboarding form](https://docs.google.com/forms/d/e/1FAIpQLSe8ohRL1PP00KTXEGPYawjY-PgrbxCH6Z88wBqxEDMa4hw2qw/viewform).
   - type: dropdown
     id: PackageType
     attributes:
       label: Package Type
-      description: Type of the package (e.g., Terraform, Native, Component).
+      description: Type of the package (see https://www.pulumi.com/docs/iac/concepts/packages/).
       options:
-        - Pulumi Provider (bridged or native)
+        - Native Pulumi Provider
         - Terraform/OpenTofu provider
         - Component
+        - Other/Not sure
   - type: input
     id: PackageName
     attributes:
@@ -26,7 +33,15 @@ body:
       label: Repository URL
       description: URL of the repository where the package can be found.
       placeholder: https://github.com/pulumi/pulumi-xyz
-  - type: markdown
+  - type: input
+    id: Connection
     attributes:
+      label: Connection to the Package
+      description: How are you connected to the package? (e.g., maintainer, user, etc.)
+      placeholder: I am a [maintainer/user/etc.] of the package.
+  - type: textarea
+    id: OtherInfo
+    attributes:
+      label: Other Information
       value: |
         Please provide any additional information or context that may be helpful in reviewing this request.
--- a/.github/ISSUE_TEMPLATE/new-provider.yml
+++ b/.github/ISSUE_TEMPLATE/new-provider.yml
@@ -1,0 +1,32 @@
+name: New Provider
+description: Request to add a new provider to the registry.
+type: Task
+labels:
+  - needs-triage
+  - kind/package
+body:
+  - type: dropdown
+    id: PackageType
+    attributes:
+      label: Package Type
+      description: Type of the package (e.g., Terraform, Ansible).
+      options:
+        - Pulumi Provider (bridged or native)
+        - Terraform/OpenTofu provider
+        - Component
+  - type: input
+    id: PackageName
+    attributes:
+      label: Package Name
+      description: Name of the package from the source.
+      placeholder: Your package name
+  - type: input
+    id: RepositoryUrl
+    attributes:
+      label: Repository URL
+      description: URL of the repository where the package can be found.
+      placeholder: https://github.com/pulumi/pulumi-xyz
+  - type: markdown
+    attributes:
+      value: |
+        Please provide any additional information or context that may be helpful in reviewing this request.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,30 +1,35 @@
...
     * `docs/installation-configuration.md`, which should contain links to SDK packages in each language along with instructions for configuring the provider (e.g., required credentials and/or environment variables). This file will render as the Installation & Configuration page for your provider ([example](https://www.pulumi.com/registry/packages/aiven/installation-configuration/)).
 1. Create a release of your provider in GitHub.
-1. Add your package to the [community packages list](./community-packages/package-list.json) via pull request to this repository.
 
-For assistance, please reach out on the [Pulumi community slack](https://slack.pulumi.com/) or get in touch with us via this [contact form](https://pulumi.com/contact/?form=registry).
+### Other Package Types
+
+Terraform/OpenTofu providers which have not yet been bridged can be requested through the ["Any Terraform Provider" wrapper](https://www.pulumi.com/registry/packages/terraform-provider/), though we recommend following the [authoring guide](https://www.pulumi.com/docs/guides/pulumi-packages/how-to-author/) to create published SDKs & refine resource configurations.
```
iwahbe, line 24: I'm not sure what this means:

> Terraform/OpenTofu providers which have not yet been bridged can be requested through the ["Any Terraform Provider" wrapper](https://www.pulumi.com/registry/packages/terraform-provider/),

I'm also not sure that we do recommend Pulumi users manually bridge providers instead of using Any TF Provider, manually bridging vs any TF bridging have different tradeoffs. Specifically, you trade of of time bridging and maintaining for quality of the resulting provider.

iwahbe: > I think we'd recommend the static bridging to anyone who also maintains the upstream provider as it gives them better control and has a better experience with published SDKs. But for users who just want to use a TF provider they don't own, then dynamic makes sense.

I agree with this comment, but that's not what the README says.

.github/ISSUE_TEMPLATE/new-provider.yml
```diff
@@ -0,0 +1,32 @@
...
+type: Task
+labels:
+  - needs-triage
+  - kind/package
+body:
+  - type: dropdown
+    id: PackageType
+    attributes:
+      label: Package Type
+      description: Type of the package (e.g., Terraform, Ansible).
```
iwahbe, line 12:
```suggestion
      description: Type of the package (e.g., Terraform, Native, Component).
```

README.md
iwahbe: Suggested rewording is https://github.com/pulumi/registry/pull/7066.

iwahbe (approved):
If we are going down this route, we might want to remove the instructions for publishing from `.github/pull_request_template.md`, since that instructs users how to open a **PR** to publish a provider.

https://github.com/pulumi/registry/blob/deecc468a044a960a9804b0884f582d9790f179f/.github/pull_request_template.md#L1-L40

flostadler:
If we go down this route of using issues, we should also update the docs here: https://www.pulumi.com/docs/iac/using-pulumi/extending-pulumi/publishing-packages/#publish-the-documentation

README.md
```diff
@@ -1,50 +1,38 @@
...
-   * `themes/default/content/registry/packages/exoscale/_index.md`, as described above. This file *must* be included in the first release, but will only be included in subsequent PRs if the content has changed.
+When a community member requests adding a provider to the registry, a Pulumi staff member should perform the following steps:
 
-   Optionally, the PR may include additional content files like How-To Guides if they are present in the upstream repo.
-
-1. Merge the PR if it looks ok.
-1. In pulumi/docs, a [scheduled task](https://github.com/pulumi/docs/actions/workflows/update-theme.yml) runs hourly and will pick up any changes in this repo, generate files from the provider schema and `data/registry/${PROVIDER}.yaml`, and publish to pulumi.com.
-
-  This scheduled task currently lacks adequate monitoring, and **should be watched to ensure that it runs correctly to completion**. (If it fails, it will block all updates to pulumi.com, including marketing and manually maintained docs pages.)
+1. Add the package to the [community packages list](./community-packages/package-list.json) via pull request to this repository.
```
flostadler, line 26: The numeration isn't correct here (seems like it was also wrong before). At least they render correctly because markdown doesn't really care

danielrbradley: This is deliberate - all numbers are 1 so that diffs aren't bad if one extra point is added later on causing all the numbers to shift.

flostadler: Fair enough!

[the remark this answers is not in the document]
mjeffryes: Besides the technical details we need to add to `./docs/adding-a-new-package.md` for different kinds of providers, we should also add requesting the maintainer fill out [Pulumi Registry - Onboarding](https://docs.google.com/forms/d/e/1FAIpQLSe8ohRL1PP00KTXEGPYawjY-PgrbxCH6Z88wBqxEDMa4hw2qw/viewform)

mjeffryes (approved):
After looking at this more closely, I think this is a reasonable step forward. We need to be careful not to promise that we'll add everything requested, but asking the user to open an issue instead of a PR is much more user friendly. 

My main feedback is that we are currently requiring a maintainer to fill out our partnerships onboarding form for any new provider contributions at the moment, so we should include that in the issue template and in our playbook docs.

.github/ISSUE_TEMPLATE/new-provider.yml
```diff
@@ -0,0 +1,32 @@
...
+body:
+  - type: dropdown
+    id: PackageType
+    attributes:
+      label: Package Type
+      description: Type of the package (e.g., Terraform, Native, Component).
+      options:
+        - Pulumi Provider (bridged or native)
+        - Terraform/OpenTofu provider
+        - Component
```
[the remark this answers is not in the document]
danielrbradley: I'll update this to align with the official docs and add a link to: https://www.pulumi.com/docs/iac/concepts/packages/

.github/ISSUE_TEMPLATE/new-provider.yml
```diff
@@ -0,0 +1,32 @@
+name: New Provider
+description: Request to add a new provider to the registry.
+type: Task
+labels:
+  - needs-triage
+  - kind/package
+body:
```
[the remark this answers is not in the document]
danielrbradley: Sure, will add a description at the top pointing to the form, and a question asking their connection to the package.

README.md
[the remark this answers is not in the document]
danielrbradley: Yup, makes sense. I'll just move it as-is for now, add the link, and we can refine the internal instructions further later.

Issue #6774, opened 2025-03-03
Add visual tests to the registry


iwahbe:
Following up on https://github.com/pulumi/registry/issues/6641, we should add visual based (post-render) tests to the registry to prevent render regressions.

To keep false positives down, these tests should run against a dummy provider.

An ideal scenario would be:

- Generate a full registry locally that contains a couple of frozen/dummy providers.
- Serve them locally with `hugo serve`.
- Test the locally hosted site with cypress.

PR #7254, opened 2025-04-15
Add cypress tests to validate docs pages

sean1588:
fixes: https://github.com/pulumi/registry/issues/6774

Adds a schema to represent a test provider that we can use to validate e2e from schema file to html rendered page is structured and contains all the elements as we expect. 

If we want to go as far as rendering everything from overview page to clicking through navigation and then clicking through all the resources, then we could potentially do that, though not sure if that is necessary as what we really want to test is the pages generated contain everything we expect to guard against regressions. Though if we feel this is necessary we can follow on with that.

TODO:
- add testing for function pages (can possibly follow up with this)

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -24,6 +24,7 @@ jobs:
     - mktutorial
     - lint-markdown
     - lint-scripts
+    - test-provider-api-docs
     - preview
     - test-live-publish
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


sean1588:
@iwahbe - good idea! I reverted that commit locally and was able to repro. I also added an extra resource to the schema, one with the `<!-- Start PulumiCodeChooser -->` fences and one with the `{{% examples %}}` fences as those tend to be handled differently and are still in use depending on the provider. I also wired it into the workflow.

iwahbe (approved):

PR #7322, opened 2025-04-22
Update `github.com/pulumi/pulumi/{pkg,sdk}/v3` to v3.169.0

iwahbe:
This also updated `github.com/pulumi/pulumi-yaml` and `github.com/pulumi/pulumi-java/pkg`.

The change to `tools/resourcedocsgen/pkg/output.go` is required by https://github.com/pulumi/pulumi/pull/17898. It does not change behavior, since `tools.EnsureDir` just wrapped `os.MkdirAll`:

```go
func EnsureDir(dir string) error {
	return os.MkdirAll(dir, 0o700)
}
```

While this update is still effected by https://github.com/pulumi/pulumi/issues/19272, I think we should take it now to unblock incoming docs from bridged providers.


lunaris (approved):

flostadler (approved):

iwahbe:
Currently pinning to https://github.com/pulumi/pulumi/pull/19488, since otherwise tests panic.

PR #7377, opened 2025-04-28
feat: add Ionoscloud provider

rmocanu-ionos:
hello, Pulumi team. This PR adds the Ionoscloud provider package to the pulumi registry. This package allows users to provision Ionoscloud resources that are also available in our terraform provider [here](https://github.com/ionos-cloud/terraform-provider-ionoscloud).

- [x] Released your package with a version prefixed with `v` (e.g. `v0.1.0`). The
      part after the leading `v` must be valid semver 2.0.
- [x] Published an SDK for your provider in:
    - [x] Typescript
    - [x] Python
    - [x] Go
    - [x] C#
     - [ ] Java (optional)
- [x] Have checked in a schema.json that matches the location of the `schemaFile`
      specified in `/community-packages/package-list.json`.
- [x] Have a `/docs/_index.md` and `/docs/installation-configuration.md` filled
      out in your repo.
- [x] `/docs/installation-configuration.md` links to all published SDKs.
- [x] `/docs/index.md` shows an example of using your provider in each language.

PR #7539, opened 2025-05-12
Verify ionos-cloud

flostadler:
Verification for https://github.com/pulumi/registry/pull/7377

Files generated via:
```
./bin/resourcedocsgen metadata from-github \
        --repoSlug 'ionos-cloud/pulumi-ionoscloud' \
        --schemaFile 'provider/cmd/pulumi-resource-ionoscloud/schema.json' \
        --version 'v0.2.3'
```


iwahbe (commented):
> [!IMPORTANT]
> Since this is replacing an existing provider from the opentofu pipeline, we will need to remove this provider from the pipeline when we merge.

PR #7827, opened 2025-06-09
Update pulumi/vercel to pulumiverse/vercel

Msa360:
Change the sdk urls to the correct ones. Also the Java sdk doesn't seem to exists.


flostadler:
@Msa360 Thanks for looking into these documentation issues. The documentation is not managed in the registry though, it is fetched from the provider repositories.
Please create a PR here instead https://github.com/pulumiverse/pulumi-vercel/blob/main/docs/_index.md. Once merged&released it will be automatically pulled into the Pulumi Registry.

PR #8199, opened 2025-07-23
Update aws v7 migration guide

corymhall:

### DIFF
--- a/themes/default/content/registry/packages/aws/how-to-guides/7-0-migration.md
+++ b/themes/default/content/registry/packages/aws/how-to-guides/7-0-migration.md
@@ -48,14 +48,11 @@ import * as aws from "@pulumi/aws";
 
 ## Performing the Upgrade
 
-After you have upgraded the AWS provider library to `v7` you need to run Pulumi with the `--refresh` and `--run-program` CLI flags. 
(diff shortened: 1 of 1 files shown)


themes/default/content/registry/packages/aws/how-to-guides/7-0-migration.md
```diff
@@ -42,6 +46,17 @@ import * as aws from "@pulumi/aws";
...
+## Performing the Upgrade
+
+After you have upgraded the AWS provider library to `v7` you need to run Pulumi with the `--refresh` and `--run-program` CLI flags. 
+
+```console
+$ pulumi up --refresh --run-program
+```
+
+- `--refresh` is necessary because the major version has some schema changes and running `--refresh` makes sure the state is migrated to the new schema so that it doesn't cause a diff to be calculated.
+- `--run-program` is necessary because otherwise `--refresh` will use the old provider version to perform the refresh. `--run-program` allows refresh to use the new provider version to perform the refresh.
```
flostadler, line 58: This is very "technical", I'm wondering if it would be beneficial to leave out the technical details instead and just summarize it that those two flags are required in order to "migrate the state to be compatible with the new major version"

Graham-Pedersen (commented):
Ship it!

PR #8287, opened 2025-08-05
Update _index.md

singhvibhanshu:
This PR addresses a UI issue on the `/registry/packages/github` page where the same code snippet appears twice under the Example Usage and GitHub App Installation sections (see attached screenshots for reference). The duplication seems unintended and may confuse users browsing through the documentation.

Screenshot From 2025-08-05 15-20-53
Screenshot From 2025-08-05 15-21-05


danielrbradley (commented):
Hi @singhvibhanshu thanks for taking the time to clean this up.

Unfortunately these documents are all auto-generated and so we can't fix this up here.

This page is generated from the [pulumi-github](https://github.com/pulumi/pulumi-github) provider, but which is in turn generated from the [tf github provider](https://github.com/integrations/terraform-provider-github).

The specific duplicate example can be seen here: 
https://github.com/integrations/terraform-provider-github/blob/9fceeda3663a1bcb48c01e51da109947be65896c/website/docs/index.html.markdown?plain=1#L19-L54

The reason for the duplicate example was to demonstrate a syntax change in terraform between two versions, which for Pulumi has no impact, so they produce exactly the same example code.

I'll open an issue instead with the team who looks after these generated providers to see if we can automatically deduplicate the examples.

danielrbradley:
This is now opened over here: https://github.com/pulumi/pulumi-terraform-bridge/issues/3146 

So I'll close this PR as we can't make the fix directly in this repository.

PR #8561, opened 2025-09-05
Community managed Dynatrace provider lives in Pulumiverse

ringods:
Community managed Dynatrace provider lives in Pulumiverse for a few months already.


iwahbe (approved):
🎉

PR #8714, opened 2025-09-25
`make test` now runs tests on `./tools/resourcedocsgen`.

pose:
The current `test` target runs `scripts/test.sh`. Unfortunately, that file is not wired up to do anything. This PR makes that script run `go test` on `./tools/resourcedocsgen`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ lint-mktutorial:
 
 .PHONY: test
 test:
-	./scripts/test.sh
+	cd ./tools/resourcedocsgen && go test ./...
 
 .PHONY: build
 build: build-assets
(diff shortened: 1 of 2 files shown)


pose:
This change is part of the following stack:

- #8713
    - #8714 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

iwahbe (commented):
`scripts/test.sh` is only referenced once in the Makefile. Let's just run the tests inline and delete `scripts/test.sh`:

```patch
 .PHONY: test
 test:
-	./scripts/test.sh
+	cd ./tools/resourcedocsgen && go test ./...
```

iwahbe (approved):

PR #8814, opened 2025-10-07
Add missing closing 'chooser' tag to fix docker-build doc rendering

pose:
Fixes pulumi/pulumi-docker-build#614

## Why is this being fixed here and not on `pulimi-docker-build`?

I did a search on that codebase and I couldn't find the `docs/_index.md` file on that repo. I suspect it was added here as a stop gap. Here is the long-term fix: https://github.com/pulumi/pulumi-docker-build/pull/622

### DIFF
--- a/themes/default/content/registry/packages/docker-build/_index.md
+++ b/themes/default/content/registry/packages/docker-build/_index.md
@@ -24,7 +24,7 @@ image, publish it to a remote AWS ECR registry, and use an [inline
 cache](https://docs.docker.com/build/cache/backends/inline/) to speed up
 subsequent builds.
 
-{{< chooser language "typescript,python,csharp,go,yaml,java" / >}}
+{{< chooser language "typescript,python,csharp,go,yaml,java" >}}
 
 {{% choosable language typescript %}}
 


pose:
> Do we know the root-cause of how this broke and why? It looks like the page hasn't changed since it was created in 2024. 

Not really. This issue pre-dates me and I'm fixing the immediate customer pain here.

> Was it always broken and we've only just noticed or has something perhaps changed in how we render these pages?

Again, I'm not sure and I wouldn't really know how to find out that. Do you have some pointers?

danielrbradley:
> > Do we know the root-cause of how this broke and why? It looks like the page hasn't changed since it was created in 2024.
> 
> Not really. This issue pre-dates me and I'm fixing the immediate customer pain here.
> 
> > Was it always broken and we've only just noticed or has something perhaps changed in how we render these pages?
> 
> Again, I'm not sure and I wouldn't really know how to find out that. Do you have some pointers?

It looks like it was rendering correctly on June 21st: https://web.archive.org/web/20250621113831/https://www.pulumi.com/registry/packages/docker-build/

No objection to fixing this page, but I think there's likely lots of other pages broken if there was a regression in how these blocks are rendered.

PR #9155, opened 2025-11-13
Stop using /preview registry URLs

fnune:
These have been stabilized, so removing the /preview bit will allow us to get rid of /preview routes sooner.

### DIFF
--- a/scripts/ci/push-registry.py
+++ b/scripts/ci/push-registry.py
@@ -90,7 +90,7 @@ for filename in os.listdir(yaml_db_path):
         continue
 
 
-    api_url = f"{backend_url}/preview/registry/packages/{source}/{publisher}/{name}/versions/{version.removeprefix("v")}"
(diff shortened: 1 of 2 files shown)


danielrbradley (approved):
How do we test this?

fnune:
@danielrbradley 

The [resourcedocsgen / Test](https://github.com/pulumi/registry/actions/runs/19328508958/job/55285514179?pr=9155) check exercises the Go code changes. I think it's [this test](https://github.com/pulumi/registry/blob/master/tools/resourcedocsgen/cmd/docs/registry_test.go#L41-L63). The [Test Live Registry Publish](https://github.com/pulumi/registry/actions/runs/19328508958/job/55285513941?pr=9155) check runs `scripts/ci/push-registry.py` in --dry-run mode which constructs the new URL but doesn't execute the HTTP request. So we're missing that one, but I'd say the former is enough of a check.

PR #10159, opened 2026-03-03
[registry] Add versioned documentation for blessed packages

fnune:
Add infrastructure for generating versioned API docs using the `pkg@X.x`
URL format (e.g., `/registry/packages/aws@6.x/api-docs/`). Existing URLs
continue serving the latest version.

Since the diff is hard to read because of all the removed stuff, here is a link to a smaller commit with just the code changes: https://github.com/pulumi/registry/pull/10159/changes/14f60f7f73a1c3606aedec54416028d8e2c86ddf

image
image

Template changes:
- Parse versioned package names and look up metadata from package_versions/
- Add version selector dropdown to package cards
- Handle nav JSON lookup for versioned packages

Generation script (`scripts/generate-versioned-docs.sh`):
- Fetches blessed package list from ci-mgmt
- Discovers last 3 major versions via registry-mirror-discover
- Generates docs with resourcedocsgen, extracts metadata from schemas
- Outputs to temp directory to preserve latest nav JSON

The script depends on registry-mirror-tools which will be made public to
support this workflow.

Generated content is excluded from git via .gitignore:
- `themes/default/content/registry/packages/*/api-docs/`
- `themes/default/content/registry/packages/*@*/`
- `themes/default/static/registry/packages/navs/`
- `themes/default/data/registry/package_versions/`

Also removes aws-v6 and azure-native-v2 content which used an older naming
convention that conflicted with module paths. Top-level redirects are added
from old URLs to the new format; deep link redirects would require thousands
of S3 objects and may be addressed separately with CloudFront functions.

## Test plan

- CI preview deployment:
  - `/registry/packages/aws/` serves latest version with working nav
  - `/registry/packages/aws@6.x/` serves v6 docs with version selector
  - Version selector switches between versions correctly
  - `/registry/packages/aws-v6/` redirects to `/registry/packages/aws@6.x/`
  - `/registry/packages/azure-native-v2/` redirects to `/registry/packages/azure-native@2.x/`

### DIFF
--- a/scripts/generate-versioned-docs.sh
+++ b/scripts/generate-versioned-docs.sh
@@ -25,7 +25,7 @@ MINOR_COUNT=1
 
 # Blessed packages are first-party providers that get versioned docs.
 # Fetch dynamically from ci-mgmt (pinned commit for stability):
-CI_MGMT_COMMIT="5ec4f6d0a1fe13e6c7d066b52c31bb74dacc3833"
+CI_MGMT_COMMIT="b3cede4113152996ec67c47a2ca0cef3c5aeb626"
 PROVIDERS_JSON_URL="https://raw.githubusercontent.com/pulumi/ci-mgmt/${CI_MGMT_COMMIT}/provider-ci/providers.json"
 PROVIDERS_JSON=$(curl -sfL "$PROVIDERS_JSON_URL") || { echo "Failed to fetch providers.json from ci-mgmt" >&2; exit 1; }
 mapfile -t BLESSED_PACKAGES < <(echo "$PROVIDERS_JSON" | jq -r '.[]')


fnune:
> The other thing here is before merging, we should probably generate bulk redirect rules mapping every old path to its new @ equivalent. Without these, Google will see thousands of soft 404s, which burns crawl budget and drops those pages from the index over weeks. External links (if any) will also break. 404s are normal / not penalized but would much rather signal a "move" to search engines to not lose out on some of the "SEO equity".

Noting that the only URLs that become 404s here are `aws-v6/*` and `azure-native-v2/*` routes that are not the base `aws-v6` and `azure-native-v2` which hold the READMEs.

For all other packages, there are no broken routes.

pose:
LGTM!

PR #10215, opened 2026-03-06
Add versioned documentation for blessed packages

fnune:
Reapplies #10159 with fix for CloudFront Function compatibility (ES5 syntax).

Changes:
- Versioned API docs using `pkg@X.x` URL format
- CloudFront Function for legacy URL redirects (`aws-v6/*` → `aws@6.x/*`)
- SEO mitigations (canonical tags, sitemap exclusion)
- Version selector in package cards

The CloudFront Function now uses ES5-compatible JavaScript (var, traditional for loops).

Using this test deploy to check for the validity of the redirects. https://github.com/pulumi/registry/actions/runs/22772542771

Also ran some tests on the AWS console validator:
image
image
image

### DIFF
--- a/scripts/ci/make-s3-redirects.sh
+++ b/scripts/ci/make-s3-redirects.sh
@@ -9,17 +9,36 @@ source ./scripts/ci/common.sh
 # response code is returned for search engines and enables better support for URL anchors.
 
 destination_bucket="$(cat "$(origin_bucket_metadata_filepath)" | jq -r ".bucket")"
+region="$(aws_region)"
 
 IFS="|"
 
-# Applies custom redirects supplied in the `scripts/redirects` directory.
-ls -l "./scripts/redirects/" | tail -n +2 | awk '{print $9}' | while read line; do
-    redirect_file="./scripts/redirects/$line"
+# Collect all redirects from all files, then apply them in parallel.
+pids=()
+for redirect_file in ./scripts/redirects/*.txt; do
+    [[ -f "$redirect_file" ]] || continue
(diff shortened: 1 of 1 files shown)


CamSoper:
## Build performance: incremental caching

This branch now includes two layers of caching that skip regenerating unchanged packages:

**Main API docs (`resourcedocsgen`):** Each package gets a `.generated` sentinel keyed on the SHA-256 of its YAML metadata + a hash of the resourcedocsgen source tree (injected at build time via `-ldflags`). On cache hit, schema fetching and doc generation are skipped entirely. The source hash ensures the cache auto-invalidates when doc generation logic or dependencies change.

**Versioned docs (`generate-versioned-docs.sh`):** Same sentinel pattern, keyed on the schema URL. Versioned doc sets are cached in `.cache/versioned-docs/` and restored between runs.

Both caches are stored via `actions/cache` under the `docs-cache-` prefix. Caches are branch-scoped — after merging, the first `master` build populates the cache, and subsequent PR branches fall back to it.

**Results (warm cache):**
- API docs generation: ~8 min → ~24s (296 packages skipped)
- Versioned docs: ~3 min → ~29s (67 doc sets skipped)
- Total build: ~33 min → ~19 min

The remaining ~19 min is Hugo rendering (~7 min) and S3 upload + Pulumi deploy (~9 min), which aren't cacheable without larger architectural changes.

CamSoper:
@fnune I think I got it. I'll pick this up Monday and beat on it a bit more.

PR #10217, opened 2026-03-06
Fix missing fetch-depth in testing-deploy workflow

CamSoper:
## Summary

- The `testing-deploy` workflow was using the default shallow clone (`fetch-depth: 1`), while `registry-mirror-discover` requires full git history to find package versions
- This caused versioned docs generation to silently produce no output, so the version selector never rendered on the testing CloudFront
- Both `pull-request.yml` and `push.yml` already have `fetch-depth: 0` for this reason; this brings `testing-deploy.yml` in line with them

## Test plan

- [ ] Run `testing-deploy.yml` against a branch with versioned docs changes and confirm the version selector renders at the CloudFront URL


CamSoper:
Cherry-picked onto fnune/versioned-docs-v2 for testing. Will be included in #10215.

PR #10240, opened 2026-03-09
Seed S3 buckets from production to speed up sync

CamSoper:
## Summary

- Before uploading the local build, server-side copies existing content from the current production S3 bucket into the new bucket
- The subsequent local `aws s3 sync --delete` then only transfers the delta, which should cut upload time by ~4-5 minutes for unchanged content
- Seeding is entirely best-effort: if the production metadata endpoint is unreachable or the copy fails, it falls back to the existing full-upload path
- Adds timing logs for both the seed step and the local sync for observability

## Test plan

- [ ] Trigger `testing-deploy.yml` against this branch — first run should complete normally (cold build, seeding from production bucket)
- [ ] Verify logs show seed timing and local sync timing
- [ ] Verify the test site renders correctly after a seeded build
- [ ] Compare S3 sync time vs the ~7.5 min baseline

### DIFF
--- a/scripts/ci/common.sh
+++ b/scripts/ci/common.sh
@@ -188,7 +188,7 @@ seed_bucket() {
         local seed_start
         seed_start=$(date +%s)
 
-        if aws s3 sync "s3://${candidate}" "s3://${destination_bucket}" \
+        if aws s3 cp "s3://${candidate}" "s3://${destination_bucket}" --recursive \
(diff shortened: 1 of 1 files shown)


CamSoper:
Seeding new S3 buckets from existing ones turned out to be slower than uploading from local (~560s seed vs ~340s direct upload). The overhead of copying hundreds of thousands of small objects between buckets negates any benefit. Closing.

PR #10412, opened 2026-03-24
feat: add Accept All / Reject All buttons to cookie consent banner

sicarul:
## Summary
- Port consent banner improvements from pulumi/docs#18172
- Add explicit Accept All, Reject All, and Manage cookies buttons to the cookie consent banner
- Defer `analytics.load()` to the consent manager so tracking only starts after user consent
- Hide the close button since users now have explicit accept/reject actions
- Extract consent banner styles into a dedicated `_consent-banner.scss` partial


CamSoper (approved):
LGTM! Clean port from pulumi/docs#18172. The explicit Accept All / Reject All buttons are a solid improvement for GDPR compliance, and extracting the consent banner styles into a dedicated partial is a nice cleanup. All CI checks pass.

PR #10536, opened 2026-04-01
Delist 9 outdated Pulumi-published packages

jkodroff:
Delist packet, confluent, yandex, metabase, azure-justrun, azure-static-website, google-cloud-static-website, str, and tls-self-signed-cert. These packages have not been updated since 2023 or earlier and are no longer actively maintained.

### DIFF
--- a/community-packages/package-list.json
+++ b/community-packages/package-list.json
@@ -65,10 +65,6 @@
       "schemaFile": "provider/cmd/pulumi-resource-astra/schema.json"
     },
     {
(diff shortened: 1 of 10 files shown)


CamSoper (approved):
LGTM! All 9 packages correctly delisted using the established `publisher: DEPRECATED` pattern, and the 5 entries in `package-list.json` are cleanly removed. 👍

borisschlosser (approved):

PR #10628, opened 2026-04-10
perf: gzip cli-docs.json bundles at upload so the CDN serves them compressed

CamSoper:
## Summary

The `cli-docs.json` bundles served at `/registry/packages/<pkg>/api-docs/cli-docs.json` -- consumed by the **not-yet-released** `pulumi docs` command (pulumi/pulumi#22538, pulumi/pulumi#22539) -- were coming back **uncompressed**. For aws the bundle is 103 MB on the wire when it could be ~7 MB gzipped. This PR pre-gzips the bundles at upload time and sets `Content-Encoding: gzip` on the S3 object, which makes both CloudFront layers pass them through verbatim without relying on (or being constrained by) their automatic compression.

No released Pulumi client fetches these URLs today, so this change has **zero blast radius on existing users**.

## Why the CDN wasn't compressing

Responses traverse two CloudFront distributions in series: the outer `www.pulumi.com` CDN (defined in `pulumi/docs/infrastructure/index.ts`) and the inner registry CDN (this repo).

1. **Outer CDN, `/registry/*` behavior:** attaches `originRequestPolicyId: allViewerExceptHostHeaderId` (the managed `AllViewerExceptHostHeader` policy) to forward viewer headers to the inner CDN. That policy forwards `Accept-Encoding`, and **CloudFront silently disables automatic compression on any cache behavior that forwards `Accept-Encoding` to the origin.** So `compress: true` on that behavior is a no-op for all registry traffic.
2. **Inner registry CDN:** uses the S3 website endpoint as a custom origin, which has edge cases that prevent reliable auto-compression even for smaller files. And for the aws bundle it's moot: CloudFront's auto-compression has a hard 10 MB object-size ceiling, and the aws bundle is 103 MB.

A follow-up brief for the `pulumi/docs` repo is queued separately to fix the outer-CDN policy so that HTML package pages and smaller JSON also start compressing on the fly. That's out of scope here.

## The fix

Localized to `scripts/ci/sync.sh`. The `cli-docs-out/` tree contains only `cli-docs.json` files, so:

1. Parallel `gzip -9` over every `cli-docs.json` in place (`xargs -P \"$(nproc)\"`).
2. Existing `s5cmd sync cli-docs-out/` invocation gains `--content-encoding gzip --content-type application/json` globally. s5cmd v2.3.0 (the pinned version in `.github/workflows/*.yml`) supports these flags on `sync`.

### Smoke test

New post-sync assertion against the S3 website endpoint for `random/api-docs/cli-docs.json`: uses `curl --dump-header` + `gunzip -t` in a single round trip to verify both that the response carries `Content-Encoding: gzip` and that the body decompresses cleanly. Fails CI closed if either check fails, catching any regression at the origin layer before either CloudFront can mask it.

## Measured impact (end-to-end through both CloudFront layers on `www.pulumi-test.io`)

| Package | Before (raw) | After (gzip wire) | Ratio |
|---|---:|---:|---:|
| random | 204 KB | 16.7 KB | 12.2x |
| azuread | 2.8 MB | 203 KB | 13.9x |
| **aws** | **103 MB** | **7.38 MB** | **14.0x** |

The aws bundle (bold) was previously **unreachable by any CDN-side fix** (over the 10 MB auto-compress ceiling). The fix cuts its wire size by ~93%. For the `pulumi docs` CLI, cold runs for aws drop from ~13 s to the low single digits. No client-side changes needed -- Go's `http.Client` decompresses `Content-Encoding: gzip` transparently.

### DIFF
--- a/BUILD-AND-DEPLOY.md
+++ b/BUILD-AND-DEPLOY.md
@@ -488,7 +488,7 @@ The `scripts/ci/build.sh` script manages the cache lifecycle:
 2. **Generate**: `make api-docs` runs `resourcedocsgen`, which skips fresh packages and regenerates stale ones.
 3. **Save**: copies the generated output (including updated sentinel files) back to `.cache/api-docs/` for the next run.
 
-CLI docs follow the same lifecycle: on restore, `.cache/api-docs/cli-docs/<pkg>/api-docs/` is copied to `cli-docs-out/registry/packages/<pkg>/api-docs/`; on save, the reverse copy is performed. Only `schema.json` (not the entire directory tree) is cached per package in the schema layer, to avoid persisting stale CLI doc files from older builds. CLI docs are stored uncompressed in the cache; `sync.sh` gzip-compresses them in place immediately before uploading to S3 (with `Content-Encoding: gzip`).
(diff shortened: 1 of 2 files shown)


CamSoper:
@claude Fix the issues you found.

sicarul (approved):

PR #10932, opened 2026-05-07
Add dagsterplus provider package

dehume:
Adds the Dagster+ provider to the Pulumi Registry as a community package.

## Changes
- Added `themes/default/data/registry/packages/dagsterplus.yaml`
- Added `dagster-io/pulumi-dagsterplus` to `community-packages/package-list.json` for nightly version tracking


tehsis:
I've re-pushed this branch to `pulumi/registry` as #10996 so the `Test Live Registry Publish` check can run — that job authenticates to ESC via OIDC, and GitHub does not grant `id-token: write` to `pull_request` runs from forks (regardless of `permissions:` config or who re-runs the job). Your commit was pushed as-is with authorship preserved.

Tracking the merge on #10996; once it lands, this PR will auto-close since the commit SHA matches.

fnune:
Consolidating on #10996, which is the same commit pushed to this repo so the publish check can run. I added the missing publisher entry there. Closing this in favor of that one.

Issue #10814, opened 2026-04-28
Update nav, fonts and colors to new branding


jeffmerrick:
Nav, font and color updates to bring in line with docs and marketing site.

PR #10998, opened 2026-05-13
Use CSS variables for custom font-family declarations

fnune:
Replace hardcoded 'Inter', 'Gilroy', and 'Iosevka' font-family declarations
across SCSS files with the existing --font-body, --font-display, and
--font-mono CSS variables defined in _theme.scss. This restores the system
font fallbacks so the registry remains legible when Iosevka, Inter, or
Gilroy fail to load.

Also fix --font-mono, which previously listed Roboto Mono (not loaded by
the site) instead of Iosevka.

Reference: https://pulumi.slack.com/archives/C85BS3LJZ/p1778525759901189

## Test plan

- `make build-assets` compiles without new warnings
- Disable custom fonts in DevTools and confirm code blocks fall back to
  system monospace, body text to system sans, headings to system sans

### DIFF
--- a/themes/default/theme/src/scss/_breadcrumb.scss
+++ b/themes/default/theme/src/scss/_breadcrumb.scss
@@ -9,7 +9,7 @@ ol.docs-breadcrumb {
         margin: 0;
 
         a {
-            font-family: 'Inter';
+            font-family: var(--font-body);
             font-weight: 500;
             font-size: 12px;
             line-height: 1.5;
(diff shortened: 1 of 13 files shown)


jeffmerrick (approved):
Looks good! We still need to do color and font updates in the registry (https://github.com/pulumi/registry/issues/10814) to bring it in line with the updated branding, this fix will make that easier on us later.

PR #11109, opened 2026-05-21
community-packages: add DefangLabs/pulumi-defang (defang-aws, defang-gcp, defang-azure)

lionello:
## Summary

Adds three new community packages for the Defang Pulumi provider — one per target cloud — at **v2.1.1**:

- `defang-aws` — Deploy containerized services to AWS
- `defang-gcp` — Deploy containerized services to GCP
- `defang-azure` — Deploy containerized services to Azure

All three share the single-source repo [`DefangLabs/pulumi-defang`](https://github.com/DefangLabs/pulumi-defang) and the same `docs/_index.md` + `docs/installation-configuration.md` (the docs cover all three clouds in one page for now; we'll split per-package in a follow-up if useful).

The previous single-schema entry (`defang`) is now obsolete — the provider has been split into three cloud-specific schemas. This PR also marks it `publisher: DEPRECATED` so the publishing scripts skip it, and removes its cached content (replaced by the three new packages).

This PR fixes the issues raised by @borisschlosser in https://github.com/pulumi/registry/pull/10890 

## Checklist

- [x] Released v2.1.1 (valid semver).
- [x] SDKs published in TypeScript ([npm](https://www.npmjs.com/package/@defang-io/pulumi-defang-aws)), Python ([PyPI](https://pypi.org/project/pulumi-defang-aws/)), Go (via repo tag), and .NET ([NuGet](https://www.nuget.org/packages/DefangLabs.DefangAws/)) — for all three clouds. Java not published.
- [x] Three `schema.json` files checked in at the paths referenced by `community-packages/package-list.json`.
- [x] `docs/_index.md` and `docs/installation-configuration.md` present in the source repo; the install doc links to all published SDKs.


fnune (changes requested):
Thanks for getting the schemas published, that clears the original reason defang came off the list in #10890.

### Blocker: only the aws plugin exists

Every release (v2.0.0 through v2.1.1) ships `pulumi-resource-defang-aws-*` archives and zero gcp or azure binaries. So this works today:

```bash
pulumi plugin install resource defang-aws v2.1.1 --server github://api.github.com/DefangLabs/pulumi-defang
```

The same command for `defang-gcp` or `defang-azure` fails on a missing release asset (`pulumi-resource-defang-gcp-v2.1.1-linux-amd64.tar.gz` not found), and no version works since no release has ever included those binaries.

Fix: publish the gcp and azure plugin archives to the release, the same way aws does.

### Documented install command is wrong for all three

The docs use `--server github://api.github.com/DefangLabs`, which Pulumi resolves to `pulumi-defang-<cloud>` repos that do not exist, so it 404s even for aws. It needs the repo on the end:

```bash
--server github://api.github.com/DefangLabs/pulumi-defang
```

The SDKs already embed that correct URL, which is why aws installs through `pulumi up` even though the written command does not.

### License does not match the repo

The three schemas declare `license: Apache-2.0`, but the repo LICENSE is AGPL-3.0 (the README badge shows both). The registry surfaces the schema value, and AGPL versus Apache is a materially different obligation for anyone adopting the provider, so these need to agree. Please confirm which one actually applies and align the schema to it.

### Docs are shared verbatim across the three packages

`_index.md` and `installation-configuration.md` are byte-identical for defang-aws, defang-gcp, and defang-azure, all titled "Defang Provider", and the code samples import the old combined packages (`@defang-io/pulumi-defang`, `pulumi_defang`, `DefangLabs.Defang`, plus a placeholder `example.com/pulumi-defang/...` Go path) rather than the per-cloud ones. The three registry pages render identically and point people at the deprecated package names.

### Minor

The docs list a Go SDK that none of the schemas declare.

fnune:
One more, and this one is on our side rather than yours.

The registry's nightly metadata workflow derives a provider's short name from the repo name, so three packages that share `DefangLabs/pulumi-defang` all resolve to `defang`. Their auto-update PRs would then collide on the same branch and title, so version bumps for the three would step on each other.

We need to fix that workflow before the registry can support multiple packages from one repo, and I am opening a PR for it. Until that lands, the cleanest path is to add `defang-aws` on its own, with gcp and azure following once both their plugin binaries and the workflow fix are in place.

lionello:
@fnune I've updated the branch. Can you have a look?

fnune:
@lionello I continued the work here: https://github.com/pulumi/registry/pull/11614, can you take a look to see if my fixes align with your expectation?

PR #11133, opened 2026-05-22
Fix mobile accordion behavior and registry visual polish

jeffmerrick:
Just some minor tweaks and fixes!

---
**Generated description of changes:**

The mobile `#accordion-package-card` and `#accordion-table-of-contents` were stuck open because `<div class="accordion-content flex">` applied Tailwind's `.flex` utility (in `@layer utilities`), which wins over the `display: none` rule from `docs/_docs-main.scss` — that file is imported inside `@layer components` in `main.scss`. Dropping the `flex` class from both partials lets the SCSS cascade control show/hide again.

## Other changes bundled in

- A–Z packages header stacks vertically on mobile.
- Replace `text-blue-600`/underline links with `text-violet-primary` + hover underline across package-alert, language indexes, and registry page templates.
- Move package-card padding/background onto the `.card` utility; trim inner padding to 16px.
- Docs main-nav active item goes full-width with a larger hit area; drop stray `sidenav-section` margins.
- Selected api-doc nav tree row uses `violet-100` instead of `#e0e0e0`.
- Right-nav ad `p-6` → `p-4`.

### DIFF
--- a/themes/default/layouts/partials/docs/right-nav-ad.html
+++ b/themes/default/layouts/partials/docs/right-nav-ad.html
@@ -1,5 +1,5 @@
 <div class="docs-marketing-ad w-full">
-    <div class="card card-hover w-full block p-6">
+    <div class="card card-hover w-full block p-4">
         <h4 class="no-anchor">
             Try Pulumi Cloud free. <br />Your team will thank you.
         </h4>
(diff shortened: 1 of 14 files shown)


cnunciato (approved):
⭐ 🙌🏻

PR #11165, opened 2026-05-26
Fix nested resource tree nodes navigating instead of toggling

### LINKED ISSUE #11164: Resource tree toggle not working for nested modules
In the resource tree on provider API docs pages, clicking the caret on a nested module that has children navigates to the module's page instead of expanding/collapsing it. Top-level modules expand/collapse correctly.

Repro

1. Open https://www.pulumi.com/registry/packages/pulumiservice/api-docs/api/
2. Expand api in the left-hand resource tree
3. Click the caret next to agents (a nested module with children)

Expected: the agents node expands in place to reveal its children.
Actual: the browser navigates to /registry/packages/pulumiservice/api-docs/api/agents/

rgharris:
`hasChildren` in `getChildNodes` was defined as `length === 0`, the inverse of its name. That bool gets passed through `!hasChildren` to `onExpansionChange`'s `isLink` parameter, so nested nodes with children navigated via `window.location.href` on every click instead of toggling expansion. Flip the comparison to match the variable name and the top-level `render()` logic.

I've tested this locally.

Closes #11164


rgharris:
Going another direction per Slack, closing

PR #11206, opened 2026-05-29
Neutralize Hugo shortcode delimiters in generated code examples

### LINKED ISSUE #11205: Hugo shortcode delimiters in generated code examples break the site build
Noticed via #11191

`resourcedocsgen` emits provider docs as Markdown that Hugo then renders. Hugo parses shortcodes even inside fenced code blocks, so an example snippet containing `{{%` or `{{<` is treated as a shortcode call. The dynatrace@v0.35.2 `HttpMonitor` examples include `{{%s|token}}` (Dynatrace vault syntax + a Java `String.format` placeholder), which Hugo reads as a call to a shortcode named `s` and aborts the whole build:

```
failed to extract shortcode: template for shortcode "s" not found
```

Deterministic, not flaky — new in v0.35.2 (0 occurrences in v0.34.0 → 8), so every build fails until fixed.

rgharris:
Stop Hugo from misreading shortcode delimiters inside generated fenced code examples, which was aborting the entire site build.

resourcedocsgen now neutralizes Hugo shortcode delimiters ({{% and {{<) that appear inside fenced code blocks when emitting .md files. Hugo parses shortcodes even inside code fences, so an example snippet like {{%s|token}} (Dynatrace credential-vault syntax plus a Java String.format placeholder) was read as a call to a shortcode named "s", failing the build with 'template for shortcode "s" not found'.

The rewrite is restricted to fence interiors, so legitimate hand-authored shortcodes ({{< chooser >}} / {{% choosable %}}, which only wrap fences from the outside) are untouched. Inline code spans and 4-space-indented code blocks are intentionally outside this change because the current generator surface emits fenced examples.

Neutralization inserts a single space after {{. It needs no delimiter pairing (the offending text often has no matching %}}) and is idempotent. Tradeoff: the inserted space changes the rendered example, so a copied snippet carries an extra space (Bearer {{ %s|token}}). That is harmless to the Java String.format call, but Dynatrace would not accept the resulting vault reference if pasted verbatim. We accept this cosmetic cost over a failing build.

Surfaced by https://github.com/pulumi/registry/issues/11191 (auto-filed workflow failure for the dynatrace@v0.35.2 bump, PR https://github.com/pulumi/registry/pull/11190). The offending {{%s|token}} examples are new in that version (0 occurrences in v0.34.0, 8 in v0.35.2), so the build fails deterministically until this lands on master.

Closes https://github.com/pulumi/registry/issues/11205

### DIFF
--- a/tools/resourcedocsgen/pkg/output.go
+++ b/tools/resourcedocsgen/pkg/output.go
@@ -17,11 +17,67 @@ package pkg
 import (
 	"os"
 	"path"
+	"regexp"
+	"strings"
 
 	"github.com/pkg/errors"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
+
+// hugoShortcodeDelim matches Hugo's shortcode delimiters `{{%` and `{{<`.
+var hugoShortcodeDelim = regexp.MustCompile(`\{\{([%<])`)
+
+// neutralizeHugoShortcodes disarms Hugo shortcode delimiters inside fenced code
+// blocks, where Hugo would resolve them before Markdown rendering. Inline code
+// spans and indented code blocks are outside the current generator surface.
+func neutralizeHugoShortcodes(contents []byte) []byte {
+	lines := strings.Split(string(contents), "\n")
+	inFence := false
+	var fenceChar byte
+	var fenceLen int
+	for i, line := range lines {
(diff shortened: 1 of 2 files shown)


rgharris:
@claude review again

CamSoper (approved):
This looks like a good fix. Approved!

PR #11244, opened 2026-06-03
Derive package short name from the schema name

fnune:
Spun out of the [#11109](https://github.com/pulumi/registry/pull/11109) review (defang).

## Problem

The nightly `generate-package-metadata.yml` derives a provider short name from the repo slug (`pulumi-<name>`). When one repo publishes multiple packages, they all collapse to the same name. [`DefangLabs/pulumi-defang`](https://github.com/DefangLabs/pulumi-defang) ships `defang-aws`, `defang-gcp`, and `defang-azure`, all resolving to `defang`, which causes:

- the PR branch (`<short_name>/<run_id>-<run_number>`) to be identical for all siblings, so they overwrite each other,
- the PR title (`Publish Package Metadata <short_name>@<version>`) to be identical, and
- the open-PR check (`contains(prs, short_name)`) to skip siblings once any one PR is open.

No repo currently publishes more than one package, so this has not surfaced in production yet.

## Fix

Read the short name from the schema's own `name` field (which is also the registry YAML filename), falling back to the repo-slug regex only when the schema name cannot be read (non-JSON schema or fetch failure).

This is the only reliable source: the schema directory is not one (`pulumi-resource-kafkaconnect` -> name `kafka-connect`), and the repo slug is not one for multi-package repos.

### Scope of change (verified against the full list)

`group2` / `provider_short_name` only controls the nightly PR branch, title, and the open-PR skip check. It does not affect the published package name, which comes from the schema `name` via `metadata from-github` (writes `<name>.yaml`) and `push-registry.py` (publishes `{source}/{publisher}/{name}`). So no package is renamed or republished.

- Every existing package: derived name now equals its schema `name`, i.e. its existing `<name>.yaml`. No mislabels.
- `runpod`: corrects a pre-existing mislabel (repo-derived `runpod-native` never matched the real name `runpod`); the published name was and stays `runpod`.
- `defang-aws` / `defang-gcp` / `defang-azure`: now get distinct PRs instead of colliding.

Also includes the schema file in the matrix job name so sibling jobs are distinguishable in the Actions UI.

Validated with `actionlint` and a live `curl + jq` simulation of the derivation across the tricky cases and a sample of normal packages.

### DIFF
--- a/.github/workflows/generate-package-metadata.yml
+++ b/.github/workflows/generate-package-metadata.yml
@@ -31,7 +31,7 @@ jobs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
       prs: ${{ steps.list.outputs.titles }}
   check-for-package-update:
-    name: ${{ matrix.repoSlug }}
+    name: ${{ matrix.repoSlug }} (${{ matrix.schemaFile }})
     needs: generate-packages-list
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


fnune:
## Verification: no impact on existing packages

To confirm this change is safe, I replicated both code paths (old repo-slug regex vs new schema `.name` with repo-slug fallback) and ran them against all 95 entries in `community-packages/package-list.json`.

Result:

- **94 / 95 packages: the new name is identical to the old name.** Behavior is unchanged for them: same short name, same target `<name>.yaml`, same skip/PR decisions.
- **1 / 95 differs, and it's a fix:** `runpod/pulumi-runpod-native`. The old regex derives `runpod-native` from the repo slug, but the package is actually `runpod` (schema `.name: runpod`, file `runpod.yaml`). The old logic was already computing a name that matches no file; the new logic aligns with reality.
- **3 packages whose schema 404s at `HEAD/<schemaFile>`** (eventstorecloud, kubernetes-coredns, kubernetes-ingress-nginx): the `${from_schema:-$from_repo}` fallback produces exactly the old repo-derived name, so behavior is unchanged.

This workflow only generates version-bump update PRs for packages already in `package-list.json`; the derived name feeds the "skip if PR open" check and the `provider_short_name` output. It never adds list entries and never publishes to the live registry, so a name-derivation change cannot onboard a new package.

Net: no existing package breaks, no new package is published. The only observable change is that multi-package repos (defang) and the mislabeled runpod now get correct, distinct names.

borisschlosser (approved):

PR #11445, opened 2026-06-19
Hide deprecated packages behind a Type filter option

borisschlosser:
## What

Deprecated provider packages (those with `publisher: DEPRECATED` in their YAML) were listed intermixed with active packages in the registry A–Z list. This hides them by default and adds a **Deprecated** option to the existing multi-select **Type** filter so users can opt in to seeing them.

## Behavior

| Type selection | Shows |
|---|---|
| *(none)* — default | all non-deprecated packages |
| Provider | non-deprecated providers only |
| **Deprecated** | only deprecated packages |
| Provider + Deprecated | non-deprecated providers **+** all deprecated (union) |

Use-case (category) filters and the name search still AND with the above, unchanged.

## Changes

- **`package.html`** — tag each card with `data-deprecated`, derived from the publisher field (`eq .package.publisher "DEPRECATED"`). 29 packages currently tagged.
- **`package-list.html`** — add a `Deprecated` option to the Type filter group.
- **`packages.ts`** — deprecated cards stay hidden unless `Deprecated` is among the selected Type filters; when selected they count as a type match (union, consistent with the other Type options). Default filtering runs on initial load so deprecated packages start hidden.

## Notes

- Hiding is client-side (toggles a `hidden` class), the same mechanism the existing Type/Use-Case/search filters already use — so deprecated cards remain in the HTML and crawlable.
- No featured package is deprecated, so the (unfiltered) featured section needs no change.

### DIFF
--- a/themes/default/layouts/partials/registry/package-list.html
+++ b/themes/default/layouts/partials/registry/package-list.html
@@ -41,6 +41,11 @@ append . }} {{ end }} {{ end }}
                                     label="Component"
                                     >Component</pulumi-filter-select-option
                                 >
+                                <pulumi-filter-select-option
+                                    value="deprecated"
+                                    label="Deprecated"
+                                    >Deprecated</pulumi-filter-select-option
+                                >
                             </div>
                         </pulumi-filter-select-option-group>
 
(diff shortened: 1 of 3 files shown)


borisschlosser:
Screenshot 2026-06-19 at 10 53 41
Screenshot 2026-06-19 at 10 54 52
Screenshot 2026-06-19 at 10 55 05

djperlovsky (approved):

PR #11650, opened 2026-07-10
Port the docs language-chooser redesign to the registry

jeffmerrick:
**Known issue:** Most of the choosers use the unwrapped version so get the floating tabs even if above a single code block. Not worth the effort to fix IMO.

Screenshot 2026-07-10 at 12 38 21 PM

---

Ports the language-chooser redesign from the docs site to the registry theme: restyled chooser toolbar, badges, copy button, notes, and tooltips, plus refreshed language/tech icons. Adds `pythontoolchain` as a new choosable/preference type across the chooser component, Redux store, and reducers, with e2e coverage. Reworks light-mode code block styling and updates the Stencil dependency and lockfile.


jeffmerrick:
@claude Re-review

themes/default/theme/stencil/package.json
```diff
@@ -25,7 +25,7 @@
         "@types/uuid": "^8.0.0",
         "jest": "^26.6.3",
         "jest-cli": "^26.6.3",
-        "puppeteer": "^9.1.1",
+        "puppeteer": "^24.43.1",
```
cnunciato, line 28: Quite a jump! 😄

cnunciato (approved):
Looks good! Clicked through all the key pages, and don't see any issues. I like what you did with the `PREVIEW` label, too, and the HCL icon looks great. Been quite a while since this stuff was last updated here! 😅 

> Not worth the effort to fix IMO.

Yep, agreed.

PR #11702, opened 2026-07-15
Remove all traces of the tf2pulumi-service conversion backend

kmosher:
## Summary
`pulumi/home#4844` is decommissioning the tf2pulumi-service conversion backend — its hosted converter pages and the `<pulumi-convert>` web component that called it were already removed from the docs site in `pulumi/docs#20250`. This repo held the last remaining references to it:

- `scripts/ci/build.sh` — dropped the `PULUMI_CONVERT_URL` stack-output lookup against `pulumi/tf2pulumi-service/production`. Left in place, this would go from "returns empty" (the stack currently has no `url` output) to a hard failure once the stack is removed — which would take down every CI build under `set -o errexit`.
- The `pulumi-convert` Stencil component (`themes/default/theme/stencil/src/components/convert/`) and its Hugo partial (`themes/default/layouts/partials/convert.html`) — confirmed unreferenced by any live template; nothing renders this widget on the current site.
- Matching cleanup: `config.yml`'s `getenv` allowlist, the custom-element `display: block` list in `main.scss`, a commented-out PurgeCSS safelist entry, and `BUILD-AND-DEPLOY.md`'s build-step/stack-reference docs.
- Regenerated `components.d.ts` via `stencil build` rather than hand-editing the generated file.

## Test plan
- [ ] `stencil build` completes cleanly (done locally as part of this change — regenerated `components.d.ts` with zero `convert` references and no other diff noise)
- [ ] A normal preview/production doc build completes with `PULUMI_CONVERT_URL` no longer set
- [ ] Spot-check a registry package page still renders correctly (no missing custom element)

Linked from pulumi/home#4844.


jkodroff (approved):
I wonder why anything was here in the first place. I think it's leftover from when the registry was broken out from pulumi/docs.

PR #11747, opened 2026-07-20
deps: combine Pulumi dependency updates and migrate to pulumi/pkg v3.253

borisschlosser:
## What

Combines the two open Pulumi dependency-update PRs into a single change:

- #11615 — renovate, "Update first-party Pulumi dependencies" (`pulumi/pkg`+`sdk` → 3.253.0, `esc-action`, `yarn.lock`, java, dotnet-language, `golang.org/x/text`)
- #11565 — dependabot, "bump the all-go-dependencies group" (`pulumi-hcl`, `pulumi-java`, `pulumi-dotnet-language`, `pulumi/pkg`+`sdk`)

Both target overlapping `pulumi/pkg` versions; this PR takes the **union of the highest** versions and reconciles `go.mod`/`go.sum`.

## Why a code change was needed

Both source PRs are **red**. Bumping `pulumi/pkg` past 3.247 is a breaking API change for `resourcedocsgen` that neither bot could resolve, so the update requires a small migration:

- `pcl.BindProgram` now takes the schema loader positionally; the `pcl.Loader` bind option was removed.
- The `plugin` package was split between `pkg` and `sdk`. The schema-loader host is rewired: `pkghost.New` now takes the loader/mapper/resolver service factories (previously passed to `plugin.NewContext`), and the loader uses `pkg/v3/resource/plugin`. A `nil` resolver is passed — schema loading downloads referenced providers via plugins and never exercises the registry resolver.
- `schema.ImportSpec` / `BindSpec` now require a **non-nil** loader; tests use `schema.NewNullLoader()` (matching pulumi's own tests).
- `pulumi-hcl` 0.10.0's `GenerateProgram` emits `main.tf` (was `main.hcl`) and takes options; we pass `SkipRequiredProvidersVersion()`, which is documented for embedded SDK doc snippets.

## Dependency notes

- **`pulumi-hcl` 0.2.0 → 0.10.0**, not dependabot's 0.9.0: 0.9.0's pinned `pulumi-yaml` pseudo-version was rebased away upstream and no longer resolves via `GOPRIVATE`-direct fetch. 0.10.0 is the latest and resolves cleanly.
- `pulumi-dotnet/pulumi-language-dotnet/v3` lands at `3.107.4-…` (required by `pulumi-hcl` 0.10.0), ≥ both PRs' 3.107.3.
- `go.mod` was tidied with `go mod tidy -e`: `pulumi-terraform-bridge`'s *own test* imports `sdk/.../resource/testing`, removed in sdk 3.253.0. It's a transitive test dependency we never build; CI runs only `go test ./...` + `golangci-lint`, neither of which touches it.

## Visible doc output changes (regenerated golden files)

- HCL constructor examples now include `lifecycle { create_before_destroy = true }` — `pulumi-hcl` 0.10.0 renders Pulumi's default (create-before-replace) semantics.
- Code-snippet trailing-newline normalization from `schema.RenderDocsToString` in `pulumi/pkg` 3.253.

## Testing

- `go build ./...`, `go test ./...`, and `golangci-lint run` (v2.10.1, matching CI) all pass in `tools/resourcedocsgen`.

Closes #11615
Closes #11565


fnune (approved):
The GitHub UI breaks trying to render this PR. I submitted https://github.com/pulumi/registry/pull/11761 to help with this in the future.

Anyway, it seems like the only important bits are the new `SkipRequiredProvidersVersion` call, which seems to have been made for docs (so 👍), and the `.hcl` -> `.tf` change, which the `*.golden` test fixtures agree with.

PR #11614, opened 2026-07-08
Split defang into defang-aws, defang-gcp, defang-azure

fnune:
Continues #11109 from @lionello (DefangLabs).

The upstream `pulumi-defang` repo split into three cloud-specific providers, and all three are now published (schemas + npm/PyPI/NuGet SDKs) at v2.3.1. #11109 only wired up `defang-azure`; this completes the split:

- Adds `defang-aws`, `defang-gcp`, `defang-azure` to `package-list.json` and bundles regenerated metadata at v2.3.1.
- Deprecates the old `defang` umbrella package (`publisher: DEPRECATED`, so the publish pipeline skips it) and removes its stale content.

This became unblocked once #11244 merged, which derives the package short name from the schema so the three no longer collide in the nightly.

@lionello could you confirm this matches what you intended for the split (naming, deprecation of `defang`, v2.3.1)? Once you give a thumbs up we'll merge.

PR #11756, opened 2026-07-21
Fix package metadata generation for multi-package repos (defang)

borisschlosser:
## What went wrong

The nightly [**Check for Community Package Updates**](https://github.com/pulumi/registry/actions/runs/29806607510) run failed for all three `DefangLabs/pulumi-defang` schemas (`defang-aws`, `defang-azure`, `defang-gcp`):

```
Error: generating package metadata: --providerName doesn't match the schema name: "defang" != "defang-aws"
```

(The `git ... exit code 128` lines in that run are unrelated post-job cleanup warnings — `No url found for submodule path 'examples'` — not the cause.)

`DefangLabs/pulumi-defang` publishes three separate schemas from one repo (the split in #11614). The **Generate Package Metadata** step invoked `resourcedocsgen metadata from-github` without `--providerName`, so it inferred the name from the repo slug (`pulumi-defang` → `defang`) and rejected it against each schema's real `name` (`defang-aws`, etc.) at `tools/resourcedocsgen/cmd/metadata.go:320`.

The workflow already derives the correct per-schema name (`steps.package.outputs.name`, added in #11244) for PR naming — it just wasn't being passed to the metadata command.

## Fix

1. **Pass `--providerName`** (the schema-derived name) to `metadata from-github`, so each package in a multi-package repo validates against its own schema name. This also makes the general multi-package-per-repo case work going forward.
2. **Ref-alignment (robustness):** the existing name read resolves at `HEAD`, but the metadata command downloads the schema at the **release tag**. A new `Read package name at the release version` step re-reads the name pinned to the same release tag, so a provider that renames its schema between its latest release and `HEAD` can't cause a spurious `providerName`/schema-name mismatch. Both `--providerName` and the PR's `provider_short_name` now use this release-pinned name. The `HEAD` read is kept solely as the early skip-dedup key (the release version isn't known at that point).

## Verification

- YAML parses cleanly.
- No change to single-package repos: the release-pinned read returns the same schema name that `resourcedocsgen` reads, so the validation passes as before. `.yaml`-schema packages fall back to the existing repo-derived name.
- The three failed defang jobs will publish on the next scheduled run (or a manual `workflow_dispatch`) once this merges, since the matrix is read from `community-packages/package-list.json`.

### DIFF
--- a/.github/workflows/generate-package-metadata.yml
+++ b/.github/workflows/generate-package-metadata.yml
@@ -98,19 +98,41 @@ jobs:
       - name: Build resourcedocsgen
         if: env.PROVIDER_VERSION
         run: go build -C tools/resourcedocsgen
+      - name: Read package name at the release version
+        # The skip-dedup read above resolves the name at HEAD, because the
+        # release version isn't known yet at that point. Re-read it here pinned
+        # to the exact ref the metadata command downloads — the release tag — so
+        # a provider that renamed its schema between its latest release and HEAD
+        # can't trigger a spurious providerName/schema-name mismatch. Falls back
+        # to the HEAD-derived name.
+        if: env.PROVIDER_VERSION
+        id: package_at_version
+        run: |
+          package_name=""
+          if [[ "$SCHEMA_FILE" == *.json ]]; then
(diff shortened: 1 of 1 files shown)


fnune (approved):
Thanks. I thought I had done enough with my previous PRs but it seems it wasn't:

- #11244
- #11614

PR #11760, opened 2026-07-21
Fix missing GitHub icon on how-to-guide "View Code" buttons

jeffmerrick:
## Problem

On how-to-guide pages (e.g. [aws-ts-nextjs](https://www.pulumi.com/registry/packages/aws/how-to-guides/aws-ts-nextjs/)), the GitHub icon in the **"View Code"** button is missing / appears as a white box on production, while it renders fine locally.

## Root cause

FontAwesome was removed from the theme; a shim in `_phosphor.scss` renders the legacy `<i class="fab fa-github">` markup (emitted by `mktutorial` into every how-to-guide page) as a masked GitHub SVG:

```scss
i.fab.fa-github { background-color: currentColor; -webkit-mask: url("/icons/github.svg") ...; }
```

`/icons/github.svg` **404s on the merged production site** (this repo's static assets under that path aren't served there), so the mask image fails to load. With no mask, the `<i>` paints as a solid `currentColor` box — white on the violet button, i.e. the icon looks missing / wrong color.

It renders fine **locally** only because `hugo serve` serves `/icons/github.svg` from the theme's static dir (200). The local↔prod difference is that single 404.

*(The "Deploy with Pulumi" `button.svg` was a red herring — it loads fine on prod; the earlier blank render was a transient CDN blip.)*

## Fix

Inline the GitHub mark as a **data-URI mask** so the icon is self-contained and no longer depends on a hosted asset path. Verified on the live prod page by swapping in the data-URI mask — the octocat renders correctly.

Fixes **all existing** how-to-guide pages on the next deploy, with **no content regeneration** needed (they all share this one shim).

Changed files: `themes/default/theme/src/scss/_phosphor.scss` only.

### DIFF
--- a/themes/default/theme/src/scss/_phosphor.scss
+++ b/themes/default/theme/src/scss/_phosphor.scss
@@ -20,6 +20,11 @@
 // every generated file from a styling refactor, so render those <i> tags as
 // the GitHub SVG via mask-image. Add more entries if other fa-* names appear
 // in content.
+//
+// The mask is an inline data URI rather than a hosted path (e.g. /icons/github.svg):
+// the registry's static assets under that path are not served on the merged
+// production site, so a hosted reference 404s and the <i> paints as a solid
+// currentColor box. Inlining keeps the icon self-contained and deploy-agnostic.
 i.fab.fa-github,
 i.fa-github {
     display: inline-block;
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
Thanks!

PR #11786, opened 2026-07-22
resourcedocsgen: re-sync neutralizer fence tracker on wrapping shortcodes

borisschlosser:
Minimal fix for the megaport publish pipeline (PR #11646, red since 2026-07-10, megaport stuck at 1.10.1).

A stray, unpaired code fence in the upstream registry.opentofu.org-mirrored doc desyncs the fence tracker in `neutralizeHugoShortcodes`, which then disarms a real `{{< chooser >}}`/`{{% choosable %}}` below it to `{{ <` — aborting the Hugo build and failing lint. Wrapping shortcodes (chooser/choosable/example(s)) wrap code examples from the outside and never appear inside one, so a line that is one of them is now never neutralized and re-syncs the tracker.

## Changes
- `pkg/output.go`: `wrappingShortcode` regex + re-sync in `neutralizeHugoShortcodes`
- `pkg/output_test.go`: regression test reproducing the megaport stray-fence desync

Verified: `metadata from-urls` against the live megaport 1.11.1 docs produces clean output. Once merged, the next regeneration of `megaport/publish-metadata` turns #11646 green.

Note: this PR originally carried a broader delimiter-handling unification; it was slimmed to the essential fix. The original commits remain in the branch history (`8b0ec3011`, `d01c29074`) if that work is picked up again later.


tools/resourcedocsgen/cmd/sanitize_cmd.go
```diff
@@ -18,63 +18,154 @@ import (
...
+//
+// It is deliberately tolerant of line endings: a pre-existing "\r\n" in
+// hand-authored content is not a broken-delimiter problem — Hugo handles CRLF
+// fine — so it compares NormalizeDocs's output against the input with only line
+// endings normalized. A file is flagged if and only if a shortcode delimiter is
+// malformed (stray whitespace outside a fence) or un-neutralized (a live
+// delimiter inside a code fence).
+func docNeedsNormalizing(content []byte) bool {
+	lineEndingsOnly := bytes.ReplaceAll(content, []byte("\r\n"), []byte("\n"))
+	return !bytes.Equal(pkg.NormalizeDocs(content), lineEndingsOnly)
```
fnune: There is semantic overlap here...

- `docNeedsNormalizing` seems to know what it means for a doc to be normalized (it embeds the knowledge of this line-endings-only fix)
- but then calls `pkg.NormalizeDocs`, which presumably also knows about the line endings

I think this should be inlined within the call-sites around `if check { ... }` so that:

- We run `pkg.NormalizeDocs` the same way for both branches `check` and `!check`
- Then decide to write back to the file or not based on `check`

Makefile
```diff
@@ -8,18 +8,23 @@ ensure:
...
 
-# The auto-generated provider _index.md pages are skipped by the front-matter
-# linter, yet a single malformed shortcode delimiter there fails the whole site build.
+# Malformed Hugo shortcode delimiters (e.g. "{{ <") abort the whole Hugo site
+# build, and the auto-generated provider _index.md pages are skipped by the
+# front-matter linter above. resourcedocsgen owns the canonical delimiter
+# normalization (pkg.NormalizeDocs), so it also owns the check — sharing that exact
+# logic means this lint can never drift from what the generator writes. It is also
+# enforced in CI by resourcedocsgen's Go test suite
+# (TestCommittedContentIsDelimiterCanonical), which runs in the resourcedocsgen job.
```
fnune: Nit: remove this comment

fnune (approved):
Generally way too much context in comments. But the intention seems fine to me.

tools/resourcedocsgen/cmd/sanitize_cmd.go
borisschlosser: Restructured the single-file path to compute `pkg.NormalizeDocs` once and branch on `--check` vs write, as suggested.

I kept `docNeedsNormalizing` as a small shared predicate rather than fully inlining, because `--check`, `--in-place`, and the CI guard all have to agree on what "non-canonical" means — inlining would duplicate the same comparison in three places (and a subtle divergence there is exactly what @claude's review caught: `--in-place` was using raw `bytes.Equal` and would have rewritten the CRLF-only how-to-guides that `--check` leaves alone). The line-endings knowledge is inherent: mktutorial-generated how-to-guides carry CRLF and never pass through `NormalizeDocs`, so the check has to compare against the CRLF-normalized input, not the raw bytes. Open to a different shape if you feel strongly.

Makefile
borisschlosser: Trimmed to a single line.

borisschlosser:
Addressed all three findings in `d01c290`.

**🔴 High — CI enforcement gap (confirmed).** You were right: `check-go.yml`'s `test` job sparse-checks-out only `${{ inputs.path }}`, so `themes/default/content` was never fetched and `TestCommittedContentIsDelimiterCanonical` hit its `t.Skip` — no job actually enforced the check. Fixed by:
- adding an optional `extra-sparse-checkout` input to `check-go.yml` and passing `themes/default/content` from the resourcedocsgen job in `pull-request.yml`, so the tree is present and the guard runs;
- making `findRepoContentDir` **fail** (not skip) under `GITHUB_ACTIONS` when the tree is absent, so a mis-scoped checkout surfaces loudly instead of silently disabling the guard again.

The PR description's "no CI plumbing change" claim was wrong and has been corrected.

**🟡 Low — `--in-place` vs `--check` CRLF inconsistency (confirmed).** `--in-place` now gates on the same `docNeedsNormalizing` predicate as `--check`, so the remediation command both of them print no longer rewrites canonical-but-CRLF files. Added two tests: `--in-place` leaves a CRLF-only canonical file byte-for-byte untouched, and `--check` passes on it. (Worth noting for the record: those CRLF files are mktutorial-generated how-to-guides that don't pass through `NormalizeDocs`, so normalizing them here would just be undone on the next mktutorial run — CRLF-tolerance is load-bearing, not cosmetic.)

**🔵 Nit — license headers.** Unified the new files on the de-indented form (matching the `pkg` package's existing convention).

Verified: `go test ./...` + `golangci-lint` clean; the guard runs (not skips) locally and fails on an injected regression; both workflow files parse.

@claude another pass when you get a chance.

tools/resourcedocsgen/cmd/sanitize_cmd.go
fnune: Hmm there's some smell here that Claude is refusing to go around or address. I don't think Claude did what I was asking.

fnune: The `--check` flag should try to do exactly the same as the command without the `--check` flag. Why do we need to differ?

tehsis (approved):

borisschlosser:
@claude I trimmed the PR to an essential change. Please re-review.

fnune (approved):

PR #11863, opened 2026-07-27
Don't generate empty API docs pages for types-only schema modules

jkodroff:
Fixes #11862.

`resourcedocsgen` registered a `modContext` for every module it saw — including modules discovered only through object types — and wrote an `_index.md` for every entry in the map. Since `index.tmpl` renders Modules / Resources / Functions and has no types section, a module holding only shared types produced a page that is empty by construction, plus a left-nav entry and a line in its parent's Modules list. awsx's `awsx:awsx:*` helper types are the motivating case: [`/registry/packages/awsx/api-docs/awsx/`](https://www.pulumi.com/registry/packages/awsx/api-docs/awsx/).

## The change

`pruneEmptyModules` runs at the end of `generateModulesFromSchemaPackage` — the single place that builds both the module map and the parent/child links, so pruning there fixes the page, the nav, the parent index and the CLI/LLM docs bundle together. A module survives if it has resources or functions of its own, or a descendant that does; an empty intermediate parent like `tree` is kept when `tree/v1` has content, and the root module is always kept.

Types are unaffected — they still render inline as supporting types on the resource pages that reference them, which reads the schema rather than the module map.

### Also: index page title tags

Pruning exposed a latent bug. The index template picked its `title_tag` from `len(modules) > 0`, using "does this page list submodules?" as a stand-in for "is this the package page?", so a package whose only module was pruned would have flipped from `foo Package` to `foo.foo`. It is now keyed off the root module. That corrects two pre-existing cases as a side effect, which is most of the golden churn in this PR:

- module-less packages: root page `example.example` → `example Package`
- module pages that have submodules: `tree Package` → `plant.tree`, and they now get a real meta description instead of an empty one

## Verified

- `make test` and `make lint-resourcedocsgen` pass; goldens regenerated with `PULUMI_ACCEPT=1` / `-update`. Every deleted golden page was empty; the two deleted parents (`crd.k8s.amazonaws.com`, `meta`) listed nothing but their own empty child.
- New `pkg/docs/prune_test.go` — four tests over an inline schema covering a types-only module, an empty parent with a non-empty child, and a function-only module. They assert the module map, the parent's `children`, the emitted file set, the nav tree, and **that a pruned module's types still render on the referencing resource's page**. All four fail with the prune commented out.
- **awsx end to end** (`make api-docs/awsx`): `api-docs/awsx/` gone, root index lists only `cloudtrail`/`ec2`/`ecr`/`ecs`/`lb`, nav JSON has no `awsx` module, and `ecs/fargateservice` still documents `DefaultLogGroup` / `DefaultRoleWithPolicy` / `LogGroup` / `RoleWithPolicy`. No dangling links to the removed page.
- **kubernetes end to end**: the `meta/v1` sanity anchor survives with `Status` and `StatusPatch`. 9 type-only modules pruned. **eks**: unaffected.

## Blast radius

**6 packages lose 12 module pages**, each verified by generating docs with the pre-fix and post-fix binaries and diffing the output:

| Package | Pages removed | Modules pruned |
|---|---|---|
| `castai` | 6 | `autoscaling`, `azure`, `iam`, `organization`, `rebalancing`, `workload` |
| `defang-aws` | 2 | `aws`, `compose` |
| `awsx` | 1 | `awsx` |
| `defang-azure` | 1 | `compose` |
| `defang-gcp` | 1 | `compose` |
| `netskope-publisher` | 1 | `provider` |

All 12 were empty in the pre-fix output (654-877 bytes of front matter and a Package Details block). `kubernetes`, `aws`, `gcp`, `airbyte` and `docker` are unaffected.

A types-only module only ever got a page when one of its types was reachable from some resource's `inputProperties`/`stateInputs` — that is the only path that calls `getMod(add: true)` for a type. The `pkg:config/*` types of bridged providers are referenced by nothing, so they were never registered and never had a page or nav entry in the first place. See the [detailed catalog and spot-check list](https://github.com/pulumi/registry/pull/11863#issuecomment-5096511949) for the full analysis, including one page that *moves* rather than disappears (`netskope-publisher`'s Provider resource docs relocate from `res-provider/` to `provider/` once the module name collision is gone).

The `title_tag` fix is the wider change: 221 of 268 packages have no modules, so their root page title goes from `airbyte.airbyte` to `airbyte Package`.

## Caveat

Pruning removes pages that exist today, so external links to e.g. `/registry/packages/awsx/api-docs/awsx/` will start 404ing, as will `netskope-publisher`'s `res-provider/`. Given the pruned pages are empty I did not add Hugo aliases - flagging it for review.

Generated with [Claude Code](https://claude.com/claude-code)


jkodroff:
## Correction: the affected-package catalog above is wrong

My original catalog claimed 50 packages / 65 pages. The real numbers are **6 packages / 12 pages**. Two specific claims in the PR description are wrong and should be ignored: **kubernetes loses nothing** (I claimed 9 pages) and **aws loses nothing** (I claimed `applicationloadbalancing`).

### Why the first count was wrong — and why `config` never displayed

@jkodroff spotted that `config` never renders as a module on the live site ([airbyte](https://www.pulumi.com/registry/packages/airbyte/)). It never did, and the reason is that a module only enters the module map via `getMod(..., add: true)`, which happens in exactly three places in `generateModulesFromSchemaPackage`:

* the module of each resource (including the provider),
* the module of each non-method function,
* the module of each `ObjectType` **reachable from some resource's `inputProperties` or `stateInputs`** (`visitObjects`, gen.go:2322).

The trailing "Find nested types" loop passes `add: false`, so a type discovered only there never creates a module.

Bridged providers emit `pkg:config/*` types for the SDK's config namespace, and **no resource input references them** — the Provider resource gets a parallel `index/Provider*` type instead. airbyte is the clean case: `airbyte:config/googleIap:googleIap` has zero references anywhere in the schema, while the Provider's `googleIap` input points at `airbyte:index/ProviderGoogleIap`. So `config` was never registered, never had a page (hence `/api-docs/config/` 404s across aws, gcp, docker, snowflake, tls, github, postgresql) and never had a nav entry. Same story for kubernetes' `pkg/version`, `authentication.k8s.io/v1` and friends, all of which 404 today and are absent from the live nav.

My first scan keyed off token namespaces alone and ignored reachability, so it counted every types-only module instead of every *registered* types-only module. `aws:applicationloadbalancing` was a third variant: it isn't in the live nav either, and `/api-docs/applicationloadbalancing/` 200s only because it serves the `aws.alb` page via an alias.

### Corrected catalog — verified against the generator, not inferred

Each row below was produced by generating docs with the pre-fix binary and the post-fix binary and diffing the emitted pages.

| Package | Pages removed | Modules pruned |
|---|---|---|
| `castai` | 6 | `autoscaling`, `azure`, `iam`, `organization`, `rebalancing`, `workload` |
| `defang-aws` | 2 | `aws`, `compose` |
| `awsx` | 1 | `awsx` |
| `defang-azure` | 1 | `compose` |
| `defang-gcp` | 1 | `compose` |
| `netskope-publisher` | 1 | `provider` |

All 12 removed pages were confirmed empty in the pre-fix output (654–877 bytes: front matter plus a Package Details block, no Modules/Resources/Functions section).

Confirmed **unaffected**: `airbyte` and `docker` by generator diff; `kubernetes`, `aws` and `gcp` by their live navs, whose module counts match the content-bearing module count exactly (71/230/142) and contain none of the candidates. `kubernetes`'s `meta/v1` is in the live nav and keeps `Status`/`StatusPatch`.

### One page moves rather than disappears

`netskope-publisher` declares a module named `provider`, which collided with its `Provider` resource, so the conflict resolver had been renaming the *resource* page to `res-provider/`. With the empty `provider` module pruned the collision is gone, so:

* before: `/api-docs/provider/` = the empty module page, `/api-docs/res-provider/` = the Provider resource docs (17 KB)
* after: `/api-docs/provider/` = the Provider resource docs, no `res-provider/`

The resource docs land on the better URL, but `/api-docs/netskope-publisher/api-docs/res-provider/` will start 404ing. This is the only case in the PR where a non-empty page changes location — worth a look on the preview.

### Spot-check list for the preview site

| URL | Expect |
|---|---|
| `/registry/packages/awsx/api-docs/` | no `awsx` in Modules; `awsx/` 404s |
| `/registry/packages/awsx/api-docs/ecs/fargateservice/` | still documents `DefaultLogGroup`, `DefaultRoleWithPolicy`, `LogGroup`, `RoleWithPolicy` |
| `/registry/packages/castai/api-docs/` | only the one content-bearing module remains |
| `/registry/packages/defang-aws/api-docs/` | no Modules section at all |
| `/registry/packages/netskope-publisher/api-docs/provider/` | the **Provider resource** page, not an empty module page |
| `/registry/packages/kubernetes/api-docs/meta/v1/` | unchanged, still lists `Status` and `StatusPatch` |
| `/registry/packages/aws/api-docs/` | unchanged, 230 modules |

### The `title_tag` change is the wide one

The prune touches 6 packages, but the `title_tag` fix touches nearly all of them: **221 of 268 packages have no modules**, so their root API-docs page `<title>` goes from `airbyte.airbyte` to `airbyte Package`. The remaining 47 keep their root title and instead see nested module pages that have submodules change from `tree Package` to `plant.tree` (plus gaining a real meta description). Both are corrections, but that is where the golden churn comes from and it is the part most worth a second opinion.

CamSoper (approved):
Most of this is golden tests, so I didn't look at those. The actual code changes look right and we'll know pretty quickly if they're not.

I've kicked off a [test deployment to pulumi-test.io](https://github.com/pulumi/registry/actions/runs/30306179598). You should watch this to make sure everything comes out right before you merge.

PR #11946, opened 2026-08-03
Add automated provider inventory script

CamSoper:
Adds `scripts/provider_inventory.py`, an automated inventory of all registry provider pages. The script scans every provider YAML under `themes/default/data/registry/packages/`, cross-references the committed content pages under `themes/default/content/registry/packages/` (presence of `_index.md` and `installation-configuration.md`), cross-references the community package list, and analyzes each page's structure (heading outline, section concepts, example language coverage, word count, vendor-fetched provenance, extras). It writes a single JSON inventory.

Run with:

```bash
uv run --with pyyaml scripts/provider_inventory.py -o provider-inventory.json
```

Current output covers 308 packages. Notable data-quality signals it surfaces today: 30 packages with publisher `DEPRECATED`, 103 packages missing a committed `installation-configuration.md`, 87 packages not updated in over a year, 181 distinct overview heading outlines across 308 providers, and a `Pulumiverse`/`pulumiverse` publisher-casing inconsistency.

Unit tests covering the community-list matching and page-analysis logic live in `scripts/test_provider_inventory.py`, wired into the existing `Test CI scripts` workflow.


CamSoper:
Thanks — verified each finding against the tree before applying. All three main findings were real; one minor claim did not hold up. Fixed in 297b42a (plus 04b8f52 untracking `__pycache__` I committed by mistake).

### Applied

**1. `in_community_list` — confirmed and fixed.** Reproduced exactly the four packages named. Took both halves of the suggestion (schema-path name + `repo_url` match against `repoSlug`). Verified the flips and, importantly, that `defang` itself does *not* regress to `false` — it no longer appears in the derived slug set once the three `defang-*` entries resolve via `schemaFile`, and is now covered by the `repo_url` half instead:

```
defang-aws:   False -> True
defang-azure: False -> True
defang-gcp:   False -> True
runpod:       False -> True
defang:       True  -> True   (via repo_url)
total in list: 97 -> 101
```

**2. `chooser_langs` — confirmed and fixed.** Switched to extraction. Counts across committed `_index.md` pages matched the report (`hcl` came out at 67 rather than 68, immaterial). 49 pages changed, mostly 6→7 picking up `hcl`; `metabase` now resolves the `"csharp,fsharp,visualbasic"` group correctly. Checked for alias double-counting before taking this — no page uses both `nodejs` and `typescript`, so treating them as distinct is safe.

**3. `analyze_extras` / `api-docs` — confirmed as a latent bug, fixed.** `api-docs/` is git-ignored at `.gitignore:32` and absent from a clean checkout, so it does not affect current output, but it would corrupt the inventory on any built tree exactly as described.

**Minors applied:** dropped the no-op `round()`; `try`/`except yaml.YAMLError` with the same WARN treatment; `isinstance` guard on non-dict frontmatter; hoisted `import re`; documented the `uv run --with pyyaml` invocation; added `provider-inventory.json` to `.gitignore`; settled on f-strings throughout. The `l` comprehension variable went away with finding 2.

**Tests added.** Renamed to `provider_inventory.py` so it is importable (nothing referenced the old path), and added `scripts/test_provider_inventory.py` — 13 tests covering `load_community_slugs`, `analyze_page`, and `analyze_extras`, including regression cases for all three findings above. Existing discovery was scoped to `scripts/ci`, so `Test CI scripts` gained a second discovery step and a `scripts/*.py` path filter.

### Not applied

**The `aws-v6` / `azure-native-v2` claim does not hold.** Neither package exists in the tree — no package directory at all, let alone a committed `how-to-guides` directory (`scripts/redirects/aws-v6-redirects.txt` suggests `aws-v6` was retired). `scripts/ci/mktutorial.sh:33-45` guards both with `[ -d "$dest_dir" ] || continue`, and that second loop only runs `mktutorial cleanup` — it never copies guides in. So adding them to `MKTUTORIAL_PACKAGES` would assert a refresh that does not happen. I took the other option offered and narrowed the comment to say those six are the ones refreshed from `pulumi/examples`, noting the two prune-only slugs.

---

jkodroff (approved):

PR #11949, opened 2026-08-03
Add dark mode to the Registry

### LINKED ISSUE #11668: Dark mode support for Registry
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

https://github.com/pulumi/registry/issues/244 added dark mode to the docs, but didn't add dark mode for the registry. This ticket is to to add dark mode to the registry. 

When I am browsing the registry the white background is tough on my eyes. Adding dark mode support would help immensely for late-night hackers like me and also for those who prefer a dark theme at all times.

jeffmerrick:
This is probably not perfect but should be good enough while we plan larger improvements to the registry.

---

Ports the light/dark/system theme system from `pulumi/docs` to the Registry, reusing the same `pulumi-docs-theme` localStorage key so the preference is one setting across pulumi.com — picking dark on `/docs` no longer drops you onto a white wall when you click through to a provider's API docs. Dark is a pure override gated on `html[data-theme="dark"] body.section-registry`, driven by semantic `--docs-*` tokens in the new `theme/src/scss/docs/_docs-theme.scss`, with a pre-paint inline script in `head.html` so there's no flash and a three-state toggle pinned to the bottom of the sidebar. Package logos are generated rather than hand-maintained: `scripts/generate-dark-logos.py` writes `-on-dark.svg` variants for the 74 local marks whose paints are too dark for the dark page, and `scripts/classify-external-logos.py` measures each external `logo_url` for contrast against the dark background and records the 48 of 93 that need a light chip. Along the way this fixes three pre-existing layout bugs the toggle exposed — the copy button sat 16px above standalone code panels, the sidebar was sized taller than the viewport, and it carried dead padding at its foot — so a few rules also change in light mode; `AGENTS.md` documents the token vocabulary and both generator scripts for future contributors.

Verified in-browser across the list, overview, installation-configuration and API-reference pages in both modes: `make lint`, `make test`, Cypress (14/14) and a full `hugo` build (9521 pages) all pass.

Fixes #11668.


jkodroff (approved):
Clicked around a few representative pages in the AWS package, and it looks good to me. Approving.

PR #11997, opened 2026-08-06
[registry] Gate every publish path on a matching schema version

fnune:
## Why

The registry publishes a package's schema straight from its release tag. The service rejects a schema whose own `version` field disagrees with the version being published:

```text
400 Bad Request: Schema version (0.0.0) must match query parameter version (0.4.5)
```

#11995 catches this for community submissions via `/check`, but that's one of three paths into the registry. First-party providers arrive through `publish-provider-update.yml`, and nightly bumps arrive through `generate-package-metadata.yml`; neither had a gate. `terraform-provider` is a first-party package and has failed the primary publish path on every release since March for exactly this reason.

## What

`scripts/ci/validate_schema_version.py` fetches a schema and compares its declared version against the version being published. An absent version passes — that's the bridged-provider convention, and the service then takes the version from the publish request. It reads the version out of the first 512 KB where it can, so the multi-hundred-megabyte schemas don't have to be downloaded whole.

It runs from the two uncovered paths:

- `pull-request.yml` — a `Validate Schema Versions` job over every changed package YAML. This covers the nightly community bumps and the first-party PRs, since both land as PRs.
- `publish-provider-update.yml` — validates the schema at the dispatched tag before generating metadata, so a first-party provider fails loudly at release rather than in a PR nobody is watching.

The `push-provider-update` path needs no gate: it already derives the version from the schema with `jq .version`, so it cannot mismatch.

## Not breaking existing packages

Ran the validator across the whole registry:

```text
278 packages checked: 104 matching, 169 without a version field, 5 mismatched, 0 not checkable.
```

All five mismatches are seeded into `ALLOWED_MISMATCHES` with a link, so the gate reports them but does not block:

| package | declares | publishes | upstream |
|---|---|---|---|
| `aem` | `0.1.2-alpha.1718635808+2feb6b9e` | `0.1.3` | wttech/pulumi-aem |
| `genesiscloud` | `v0.0.35` | `0.0.35` | schema keeps the `v` prefix |
| `kafka-connect` | `0.0.8+dirty` | `0.0.12` | azaurus1/pulumi-kafka-connect |
| `powerplatform` | `0.0.0` | `0.4.5` | rpothin/pulumi-powerplatform#82 |
| `terraform-provider` | `0.0.0-dev` | `1.3.0` | pulumi/pulumi-terraform-provider#113 |

`aem`, `genesiscloud` and `kafka-connect` are new information, and all three are third-party (wttech, genesiscloud, azaurus1) — none has released since March, so none has ever exercised the current publish path. They are latent failures that would have surfaced on their next release. `genesiscloud` is only the `v` prefix.

Remove each waiver as its upstream lands. When the map is empty, the gate is unconditional.

## Notes

The rule is duplicated between this script and `verify_entry.py` in #11995 because they check different inputs — a package YAML in the registry versus an upstream release tag during onboarding. Worth unifying if a third caller shows up.

Reviewing after #11995 is easier; they don't conflict.


fnune:
Superseded: rather than porting only the schema-version check to the upgrade paths, the follow-up runs the full `/check` suite (docs generation, plugin and SDK install probes, doc-lint, publisher allowlist, schema version) on every changed package YAML. Built on top of #11995.

PR #12056, opened 2026-08-11
Pin the PR preview comment and link the changed pages

CamSoper:
Ports the pinned preview comment from `pulumi/docs` to this repo.

Today `scripts/ci/sync.sh` posts a **brand-new comment on every commit**, so a long-running PR accumulates a wall of near-identical "Your site preview for commit … is ready!" comments — and none of them tell a reviewer which pages to actually look at. This replaces that with **one comment per PR, rewritten on every build**, carrying:

1. The preview URL for the current commit (`<bucket-website>/registry/`).
2. A **Changed pages** list — direct links to the pages this PR changed.

Rendered body:

```
<!-- registry-preview-link -->
Your site preview for commit abcd1234 is ready! :tada:

http://registry--origin-pr-42-abcd1234.s3-website.us-west-2.amazonaws.com/registry/

**Changed pages:**
- [/registry/packages/aws/](http://…/registry/packages/aws/)
- [/registry/packages/gcp/](http://…/registry/packages/gcp/)
```

### How it stays pinned

The body opens with the marker `<!-- registry-preview-link -->`. `upsert_github_pr_comment` (`scripts/ci/common.sh`) finds the comment carrying it and `PATCH`es that one, only `POST`ing when there's no match.

Two deliberate departures from the `pulumi/docs` original:

- **The comment list is paginated.** The docs version issues an unpaginated `GET` on `…/issues/<n>/comments`, and GitHub returns issue comments oldest-first, 30 per page — so it only ever searches the 30 oldest. The pinned comment is normally created by the first build and sits safely inside that window, but on a PR that collects 30+ comments *before* its first successful preview build (fact sheets, `/check` runs, review chatter, builds that fail before commenting), it lands at position 31+ and becomes permanently invisible to the lookup. Every later build then appends another duplicate. This version walks up to 10 pages of 100, so ordering doesn't matter.
- **The lookup matches the comment author** (`github-actions[bot]` / `pulumi-bot`) as well as the marker, so a contributor can't redirect the pinned comment by quoting the marker in a comment of their own.

The stray `.` glued to the end of the old URL (`…/registry.`) is gone too.

### How changed pages are resolved

The changed-file list comes from the GitHub files API rather than a local `git diff`, so it behaves identically for the `pull_request` build and the maintainer-triggered `/preview` command (which runs the same composite action against a synthetic event file). `changed_paths_to_urls` maps each path under two rules:

| Changed path | URL |
|---|---|
| `themes/default/content/**/*.md` | Hugo's own rules (`content_path_to_url`) |
| `themes/default/data/registry/packages/<pkg>.yaml` | `/registry/packages/<pkg>/` |

The YAML rule is the one that carries this repo — the generated `api-docs/` content is gitignored and never appears in a PR diff, so a straight port of the docs' content-markdown-only rule would leave the section empty on most registry PRs. (The `<pkg>` slug is the YAML basename; verified it equals the file's `name:` field for all 308 packages, and results are existence-checked anyway.)

Results are de-duplicated (a PR touching both `aws.yaml` and `aws/_index.md` lists the page once) and filtered to URLs that actually rendered in this build, so `url:`/alias overrides and removed files drop out instead of becoming dead links. Capped at 50 with an "…and N more" line.

### Safety

The whole block is reporting, not deployment, so it's invoked as `post_preview_comment || log …` and can never fail an otherwise-good build. It no-ops cleanly when the event payload or token is missing (local runs). The preview URL also goes to `$GITHUB_STEP_SUMMARY`, so the link is on the job page even if the comment can't be posted.

Behavior that did **not** change: the comment is still posted only after a fully successful build, since `sync.sh` runs under `set -o errexit` after the Cypress smoke test.

### Testing

`scripts/ci/test-preview-comment.sh` (new) covers the path→URL mapping, de-duplication, ordering, and the existence gate — offline, no network or Hugo build. Wired up as `make test-preview-comment` and run in the existing `Lint Scripts` job, which is already in `sentinel.needs`.

Verified locally:

- `make test-preview-comment` — 12/12 pass.
- A stubbed-`curl` harness driving the real functions end-to-end: no existing comment → `POST`; existing pinned comment → `PATCH` of that comment (and *not* of a decoy comment from another author carrying the same marker); files API returning an error object → comment still posts, `Changed pages` section omitted; missing token → clean skip, exit 0.
- `bash -n` on both modified scripts. Note `make lint` doesn't cover bash — `yarn run lint` is eslint + prettier over `.js`, and shellcheck isn't in the toolchain.

Still to verify on this PR itself: that the real `preview` job produces exactly one comment, that a second push edits it rather than adding another, and that the changed-page links resolve.

### Known limitation (not addressed here)

`.github/workflows/pull-request.yml` has no `concurrency:` block, so two rapid pushes can produce overlapping preview runs that both read-then-write the pinned comment and create a duplicate. Adding `concurrency` with `cancel-in-progress` would fix it but would also start cancelling in-flight preview builds — a behavior change worth its own PR.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -87,6 +87,8 @@ jobs:
         run: yarn install
       - name: Run Linter
         run: yarn run lint
+      - name: Test preview comment
+        run: make test-preview-comment
 
   lint-dark-logos:
     name: Lint Dark Logo Variants
--- a/BUILD-AND-DEPLOY.md
+++ b/BUILD-AND-DEPLOY.md
@@ -894,10 +894,12 @@ marker means a contributor can't redirect the pinned comment by quoting the mark
 comment list is paginated deliberately — GitHub returns 30 comments per page by default, and an
 unpaginated search would miss the marker on a long PR and post a duplicate on every build.
 
-**How changed pages are resolved**: the changed-file list comes from the GitHub API
-(`/pulls/<n>/files`), not a local `git diff`, so it works identically for the `pull_request`
-build and the maintainer-triggered `/preview` command. Each path is mapped to a URL by
-`changed_paths_to_urls` (`scripts/ci/common.sh`) under two rules:
+**How changed pages are resolved**: `changed_pages_section` (`scripts/ci/common.sh`) reads the
+changed-file list from the GitHub API (`/pulls/<n>/files`), not a local `git diff`, so it works
+identically for the `pull_request` build and the maintainer-triggered `/preview` command. It
+collects every API page before mapping — `changed_paths_to_urls` de-duplicates only within a
+single invocation, so mapping page by page would double-list a package whose YAML and landing
+page straddle the 100-file page boundary. Each path is mapped under two rules:
 
 | Changed path | URL |
 |---|---|
--- a/BUILD-AND-DEPLOY.md
+++ b/BUILD-AND-DEPLOY.md
@@ -24,7 +24,8 @@ This document describes the build, test, and deployment system for the `pulumi/r
    - [6.1 Pulumi IaC](#61-pulumi-iac)
    - [6.2 AWS Resources](#62-aws-resources)
    - [6.3 S3 Preview Bucket Lifecycle](#63-s3-preview-bucket-lifecycle)
-   - [6.4 Custom Redirects](#64-custom-redirects)
+   - [6.4 The Pinned Preview Comment](#64-the-pinned-preview-comment)
+   - [6.5 Custom Redirects](#65-custom-redirects)
 7. [Registry Publication (push-registry.py)](#registry-publication-push-registrypy)
 8. [Testing Strategy](#testing-strategy)
    - [8.1 Go Unit Tests](#81-go-unit-tests)
@@ -588,7 +589,7 @@ PR opened / committed
         │                       ├── s5cmd sync llm-docs-out/ → bucket (Content-Encoding: gzip)
         │                       ├── Run browser tests (Cypress smoke test)
         │                       ├── Write origin-bucket-metadata.json
-        │                       └── Post PR comment with preview URL
+        │                       └── Update the pinned PR comment (preview URL + changed pages)
         │
         └── sentinel (depends on all jobs above)
                 └── Writes "Sentinel" GitHub status check = success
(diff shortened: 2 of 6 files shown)


CamSoper:
Thanks — #2 was a real bug and #5 was the right call. Addressed in 812e85f.

**#2 (per-page de-duplication) — fixed.** Correct diagnosis: `changed_paths_to_urls` was re-invoked per API page with a fresh `seen` map. Took the accumulate-then-map-once shape you suggested, and added a regression test that stubs a 100-entry page one holding `aws.yaml` and a page two holding `aws/_index.md`. I checked the test actually fails against the old logic before keeping it:

```
OLD (per-page mapping):
- /registry/packages/aws/
- /registry/packages/aws/
NEW (accumulate then map):
- [/registry/packages/aws/](http://preview.example/registry/packages/aws/)
```

**#1 (jq on a non-array) — fixed, but the impact was narrower than described.** The premise is right: `.[]` on `{"message": "Bad credentials", ...}` yields strings and `.body` is then a hard jq error, exit 5. The stated consequence isn't, though. `errexit` suppression is inherited through the whole dynamic extent of the left side of a `||`, so inside `post_preview_comment || log …` the failing assignment doesn't abort the function — execution continues, `found` is empty, `page_count` computes 0, the loop breaks, and the `POST` fallback runs. Verified against the real function with a stubbed error response:

```
--> POSTED (fallback worked)
--> script survived
```

So the behavior already was "breaks cleanly", and no comment was being lost. Applied the type guard anyway — `2>/dev/null` was masking a genuine jq error, and the function would abort if it were ever called outside a `||` context. Locked it in with a test.

**#5 (no coverage of the pinning logic) — fixed.** Fair hit, and it's what would have caught #2. Committed the `curl`-stub harness: no existing comment → `POST`; pinned comment present → `PATCH` of that id; decoy carrying the marker under another author → ignored (both orderings); error payload → falls back to `POST`. To make this testable I moved `changed_pages_section` into `common.sh` with explicit `build_dir`/`base_url` parameters rather than reaching for `sync.sh` globals. Suite is 19 assertions now.

**#6 (silent failures) — fixed.** The write now checks `-w '%{http_code}'` and logs the method, URL, and status on non-2xx instead of dropping the comment silently. Added `--max-time 30` to every GitHub call, which also removes the ten-page stall you flagged.

**#3 (`max_listed_pages`) — renamed** to `max_listed_urls`.

**#4 (shared pagination helper) — declining for now.** With `changed_pages_section` moved, both loops are in `common.sh` within ~40 lines of each other and both now carry the same `if type == "array"` idiom, so the duplication is visible and consistent. A generic helper in bash would have to take a jq filter plus an accumulator callback, which reads worse than the two explicit loops. Happy to revisit if a third paginated call site shows up.

One note on your closing point: since #1 doesn't actually swallow the comment, the concurrency window is unchanged — a race still produces a duplicate rather than nothing. Still worth its own PR.

---

jkodroff (approved):

PR #12086, opened 2026-08-13
[registry] Mark secret properties in the API docs

### LINKED ISSUE #1620: Add symbol to represent outputs that are marked as secrets
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently we have a symbol in the API docs that shows if an input will cause the resource to be replaced if changed. 

Having a symbol to show if an output is a secret would be useful. 

### Affected area/feature

API docs

jkodroff:
Property lists already flag required inputs and inputs that force replacement, but nothing showed which properties the provider schema marks `secret: true`. Reading a resource's Outputs section, there was no way to tell which values are encrypted in state.

Fixes #1620

## Changes
- `resourcedocsgen`: carry `schema.Property.Secret` through to a `property-secret` class, wherever the flag is set (outputs, inputs, nested types, function results).
- LLM docs bundle: add `*(secret)*`, plus the `*(replaces on change)*` suffix `cli_properties.tmpl` never emitted despite the data being on the struct.
- `misc.ts`: the badge injection was three combinatorial passes over required × replacement — a third flag would have made it seven. Now one table-driven pass. Both icons are inline `currentColor` SVGs so they follow the dark theme; the old `` painted its own baked-in gray.
- Tests: new `gen_cli_test.go` subtest (that template had no coverage), a secret output on the synthetic provider, and a Cypress assertion on the rendered badge.

## Verification
`make lint`, `go test ./...`, and `make test_provider_api_docs` (22/22 Cypress specs in a real browser) all pass. Dark mode checked against the compiled stylesheet: `--docs-fg-muted` flips gray-700 → gray-300 and nothing overrides the badge color.

Second commit is mechanical golden regeneration — review the first commit.


jkodroff:
## Preview pages to review

Preview for commit `e1b4dc6f`. These six cover the distinct paths through the change — chosen by scanning real provider schemas, not guessed.

**1. Three badges on one property** — the highest-risk layout case
http://registry--origin-pr-12086-e1b4dc6f.s3-website.us-west-2.amazonaws.com/registry/packages/consul/api-docs/peering/#peeringtoken_nodejs

`peeringToken` carries `property-required property-replacement property-secret` together — asterisk, arrows and padlock in one row. This combination was unrepresentable in the old two-badge CSS. The page also has 7 `property-optional property-replacement property-secret` rows (two icons, no asterisk).

**2. Secret outputs — the literal ask in #1620**
http://registry--origin-pr-12086-e1b4dc6f.s3-website.us-west-2.amazonaws.com/registry/packages/random/api-docs/randompassword/

`result` and `bcryptHash` in Outputs; padlock alone, no other badge.

**3. Two badges, required + secret**
http://registry--origin-pr-12086-e1b4dc6f.s3-website.us-west-2.amazonaws.com/registry/packages/tls/api-docs/certrequest/#privatekeypem_nodejs

The combination the old code special-cased with its own container div — confirms the refactor didn't regress it.

**4. Nested modules + supporting types**
http://registry--origin-pr-12086-e1b4dc6f.s3-website.us-west-2.amazonaws.com/registry/packages/kubernetes/api-docs/core/v1/secret/

`data`/`stringData` are secret on both the resource and the nested type.

**5. Functions/invokes** — a separate call into the shared property template
http://registry--origin-pr-12086-e1b4dc6f.s3-website.us-west-2.amazonaws.com/registry/packages/tls/api-docs/getpublickey/

**6. Regression at scale**
http://registry--origin-pr-12086-e1b4dc6f.s3-website.us-west-2.amazonaws.com/registry/packages/aws/api-docs/rds/instance/

217 replacement badges plus 28 secret ones — confirms the rewritten injection loop still handles a large page.

### Already verified, so reviewers don't need to
- All six return 200 and contain `property-secret` (counts are multiples of 7, one per language tab).
- The deployed JS bundle contains the new `property-badges` injection loop.
- **The CSS survived PurgeCSS.** Worth calling out: `.property-badges` and `.property-indicator-secret` appear only inside the JS bundle and never in the HTML, so the CI purge step could have stripped them — and a local `make serve` would not have caught it. Both rules are present in the deployed stylesheet, along with `--docs-fg-muted` at gray-700 (light) / gray-300 (dark).

### Still needs human eyes
The mechanics are verified; the aesthetics are not. On page 1 especially: whether the 18px padlock sits right against the small red asterisk and the arrows, and whether the `gap` spacing reads better than the old hand-tuned margins. Then hover for the tooltip, and toggle dark mode from the sidebar — padlock and arrows should be the same muted gray, asterisk still red.

Note: preview links are per-commit, so these go stale if new commits land.

jkodroff:
Verifying the URLs by hand:

All are correct, but note that for 3, the assessment is wrong: privateKeyPem is not required (or there's an existing bug in rendering it): https://www.pulumi.com/registry/packages/tls/api-docs/privatekey/#privatekeypem_nodejs

CamSoper (approved):
This looks fine to me, though we should probably check and see if anything is actually *using* the LLM bundles (I doubt we are) so we might think about deprecating those.

Not a thing I want to investigate right now. Let's :shipit: as is

PR #12103, opened 2026-08-14
[ci] Bound and retry every HTTP call in the build

fnune:
`Build and deploy preview` hung 59 minutes on #12098 against a 17-minute
median, reporting only "The hosted runner lost communication with the server".
`resourcedocsgen` made every request on `http.DefaultClient`/`http.Get`, which
have no timeout, and no job set `timeout-minutes`.

Across 1604 runs of five workflows (Jan 28 - Aug 14) there were 334 failed
jobs. About half are flakes, meaning they would pass on a re-run with no code
change; the rest fail for real reasons and are out of scope here. This change
addresses roughly 56% of the flakes.

| Bucket | Failed | Flakes (est) | Fixed here (est) |
| --- | ---: | ---: | ---: |
| `Check for a new version` | 134 | 134 | 67 |
| `Build and deploy preview` | 24 | 20 | 20 |
| other | 50 | 10 | 5 |
| `Generate Package Metadata` | 52 | 0 | 0 |
| `Test Provider API Docs` | 22 | 0 | 0 |
| `resourcedocsgen / Test` | 18 | 0 | 0 |
| `Test Live Registry Publish` | 13 | 0 | 0 |
| `resourcedocsgen / Lint` | 12 | 0 | 0 |
| `Lint Markdown` | 9 | 0 | 0 |
| **Total** | **334** | **164** | **92** |

Of ten sampled `Check for a new version` failures, two had expired logs and the
remaining eight split evenly: four are an unretried TCP reset in
`cmd/pkgversion.go`, and four are `go: downloading` against `proxy.golang.org`,
which is the unfixed half of that bucket. `Generate Package Metadata` is one
broken package across four schema variants, and the remaining buckets are lint
findings, golden-file drift and Cypress assertions.

### Network timeouts

- `pkg.NewHTTPClient` bounds dial, TLS handshake and response headers, and caps
  one attempt at 10 minutes, since schemas reach 50MB. It replaces
  `http.DefaultClient`, `http.Get` and one ad-hoc `&http.Client{}` across seven
  call sites
- `CURL_NETWORK_OPTS` in `generate-versioned-docs.sh` adds a connect timeout and
  a stall detector that aborts below 1KB/s for 60s. A flat `--max-time` would
  kill a legitimately slow 50MB download
- The two fetches that captured stdout now write to a file first. `curl` cannot
  rewind a pipe, so a retry appends the second body to the partial one already
  written and still exits 0, feeding `partial + full` to `yq` and `jq`.
  Reproduced against a local server that stalls once, then serves cleanly; `-o`
  truncates on retry and yields the clean body

### Retries

- The Go client retries connection errors, 429 and 5xx three times with backoff.
  404 is not retried, so `getSchemaFromRegistry`'s `ErrPackageNotFound` path is
  unchanged. `hashicorp/go-retryablehttp` was already in the module graph
- `curl` retries three times. `--retry-all-errors` is omitted so the `_index.md`
  fetch, which 404s by design and falls back to the latest version's copy, does
  not retry three times per package

### Job timeouts

- All 41 jobs across 23 workflows set `timeout-minutes`, sized at 2-3x the
  observed maximum over recent runs, replacing the 6-hour default

### Failure legibility

- Drops the `examples` gitlink and gitignores the path.
  `update-tutorials.yml` checks `pulumi/examples` out into the repo tree and
  `create-pull-request` stages it, so git recorded a gitlink with no
  `.gitmodules` entry, 140 commits' worth. Every checkout then ended in `fatal:
  No url found for submodule path 'examples'` and a `git` exit 128 warning,
  which GitHub shows as the first annotation on any failed job, masking the real
  cause. Deleting the gitlink alone would not hold: the next nightly run
  recreates it. The examples SHA stays recorded in the bot's commit message

### Wiring

- The client is a constructor rather than a package-level `var`, injected from
  `main` through a one-method `HTTPDoer` per consuming package, per
  pulumi/pulumi-service#46674

## Test plan

1. Automated tests
    - `go build`, `go vet`, `go test ./...` in `tools/resourcedocsgen`: pass
    - `golangci-lint run`: no findings in changed files (two pre-existing gosec G122 hits remain in the untouched `internal/tests/util/fs.go`)
    - `shellcheck scripts/generate-versioned-docs.sh`: clean
2. Manual checks
    - All workflows parse; all 41 jobs report a `timeout-minutes`
    - No `http.DefaultClient`, `http.Get` or ad-hoc `&http.Client{}` remains outside `NewHTTPClient`


borisschlosser (approved):
Good improvement! Guess we need to monitor the timeouts and adjust them if they are too strict.

PR #12153, opened 2026-08-19
Fix duplicated "Packages" crumb in registry breadcrumbs

### LINKED ISSUE #11874: Breadcrumbs show "packages" twice in path
"Packages" appears twice the breadcrumbs for a package's home page. Example: https://www.pulumi.com/registry/packages/kubernetes/

Image

jkodroff:
Every registry package page rendered its breadcrumb as `Packages / Packages / Kubernetes`. The `/registry/` crumb was hardcoded to the literal string "Packages", colliding with the real `/registry/packages/` section page whose own `linktitle` is also "Packages". The root crumb is now labeled "Registry" — the name the JSON-LD BreadcrumbList and the Algolia ancestor mapping in `scripts/search/page.js` already use.

Also syncs the JSON-LD BreadcrumbList, which was missing the Packages level, so the structured data mirrors what's visible on the page.

## Changes

- `registry/breadcrumb.html` — label the `/registry/` crumb "Registry"
- `registry/structured-data.html` — add the Packages crumb to the BreadcrumbList and renumber positions; keyed off the URL rather than the package data so `/registry/packages/` gets it too
- `structured-data.cy.js` — update the BreadcrumbList name assertions

## Verification

| Page | Visible | JSON-LD |
|---|---|---|
| `/registry/packages/` | Registry › Packages | 1:Registry › 2:Packages |
| `/registry/packages/kubernetes/` | + Kubernetes | + 3:Kubernetes |
| `…/installation-configuration/` | + Installation & Configuration | + 4:Installation & Configuration |
| `…/eks/api-docs/` | + API Docs | + 4:API Docs |

`make lint` clean.

Fixes #11874

### DIFF
--- a/cypress/e2e/structured-data.cy.js
+++ b/cypress/e2e/structured-data.cy.js
@@ -107,8 +107,10 @@ describe("Registry", () => {
                     expect(crumb, "BreadcrumbList block").to.exist;
                     const names = crumb.itemListElement.map(i => i.name);
                     // Literal equality proves no embedded quote characters.
+                    // The trail mirrors the visible breadcrumb: Registry > Packages > <package>.
                     expect(names[0]).to.equal("Registry");
-                    expect(names[1]).to.equal("AWS");
+                    expect(names[1]).to.equal("Packages");
+                    expect(names[2]).to.equal("AWS");
                 });
             });
         });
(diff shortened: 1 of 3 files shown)


jeffmerrick (approved):
Sorta strange that both /registry/ and /registry/packages exist and are the same but that's not really relevant to this fix. Looks good!

Issue #12209, opened 2026-08-25
Audit duplicated content between _index.md and installation-configuration.md


jkodroff:
The Overview and Install and Config pages for Aiven are duplicated:

- https://www.pulumi.com/registry/packages/aiven/
- https://www.pulumi.com/registry/packages/aiven/installation-configuration/

We need to figure out how many providers have this problem and how to fix.

https://github.com/pulumi/registry/issues/12206#issuecomment-5402322260 explains how most of the providers are published.

jkodroff:
Audited all 311 package directories under `themes/default/content/registry/packages/` against `origin/master` @ `37892ade3`. Method: strip front matter, drop fenced code, split both pages on `##`, alias equivalent headings (`Schema` ≈ `Configuration Reference` ≈ `Configuration Options`, `Example Usage` ≈ `Example`, `Credentials` ≈ `Authentication`), then compare section-by-section with a sequence-similarity ratio. Scripts and the full per-package JSON are reproducible; happy to land them under `scripts/` if useful.

## How many providers have this

| | packages |
|---|---|
| Package directories | 311 |
| …with **no** `installation-configuration.md` (already single-page) | 105 |
| …with **both** pages | **206** |
| **…of those, with at least one substantially-duplicated section** | **65** |

Broken into severity:

| tier | what it looks like | n |
|---|---|---|
| **A1 — page is wholly redundant** | every section of the config page also appears in `_index.md`, near-verbatim. **This is the Aiven case.** | **7** |
| **A2 — page is *only* a duplicated install list** | the config page's sole content is the "available as a package in all Pulumi languages" bullet list, which `_index.md` already shows | **8** |
| **B — substantial overlap** | ≥40% of the config page's prose, or ≥2 sections, duplicated; the rest is unique | **23** |
| **C — install-list overlap only** | the pages differ, but both open with the same install bullet list | **27** |
| **D — no meaningful overlap** | genuinely distinct pages (`aws`, `gcp`, `azure`, and every dynamically-bridged package) | 141 |

**A1:** `aiven` `alicloud` `artifactory` `auth0` `consul` `dbtcloud` `digitalocean`
**A2:** `cloudinit` `heroku` `ise` `random` `scm` `sdwan` `tailscale` `tls`

Per-tier tables with the exact duplicated sections are in the appendix below.

## Root cause

[...]

Issue #11866, opened 2026-07-27
Epic: Overview, Installation and Configuration Cleanup


jkodroff:
Cleanup of existing issues with our 2 "standard" docs pages for providers: Overview, and Install & Config.

Goals:

- Remove duplicated content
- Fix any discovered bugs in the current automation
- Ensure complete and accurate documentation of the current automation
- Clarify any gaps in the current requirements for these pages and ensure automation enforces these requirements

Non-goals:

- Any substantive changes to the submission/automation process

PR #12231, opened 2026-08-26
Delete the Install & config pages wholly duplicated by the Overview (tier A)

### LINKED ISSUE #12225: Delete the Install & Config pages that are wholly duplicated by the Overview (tier A)
Part of #11866. Follows the audit in #12209.

Every section of the Install & Config page also appears on the Overview, near-verbatim. These pages carry nothing of their own — `aiven` is the case that opened #12209:

- https://www.pulumi.com/registry/packages/aiven/
- https://www.pulumi.com/registry/packages/aiven/installation-configuration/

## The pattern

`installation-configuration.md` says substantially the same thing as `_index.md`, and **`pulumi/registry` is the source of truth for that page** — it is hand-written in this repo, carries no `# WARNING: this file was fetched from` provenance header, and no automation has ever touched it or ever will.

That is what makes this set actionable on its own: the Overview beside it is regenerated from the provider repo on every release, so the duplicate drifts further out of date with each one and nothing in the pipeline will ever correct it.

Matching is on section-level similarity, not exact text — most of these forked from the same content in Aug 2024 and have diverged in wording since while staying the same page. Verified against `master` @ `37892ade3`.

**11 packages** are in this tier. The other two tiers are #12226 (substantial overlap) and #12227 (install list only), and the full audit is #12209.

## Packages

| package | duplicated sections | config page last changed | overview last changed |
|---|---|---|---|
| `aiven` | Authentication, Env vars, Example usage, Installation, Resource Options | 2024-08-22 | 2026-07-02 |
| `alicloud` | Authentication, Config reference, Example usage, Installation | 2024-08-26 | 2026-08-24 |
| `artifactory` | Authentication, Config reference, Example usage, Installation | 2024-08-23 | 2026-08-14 |
| `auth0` | Config reference, Env vars, Example usage, Importing Resources, Installation | 2024-08-23 | 2026-08-18 |
| `cloudinit` | Installation | 2024-08-23 | 2026-07-22 |
| `consul` | Config reference, Env vars, Example usage, Installation | 2024-08-23 | 2026-07-23 |
| `dbtcloud` | Config reference, Installation | 2024-08-26 | 2026-08-06 |
| `digitalocean` | Config reference, Example usage, Installation | 2024-08-26 | 2026-08-19 |
| `random` | Installation | 2023-06-05 | 2026-07-22 |
| `tailscale` | Installation | 2023-11-28 | 2026-07-13 |
| `tls` | Installation | 2023-11-28 | 2026-07-22 |

## Scope

Delete `installation-configuration.md` for all 11. Nothing needs to be salvaged first — every section is already on the Overview, and where the two differ the Overview is the fresher copy in every case (compare the last-changed columns above).

The sidebar entry disappears on its own: `themes/default/layouts/partials/registry/package-nav.html:25-31` guards the link on `fileExists`, and 105 packages already ship with no config page, so this is a supported shape needing no template work.

One caveat worth carrying even though it does not apply to any package here: for `heroku`, `junipermist` and `ec` the *generated* Overview advertises an SDK that does not exist (`@pulumi/heroku`, `@pulumi/junipermist`, `Pulumi.Ec` all 404 on npm/NuGet) because the install-list generator derives names by convention instead of reading `language.*.packageName` from the schema. Do not generalise "the Overview is fresher, delete the other" into a blanket rule.

## Upstream is clear for this tier

Checked every package here for `docs/installation-configuration.md` in its provider repo at the currently released tag: **all return 404.** No upstream file exists to re-create these pages, so deleting in `pulumi/registry` is sufficient and durable.

## Re-check upstream when the copy bug is fixed

`installation-configuration.md` is not reliably copied from upstream today — that is the underlying bug, and it is why these pages froze while their Overviews kept moving. **The upstream check has to be re-run as part of fixing it.** The clean 404 sweep above describes today's behaviour, not a guarantee: once the copy path works, any provider repo that does ship `docs/installation-configuration.md` will have that content flow into the registry, and if the upstream file is itself a duplicate of the provider's own `docs/_index.md` then deleting the registry copy achieves nothing — it returns on the next release, and the deletion has to happen **upstream in the provider repo** instead.

That shape already exists for 16 packages (`heroku` `ise` `scm` `sdwan` `harness` `junipermist` `slack` `ovh` `cpln` `vercel` `bytepluscc` `checkly` `descope` `formal` `nvidia-aicr` `volcenginecc`), which duplicate the Overview from an upstream-owned config page. They are out of scope here and need upstream PRs — see #12209.

## Blocked on redirects

Deleting the page 404s a live URL. Hugo `aliases:` on the Overview is the right mechanism, but it cannot be hand-added: `tools/resourcedocsgen/cmd/metadata.go` overwrites `_index.md` wholesale on every release, and every package in this tier has a generated Overview, so a hand-written alias is silently erased at the next release. The alias has to be emitted by the generator. Tracking separately.

jkodroff:
Fixes #12225. Part of #11866, following the audit in #12209.

## What

Deletes `installation-configuration.md` for the 11 tier-A packages — `aiven`, `alicloud`, `artifactory`, `auth0`, `cloudinit`, `consul`, `dbtcloud`, `digitalocean`, `random`, `tailscale`, `tls` — and 301s their URLs to the package Overview.

## Why these 11 are safe

Section by section, **each Overview is a strict superset** of the page being removed. Verified by comparing headings across all 11 pairs; where the two differ, the Overview is the fresher copy in every case.

`tailscale` is the clearest example of the drift this fixes. Its config page (last touched 2023-11-28) still documents:

```
$ export OAUTH_CLIENT_SECRET=XXXXXX
$ export OAUTH_CLIENT_ID=XXXXXX
```

The provider actually reads `TAILSCALE_OAUTH_CLIENT_ID` / `TAILSCALE_OAUTH_CLIENT_SECRET`, which the Overview's generated Configuration Reference gets right. The stale page was never going to self-correct — it's hand-written here (no `# WARNING: this file was fetched from` header), so no automation has ever touched it, while the Overview beside it is rebuilt from the provider schema on every release.

## Why the deletion sticks

`docs/installation-configuration.md` returns **404 in all 11 provider repos** at their currently released tags — re-verified against `pulumi-aiven@v6.58.0`, `pulumi-alicloud@v3.106.0`, `pulumi-artifactory@v8.11.6`, `pulumi-auth0@v3.51.0`, `pulumi-cloudinit@v1.6.1`, `pulumi-consul@v3.15.1`, `pulumi-dbtcloud@v1.11.4`, `pulumi-digitalocean@v4.79.0`, `pulumi-random@v4.21.1`, `pulumi-tailscale@v0.29.0`, `pulumi-tls@v5.5.1`. `metadata.go` skips the emit on empty content, so no release re-creates these.

## Redirects, rather than tracking separately

#12225 lists this as blocked on redirects, on the assumption that Hugo `aliases:` was the only mechanism — and that can't work, since `resourcedocsgen` rewrites each `_index.md` wholesale on every release and would erase hand-added front matter.

`infrastructure/redirects.js` is a second mechanism the issue doesn't consider: a CloudFront viewer-request function, **not generated**, sitting on the distribution's default cache behavior (only `/metadata.json` and `/fingerprinted/*` have their own behaviors, so `/registry/packages/*/installation-configuration/` is covered), already used to retire package URLs. One rule there unblocks this:

```js
{
    match: /^\/registry\/packages\/(aiven|alicloud|...|tls)\/installation-configuration(?:\.md|\/.*)?$/,
    replace: "/registry/packages/$1/",
},
```

The suffix group catches every form the URL can arrive in — bare, trailing slash, `/index.html`, and `/index.md` or `.md` from markdown content negotiation. Markdown clients keep their `Accept` header across the 301 and get the Overview as markdown at the target. Prefix redirects run before the `.md` rewrite, so the `.md` form hits the 301 rather than being rewritten to a now-missing origin object.

## Sidebar

No template work needed — `package-nav.html:25-31` guards the link on `fileExists`, the shape 105 packages already ship. Confirmed against production: `/registry/packages/airbyte/` (no config page) renders with zero `installation-configuration` references; `/registry/packages/aiven/` currently has exactly one, the sidebar link.

## Also

Repoints the README's Install & config example link at `aws`, since `aiven` no longer has one.

## Testing

- `make test-infra` — 35 pass, including 11 new cases covering each URL form, all 11 packages, and negative cases (`aws` still serves its page; `randomly` and `installation-configuration-old` are not caught by the alternation).
- `make lint` — clean (Go, shortcodes, markdown, dark logos, eslint/prettier).
- `scripts/ci/validate-packages.sh` — passes; the 11 now emit the expected `WARN: ... does not exist. Package will be displayed with a single overview file.`

## One thing spotted, not fixed here

`scripts/ci/push-registry.py:132` guards on `_installation-configuration.md` (leading underscore) before passing `--installation-configuration` for the un-underscored path, so that flag looks like it never fires. Pre-existing and unrelated to this change, but adjacent enough to be worth a separate issue.

https://claude.ai/code/session_014sMPUFA2Ft1r9XY9hNqTkh


jkodroff:
@CamSoper I'll spot check a few of the removed pages to verify that they were indeed duplicate content before merging.

CamSoper (approved):
YES. More of THIS.

PR #12240, opened 2026-08-26
Post the priority digest with a bot token, to a channel ID

borisschlosser:
## Why

`Scheduled jobs: Priority digest` has failed every run since 2026-08-21 (first: https://github.com/pulumi/registry/actions/runs/32495917947) with:

```
Slack rejected the message: 404 channel_not_found
```

It posted through the legacy incoming webhook's `channel` override, by name, and the target channel was renamed. The override resolves the name it is given against current names only, so the post has been going to a name that no longer exists.

## What

- `scripts/ci/priority_digest.py`: `post()` now calls `chat.postMessage` with `SLACK_ACCESS_TOKEN` (the bot token the link checker already uses from ESC) and checks the API's `ok` field (it returns HTTP 200 on failure). The `username`/`icon_url` override is dropped: bot posts carry the app's own identity.
- The channel is a channel **ID**, taken from `SLACK_CHANNEL`, which the workflow sets from the `SLACK_TEAM_CHANNEL` repository variable. An ID survives renames, and nothing about the channel lives in the code.
- `test_priority_digest.py` covers the request `post()` sends and the `ok: false` failure path.

Tracked in #12181.

## Before merging

1. ~~Invite the bot to the target channel; it is not a member today, and a bot can only post where it has been added (`not_in_channel` otherwise).~~ Done.
2. ~~Run `Scheduled jobs: Priority digest` via `workflow_dispatch` on this branch (dry-run off) and confirm the digest arrives.~~ Done: https://github.com/pulumi/registry/actions/runs/32961421697 posted the digest.

https://claude.ai/code/session_01C1TRPjAnd8ggqDYGHcMSDh

### DIFF
--- a/.github/workflows/priority-digest.yml
+++ b/.github/workflows/priority-digest.yml
@@ -37,7 +37,8 @@ jobs:
           ${{ inputs.dry-run && '--dry-run' || '' }}
         env:
           GITHUB_TOKEN: ${{ steps.esc-secrets.outputs.PULUMI_BOT_TOKEN }}
-          SLACK_WEBHOOK_URL: ${{ steps.esc-secrets.outputs.SLACK_WEBHOOK_URL }}
+          SLACK_ACCESS_TOKEN: ${{ steps.esc-secrets.outputs.SLACK_ACCESS_TOKEN }}
+          SLACK_CHANNEL: ${{ vars.SLACK_TEAM_CHANNEL }}
 
   notify:
     if: failure()
--- a/BUILD-AND-DEPLOY.md
+++ b/BUILD-AND-DEPLOY.md
@@ -785,9 +785,9 @@ Auto-merge is currently disabled (commented out in the workflow).
 
 **Trigger**: Daily at 3:00 PM UTC; also `workflow_dispatch` with a `dry-run` input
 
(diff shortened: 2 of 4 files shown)


borisschlosser:
Thanks — applied in the latest commit:

- **`BUILD-AND-DEPLOY.md`**: updated the `priority-digest.yml` section and the §9.2 table (`SLACK_ACCESS_TOKEN` now lists the digest, `SLACK_WEBHOOK_URL` no longer does, new `SLACK_TEAM_CHANNEL` row). Worded without naming the channel, since this repo is public.
- **`body.get('error')` → `None`**: now falls back to the raw response text; covered by `test_falls_back_to_the_raw_response_when_slack_gives_no_error`.
- **HTTP-error path**: `test_raises_on_http_errors` pins that `raise_for_status()` propagates.
- **`mrkdwn` / `unfurl_links`**: asserted in the request-shape test.

Already moot after earlier revisions: the channel ID is no longer in the code at all (only `vars.SLACK_TEAM_CHANNEL`), and both pre-merge steps are done — the bot is in the channel and a `workflow_dispatch` from this branch posted successfully (run 32961421697).

Intentionally not changed: the `notify` jobs' channel-by-name resolution. Those channel names are treated as stable, so that stays as is.

fnune (approved):

Issue #3391, opened 2023-10-25
Make docs contribution more flexible


cnunciato:
Today, we expect package authors to craft Hugo-specific Markdown files, named in a very specific way and containing specific frontmatter fields, to comprise their packages documentation. 

This is both too restrictive (authors only get two docs -- one overview page and one install page) and too implementation-specific (authors shouldn't have to deal with the esoteric details of Hugo or whatever other platform we're using in order to ship their packages, nor should we expect them to come up with page titles, meta descriptions, etc., for us). 

Instead, we should allow authors to submit plain GitHub-flavored Markdown files for their overview and install docs. Moreover, we should also provide some way for authors to provide additional free-form pages (e.g., how-to guides, upgrade/migration guides) to help their users address common scenarios.

jkodroff:
Worth reviving: #3159 asked for the same thing and was closed `by-design`, but its stated reasons no longer hold. The plan then was to move how-to guides off the Registry and to explore alternatives to building it statically with Hugo. Neither happened — guides are still rendered here, still via Hugo.

The change surface as the code stands today:

- **Fetching is the easy part.** `tools/resourcedocsgen/cmd/metadata.go` already builds `raw.githubusercontent.com/<repo>/<tag>/docs/<name>` from a repo slug and version it has in hand, and `pkg.AddGitHubAuthHeaders` is wired up. Enumerating `docs/` via the contents API or a release tarball is a small change against that.
- **Front matter is the real friction, and it's this issue's point.** Pages need `title`, `meta_desc`, and `layout: package`, and a fetched `_index.md` without front matter is a hard failure. Free-form files mean synthesizing those from the first H1, the filename, or the schema's `displayName`.
- **Nav derivation is mostly there.** `themes/default/layouts/partials/registry/package-nav.html` is a hardcoded list of slots, but the How-to guides section already ranges over `.Pages`. Deriving a package's nav from its file tree needs an ordering signal and a nesting decision, not new machinery.
- **Blast radius is the part that needs designing.** One build produces the whole site, so a single malformed page fails it for every package. Not hypothetical: `tools/resourcedocsgen/cmd/sanitize.go` exists because mirrored provider docs emit `{{ <` delimiters that abort the Hugo build, and we repair them on fetch. Going from two known files to arbitrary `docs/` contents widens that across repos we don't control. The mitigation is per-file validation that drops a bad page rather than failing the build.
[...]

PR #12248, opened 2026-08-26
Document previewing registry changes, split by audience

### LINKED ISSUE #11638: Add external documentation for building provider docs locally
We have internal documentation for [previewing registry changes](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/How-To/Preview%20registry%20changes.md). There's no reason that we can't make those public.

Related [internal Slack post](https://pulumi.slack.com/archives/C02FXTZEZ6W/p1713468287457019).

jkodroff:
Makes the internal "Preview registry changes" playbook public, split into two guides because it was mixing two unrelated jobs: previewing a package release before publishing it, and checking a change to this repository against packages already published.

- `docs/previewing-package-docs.md` — for package authors. Renders a schema on disk, stages a whole release locally, and gets a shareable preview.
- `docs/previewing-registry-changes.md` — for contributors here. Generates one package to work against, and covers the PR preview build.

Both open with the sources a registry page is assembled from, including how-to guides — which anyone can contribute here by pull request without owning the provider, and which the publish pipeline never overwrites.

## Changes

- Two guides, each pointing at the other, written in the style of `pulumi/docs`.
- `api-docs/<pkg>` now honors `SKIP_VERSIONED_DOCS`, matching the bulk target. Without it the target fails on a fresh checkout: `generate-versioned-docs.sh` installs `registry-mirror-discover` from the private `pulumi/registry-mirror-tools` ahead of the `is_blessed` filter that would skip it. CI is unaffected.
- README's single-package command pointed at the internal `.make/` sentinel path; switched to the `make api-docs/<pkg>` alias.

## Corrections found while verifying

Three claims in the first draft were wrong, each checked against a Hugo build:

- A missing nav JSON empties the sidebar rather than 404ing the page — it's fetched client-side and the error is swallowed.
- A missing `_index.md` costs the Overview page; the API docs still render.
- A `how-to-guides/` directory needs its own `_index.md`, or the guide renders with no section page and no sidebar entry. The original instructions would have produced an invisible guide.

Also verified that a file in the git-ignored root `content/` overrides the committed copy of the same page, which is what makes staging an entire release locally possible — schema output, Overview, install-config, and guides together, without touching a tracked file.

**One suggestion I did not take:** `make -B` in place of deleting `.generated`. The target's sentinel is an intermediate file Make deletes after each run, so the recipe already re-runs and the skip happens inside `resourcedocsgen`. Verified: `make -B SKIP_VERSIONED_DOCS=1 api-docs/random` still logs `Skipping (output is fresh)`.

Left out deliberately: consolidating with `docs/adding-a-new-package.md`. Adding a package and previewing a change are different jobs; a single source of truth deserves its own pass. Also filed the case for sourcing a provider's whole `docs/` folder on #3391 rather than acting on it here.

`make lint-markdown` passes. Since it only scans `themes/default/content`, I also checked every bash block with `bash -n` and verified all links and anchors resolve.

Fixes #11638

https://claude.ai/code/session_01LrrKUFU2J8AnQx5qX44Ma5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ api-docs/%: .make/content/registry/packages/$$*/api-docs ;
 		--baseSchemasOutDir ./static/registry/packages \
 		--baseLLMDocsOutDir ./llm-docs-out/registry/packages \
 		$*
-	CONTENT_DIR=$(CURDIR)/content/registry/packages STATIC_DIR=$(CURDIR)/static/registry/packages ./scripts/generate-versioned-docs.sh $*
+	$(if $(SKIP_VERSIONED_DOCS),,CONTENT_DIR=$(CURDIR)/content/registry/packages STATIC_DIR=$(CURDIR)/static/registry/packages ./scripts/generate-versioned-docs.sh $*)
 	@mkdir -p "$(@D)"
 	@touch $@
 
--- a/docs/previewing-registry-changes.md
+++ b/docs/previewing-registry-changes.md
@@ -1,0 +1,173 @@
+# Previewing Registry Changes
+
+This guide covers how to see Pulumi Registry pages — including provider API docs —
+before they are published, whether the change lives in a provider's schema, in this
+repository's templates, or in a pull request.
+
+Pick a route based on what you changed:
+
+| What you changed | Route |
+|---|---|
+| A provider schema you have locally, and it isn't released yet | [Local preview from a schema file](#local-preview-from-a-schema-file) |
+| A package that is already published in the registry (templates, layouts, CSS, metadata) | [Local preview of a published package](#local-preview-of-a-published-package) |
+| A provider schema you can publish to a public URL (e.g. a branch on GitHub) | [Local preview from a branch schema](#local-preview-from-a-branch-schema) |
+| Anything you're ready to push to `pulumi/registry` | [Pull request preview](#pull-request-preview) |
+
+## Prerequisites
+
+All local routes assume you've set the repository up once:
+
+```bash
+mise trust && mise install
+make ensure
+make build-assets
+```
+
+See [the README](../README.md#using-this-repository) for details.
+
+Generated docs land in git-ignored directories (`content/` and `static/registry/` at
+the repository root), so nothing from these routes should ever end up in a commit.
+
+## Local preview from a schema file
+
+Use this when you're changing a provider and want to see how its API docs will render
+before you cut a release. It reads a `schema.json` straight off your disk, so nothing
+needs to be pushed or published.
+
+1. Build the docs generator:
+
+    ```bash
+    make bin/resourcedocsgen
+    ```
+
+2. Generate your provider's schema in the provider repo. For a bridged provider that's
+   usually `make generate_schema`, which writes
+   `provider/cmd/pulumi-resource-<name>/schema.json`.
+
+3. From the root of this repository, run `resourcedocsgen docs` against that file,
+   writing into the two locations Hugo serves from:
+
+    ```bash
+    ./bin/resourcedocsgen docs \
+        --schemaFile ../pulumi-aws/provider/cmd/pulumi-resource-aws/schema.json \
+        --version v9.9.9 \
+        --docsOutDir ./content/registry/packages/aws/api-docs \
+        --packageTreeJSONOutDir ./static/registry/packages/navs
+    ```
+
+    `--version` is required and must be valid semver. Use a dummy version higher than
+    anything published so it's obvious in the rendered page that you're looking at a
+    local build.
+
+4. The package's landing pages (`_index.md` and `installation-configuration.md`) are
+   committed under `themes/default/content/registry/packages/<package>/` and are used
+   as-is. If you're previewing a package that isn't in the registry yet, create that
+   directory and add the two files by hand, copying them from your provider repo's
+   `docs/` folder.
+
+5. Serve the site:
+
+    ```bash
+    make serve
+    ```
+
+    Your pages are at `http://localhost:1313/registry/packages/<package>/api-docs/`.
+
+Re-run step 3 after each schema change; the running Hugo server picks up the new files.
+If you're also changing CSS or JavaScript under `themes/default/theme`, use
+`make serve-all` instead so assets rebuild too.
+
+## Local preview of a published package
+
+Use this when the schema is already published and you're changing something on this
+side — a docs template, a Hugo layout, the theme, or a package's YAML metadata.
+
+```bash
+make SKIP_VERSIONED_DOCS=1 api-docs/aws
+make serve
+```
+
+`make api-docs/<package>` reads
+`themes/default/data/registry/packages/<package>.yaml`, fetches the corresponding
+schema, and writes to `content/registry/packages/<package>/` at the repository root.
+
+`SKIP_VERSIONED_DOCS=1` skips generating the older-major-version snapshots, which
+requires a Pulumi-internal tool that most contributors can't install. Leave it set
+unless you are specifically working on versioned docs.
+
+To force a rebuild after a change the generator can't see (it caches on the package
+YAML and its own build), delete the sentinel file:
+
+```bash
+rm content/registry/packages/aws/api-docs/.generated
+```
+
+## Local preview from a branch schema
+
+Use this when you want the full registry pipeline — metadata, nav tree, published
+schema file, and all — but against a schema that only exists on a branch.
+
+1. Publish the schema to a public URL. For most providers the schema is committed, so
+   pushing your branch is enough; the raw URL looks like
+   `https://raw.githubusercontent.com/<org>/<repo>/<branch>/provider/cmd/pulumi-resource-<name>/schema.json`.
+   For providers whose schema is too large to commit (Azure Native, for example),
+   upload it to S3 or any other public host.
+
+2. Edit `themes/default/data/registry/packages/<package>.yaml`:
+
+    - Set `schema_file_url` to the URL from step 1.
+    - Set `version` to a semver version that has **not** been published to the Pulumi
+      Registry service — a bumped dummy version such as `v9.9.9`.
+
+    Both edits are required. `resourcedocsgen` asks `api.pulumi.com` for the package at
+    `version` first and only falls back to `schema_file_url` when that lookup 404s. If
+    you leave `version` at a published value you'll silently get the published schema
+    and none of your changes.
+
+3. Generate and serve:
+
+    ```bash
+    make SKIP_VERSIONED_DOCS=1 api-docs/<package>
+    make serve
+    ```
+
+Revert the YAML edit before committing.
+
+## Pull request preview
+
+Every pull request against `pulumi/registry` gets a full site build published to a
+per-commit S3 bucket. CI maintains a single pinned comment on the PR containing:
+
+- the preview URL for the current commit, and
+- a **Changed pages** list linking directly to the pages your PR affects.
+
+The comment is updated in place on each build rather than added per commit, and the
+preview buckets are deleted when the PR closes.
+
+If your change is in a provider repo rather than here, you can still get a preview by
+opening a PR against this repository with the
+[branch schema](#local-preview-from-a-branch-schema) edits applied — the same YAML
+change works in CI. Don't merge that PR; it exists to produce the preview.
+
+For a community package pull request, a Pulumi maintainer can comment `/preview` to
+build a live preview of the package's pages.
+
+## Troubleshooting
+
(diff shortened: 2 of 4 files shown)


docs/previewing-registry-changes.md
```diff
@@ -0,0 +1,173 @@
+# Previewing Registry Changes
+
+This guide covers how to see Pulumi Registry pages — including provider API docs —
+before they are published, whether the change lives in a provider's schema, in this
+repository's templates, or in a pull request.
+
```
jkodroff: This should be one line for the entire paragraph.

docs/previewing-registry-changes.md
```diff
@@ -0,0 +1,126 @@
...
+Use this when the schema is already published and you're changing something on this side — a docs template, a Hugo layout, the theme, or a package's YAML metadata.
+
+```bash
+make SKIP_VERSIONED_DOCS=1 api-docs/aws
+make serve
+```
+
+`make api-docs/<package>` reads `themes/default/data/registry/packages/<package>.yaml`, fetches the corresponding schema, and writes to `content/registry/packages/<package>/` at the repository root.
+
+`SKIP_VERSIONED_DOCS=1` skips generating the older-major-version snapshots, which requires a Pulumi-internal tool that most contributors can't install. Leave it set unless you are specifically working on versioned docs.
```
guineveresaenger, line 75: I don't understand this at all but it's tooling I'm unfamiliar with.

docs/previewing-registry-changes.md
```diff
@@ -0,0 +1,126 @@
...
+Every pull request against `pulumi/registry` gets a full site build published to a per-commit S3 bucket. CI maintains a single pinned comment on the PR containing:
+
+- the preview URL for the current commit, and
+- a **Changed pages** list linking directly to the pages your PR affects.
+
+The comment is updated in place on each build rather than added per commit, and the preview buckets are deleted when the PR closes.
+
+If your change is in a provider repo rather than here, you can still get a preview by opening a PR against this repository with the [branch schema](#local-preview-from-a-branch-schema) edits applied — the same YAML change works in CI. Don't merge that PR; it exists to produce the preview.
+
+For a community package pull request, a Pulumi maintainer can comment `/preview` to build a live preview of the package's pages.
```
guineveresaenger, line 116: It sounds like outside contributors don't get the automatic page preview. In light of that it seems odd to advertise that opening a PR will get you one - this is only true if you can flag down a maintainer.

docs/previewing-registry-changes.md
```diff
@@ -0,0 +1,126 @@
...
+
+Re-run step 3 after each schema change; the running Hugo server picks up the new files. If you're also changing CSS or JavaScript under `themes/default/theme`, use `make serve-all` instead so assets rebuild too.
+
+## Local preview of a published package
+
+Use this when the schema is already published and you're changing something on this side — a docs template, a Hugo layout, the theme, or a package's YAML metadata.
+
+```bash
+make SKIP_VERSIONED_DOCS=1 api-docs/aws
+make serve
```
guineveresaenger, line 70: Will this route to localhost as well?

docs/previewing-registry-changes.md
```diff
@@ -0,0 +1,126 @@
+# Previewing Registry Changes
+
+This guide covers how to see Pulumi Registry pages — including provider API docs — before they are published, whether the change lives in a provider's schema, in this repository's templates, or in a pull request.
```
guineveresaenger, line 3: This reads like slop, sorry.

docs/previewing-registry-changes.md
```diff
@@ -0,0 +1,126 @@
...
+
+Pick a route based on what you changed:
+
+| What you changed | Route |
```
guineveresaenger, line 7: "Route"? I think this can just be a ToC?

docs/previewing-registry-changes.md
```diff
@@ -0,0 +1,126 @@
...
+
+```bash
+mise trust && mise install
+make ensure
+make build-assets
+```
+
+See [the README](../README.md#using-this-repository) for details.
+
+Generated docs land in git-ignored directories (`content/` and `static/registry/` at the repository root), so nothing from these routes should ever end up in a commit.
```
guineveresaenger, line 26: do we need this? we're only talking about a preview here.

docs/previewing-registry-changes.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Local preview from a schema file
+
+Use this when you're changing a provider and want to see how its API docs will render before you cut a release. It reads a `schema.json` straight off your disk, so nothing needs to be pushed or published.
```
guineveresaenger, line 30:
```suggestion
Use this when you're changing a provider and want to see how its API docs will render before you cut a release.
```

docs/previewing-registry-changes.md
```diff
@@ -0,0 +1,126 @@
...
+
+1. Build the docs generator:
+
+    ```bash
+    make bin/resourcedocsgen
+    ```
+
+1. Generate your provider's schema in the provider repo. For a bridged provider that's usually `make generate_schema`, which writes `provider/cmd/pulumi-resource-<name>/schema.json`.
```
guineveresaenger, line 38:
```suggestion
1. Generate your provider's schema in the provider repo. For a bridged provider that's usually `make schema`, which writes `provider/cmd/pulumi-resource-<name>/schema.json`.
```

docs/previewing-registry-changes.md
```diff
@@ -0,0 +1,126 @@
...
+    ./bin/resourcedocsgen docs \
+        --schemaFile ../pulumi-aws/provider/cmd/pulumi-resource-aws/schema.json \
+        --version v9.9.9 \
+        --docsOutDir ./content/registry/packages/aws/api-docs \
+        --packageTreeJSONOutDir ./static/registry/packages/navs
+    ```
+
+    `--version` is required and must be valid semver. Use a dummy version higher than anything published so it's obvious in the rendered page that you're looking at a local build.
+
+1. The package's landing pages (`_index.md` and `installation-configuration.md`) are committed under `themes/default/content/registry/packages/<package>/` and are used as-is. If you're previewing a package that isn't in the registry yet, create that directory and add the two files by hand, copying them from your provider repo's `docs/` folder.
```
guineveresaenger, line 52:
```suggestion
1. The package's landing pages (`_index.md` and `installation-configuration.md`) are committed under `themes/default/content/registry/packages/<package>/` and are used as-is.
If you're previewing a package that isn't in the registry yet, create that directory and add the two files by hand, copying them from your provider repo's `docs/` folder.
```
(nit: each sentence should have a newline)
I....think resourcedocsgen actually does just generate these if they exist, but my memory may be wrong. Worth verifying.

docs/previewing-registry-changes.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+`make api-docs/<package>` reads `themes/default/data/registry/packages/<package>.yaml`, fetches the corresponding schema, and writes to `content/registry/packages/<package>/` at the repository root.
+
+`SKIP_VERSIONED_DOCS=1` skips generating the older-major-version snapshots, which requires a Pulumi-internal tool that most contributors can't install. Leave it set unless you are specifically working on versioned docs.
+
+To force a rebuild after a change the generator can't see (it caches on the package YAML and its own build), delete the sentinel file:
+
+```bash
+rm content/registry/packages/aws/api-docs/.generated
```
guineveresaenger, line 80: this seems wrong, this should be fixed via Make. Or you can also run `make -B` for `--always-make`.

docs/previewing-registry-changes.md
guineveresaenger: Was SKIP_VERSIONED_DOCS added in this PR?

guineveresaenger (commented):
Thank you, I think this is a good start.

Can we link to/from https://www.pulumi.com/docs/iac/guides/building-extending/packages/publishing-packages/#overview-installation--configuration?

There's also https://github.com/pulumi/registry/blob/master/docs/adding-a-new-package.md. Which one is the source of truth do we think?

Makefile
```diff
@@ -108,7 +108,7 @@ api-docs/%: .make/content/registry/packages/$$*/api-docs ;
 		--baseSchemasOutDir ./static/registry/packages \
 		--baseLLMDocsOutDir ./llm-docs-out/registry/packages \
 		$*
-	CONTENT_DIR=$(CURDIR)/content/registry/packages STATIC_DIR=$(CURDIR)/static/registry/packages ./scripts/generate-versioned-docs.sh $*
+	$(if $(SKIP_VERSIONED_DOCS),,CONTENT_DIR=$(CURDIR)/content/registry/packages STATIC_DIR=$(CURDIR)/static/registry/packages ./scripts/generate-versioned-docs.sh $*)
```
jkodroff, line 111: @guineveresaenger `SKIP_VERSIONED_DOCS` is added here b/c the registry mirror is a private repo, which means that contributors would not be able to use this guide.

docs/previewing-registry-changes.md
jkodroff: Yes - I left a comment on the line explaining why.

docs/previewing-registry-changes.md
jkodroff: Checked this one against the actual behavior and `make -B` doesn't get there — the skip isn't happening at the Make layer at all.

The `api-docs/%` target's sentinel (`.make/content/registry/packages/<pkg>/api-docs`) is an intermediate file, so Make deletes it at the end of every run — you can see `rm .make/content/registry/packages/random/api-docs` as the last line of the recipe. That means the recipe already re-runs unconditionally, and `-B` changes nothing. The skip is inside `resourcedocsgen`, which keeps its own cache keyed on the package YAML plus its build identity, recorded in `.generated`.

Verified against `random`:

- `make -B SKIP_VERSIONED_DOCS=1 api-docs/random` → `Skipping (output is fresh)`
- `rm content/registry/packages/random/api-docs/.generated && make SKIP_VERSIONED_DOCS=1 api-docs/random` → `Generating docs` / `Reading remote schema file from registry`

So deleting the sentinel stays the fix. I've rewritten the section to explain *why* it's there rather than just naming the file, and to say explicitly that `-B` won't help — that seemed like the more useful thing for the next person who hits it. The real bug is arguably that the cache key ignores the resolved schema; happy to file that as a follow-up if you agree.

docs/previewing-registry-changes.md
jkodroff: You're right, and the guide was overselling it. Confirmed the gating: the `preview` job in `pull-request.yml` is conditioned on `github.event.pull_request.head.repo.full_name == github.repository`, so a fork PR builds nothing, and `/preview` is restricted to `OWNER`/`MEMBER`/`COLLABORATOR`.

Rewrote the section to say so plainly — previews build for branches pushed to this repository, fork contributors should use the local routes above, and a maintainer can build one on demand with `/preview`. That way nobody opens a PR expecting a URL that never arrives.

jkodroff:
Thanks — took most of this. Answering the two questions from your review body:

**Cross-link:** added one from this guide out to [Overview, Installation & Configuration](https://www.pulumi.com/docs/iac/guides/building-extending/packages/publishing-packages/#overview-installation--configuration). I didn't add the reverse link from `adding-a-new-package.md` in this PR, just to keep the diff to one file.

**Source of truth:** I'd keep them separate rather than merge. `adding-a-new-package.md` is about getting a package into the registry — the YAML, the package list, the onboarding steps. This one is about seeing a change render before it ships, which applies whether or not the package is new. The overlap is mostly `make api-docs/<pkg>`. Consolidating them properly means deciding what `docs/` is for as a whole, which I'd rather do as its own pass than smuggle into this PR — happy to file an issue for it.

Also rebased onto master, so the landing-page step now reflects #12341: only `_index.md` is required, and `installation-configuration.md` is the optional split.

One suggestion I pushed back on — `make -B` for the rebuild cache; details in that thread, short version is the Make sentinel is an intermediate file that gets deleted every run, so the skip is happening inside `resourcedocsgen`, not in Make.

docs/previewing-registry-changes.md
jkodroff: Filed the underlying cache bug as #12454 — laid out three possible fixes there rather than picking one, since the obvious fix (hash the schema into the key) gives up the network savings the cache exists for. This PR keeps documenting the sentinel workaround in the meantime.

docs/previewing-registry-changes.md
jkodroff: Not sure what you mean by "route" here.

CamSoper (approved):

jkodroff:
This is probably not in its final form but it should be a significant improvement. Expect follow up PRs to ensure that we have clear and complete docs here.

Issue #12252, opened 2026-08-26
Upstream Overview-page bugs found during the tier B audit


jkodroff:
Tracking issue for the upstream bugs surfaced while working #12226 (tier B). These are **not** tier B work and none of them block it — the tier B PR deletes duplicated `installation-configuration.md` pages in this repo, whereas everything below is a defect in the *generated Overview* (`docs/_index.md`, owned by the provider repo and re-fetched here on every release). Nothing in `pulumi/registry` can fix any of it.

## What the audit changed

The per-provider list in https://github.com/pulumi/registry/issues/12226#issuecomment-5429165241 reads as six independent provider bugs. Verification says otherwise: most are a handful of `pulumi-terraform-bridge` docsgen defects reproducing across a large fraction of the registry. Filing them per-provider would have put the same bug in front of six teams, none of whom own the fix.

Counts are over the 310 Overview pages in `themes/default/content/registry/packages/`:

| defect | pages affected |
|---|---|
| `Pulumi.yaml` emitted with an empty `runtime:` (invalid — won't run) | **82** |
| `value: 'TODO: var.x'` left inside copy-pasteable config | **63** |
| `Example currently unavailable in this language` shipped inside an hcl code fence | **27** |
| Sensitive config written as plaintext `value:` rather than `secure:` | vsphere, cloudflare (+) |
| upstream YAML front matter leaked as a `## page_title:` heading | 2 (dnsimple, ec) |

## Bridge issues — the high-leverage ones

Each of these regenerates dozens of Overview pages when fixed.

- [ ] pulumi/pulumi-terraform-bridge#3600 — generated `Pulumi.yaml` config blocks have an empty `runtime:` (82 pages)
- [ ] pulumi/pulumi-terraform-bridge#3601 — unconvertible TF interpolations become `value: 'TODO: var.x'` inside config values (63 pages)
- [ ] pulumi/pulumi-terraform-bridge#3602 — **Sensitive provider config emitted as plaintext `value:` instead of `secure:`** — the docs currently teach users to commit secrets in plaintext
- [ ] pulumi/pulumi-terraform-bridge#3603 — "Example currently unavailable in this language" shipped inside an hcl code fence (27 pages)
- [ ] pulumi/pulumi-terraform-bridge#3604 — comments inside resource bodies dropped, turning upstream `# ...` elisions into apparently-complete broken examples
[...]

Issue #12237, opened 2026-08-26
Guard redirects.js against the CloudFront Function size limit in PR CI


jkodroff:
Raised in review on #12231. Nothing is broken today — this is a guard against a failure mode that currently has no test and surfaces at the worst possible moment.

## The ceiling

`infrastructure/index.ts:180-183` ships `redirects.js` verbatim as CloudFront Function code, with no minification:

```ts
const redirectFunction = new aws.cloudfront.Function("legacy-version-redirects", {
    runtime: "cloudfront-js-2.0",
    code: fs.readFileSync(`${__dirname}/redirects.js`, "utf-8"),
});
```

AWS caps CloudFront Function code at **10 KB**. Nothing in the repo checks this — no size assertion in `redirects.test.js`, none in `infrastructure/`.

## Why it fails badly

PR CI never touches `infrastructure/` (the only mention in `pull-request.yml` is `infrastructure/yarn.lock` as a cache key). The Pulumi program runs only on push to master, via `scripts/ci/run-pulumi.sh update` → `pulumi -C infrastructure update --yes`, and it runs **after** the full site build and S3 sync — it is the step that reads `origin-bucket-metadata.json` and swaps the CloudFront origin to the newly built bucket.

So an oversized `redirects.js` merges green, then fails post-merge at the end of a 30-60 minute build, at the one step that makes the new content live. Content is synced to the new bucket, the origin never swaps, and the deploy is left half-applied. `testing-deploy.yml` would not catch it first — it is `workflow_dispatch` only.

## Current headroom

| | bytes | % of 10 KiB |
|---|---|---|
| before #12231 | 4,245 | 41% |
| after #12231 | 5,589 | 55% |

#12231 grew the file 32% in one change, and roughly 1.1 KB of that was comment prose rather than rules.

Projecting the rest of the #11866 tier work, which is the near-term growth driver:

- If tiers B (14 packages) and C (24 packages) are folded into the **existing** alternation: +314 bytes → 5,903 (58%). Fine.
- If each tier lands as its own rule with its own rationale comment, the way tier A did: +2.7 KB → 8,277 (**81%**). Uncomfortable.

The rules are cheap; the explanatory comments are what consume the budget.

## Ask

Assert the size in `infrastructure/redirects.test.js`. That file already runs in PR CI via `make test-infra` (`.github/workflows/pull-request.yml:152-166`), already imports `fs` and `path`, and runs offline — so this needs no new job and no new dependency:

```js
[...]

PR #12253, opened 2026-08-26
Delete the Install & config pages that substantially duplicate the Overview

### LINKED ISSUE #12226: Reconcile the Install & Config pages that substantially duplicate the Overview (tier B)
Part of #11866. Follows the audit in #12209.

Two or more duplicated sections, or ≥40% of the page's prose, shared with the Overview — but each page also has content of its own. Unlike tier A these cannot be deleted sight-unseen.

## The pattern

`installation-configuration.md` says substantially the same thing as `_index.md`, and **`pulumi/registry` is the source of truth for that page** — it is hand-written in this repo, carries no `# WARNING: this file was fetched from` provenance header, and no automation has ever touched it or ever will.

That is what makes this set actionable on its own: the Overview beside it is regenerated from the provider repo on every release, so the duplicate drifts further out of date with each one and nothing in the pipeline will ever correct it.

Matching is on section-level similarity, not exact text — most of these forked from the same content in Aug 2024 and have diverged in wording since while staying the same page. Verified against `master` @ `37892ade3`.

**14 packages** are in this tier. The other two tiers are #12225 (wholly duplicated) and #12227 (install list only), and the full audit is #12209.

## Packages

| package | duplicated sections | config page last changed | overview last changed |
|---|---|---|---|
| `azuredevops` | Example usage, Installation | 2024-08-26 | 2026-07-16 |
| `cloudamqp` | Config reference, Example usage | 2024-08-26 | 2026-06-25 |
| `cloudflare` | Example usage, Installation | 2024-08-23 | 2026-08-06 |
| `dnsimple` | Config reference | 2024-08-26 | 2026-08-04 |
| `docker` | Installation | 2024-06-20 | 2026-06-26 |
| `kafka` | Installation | 2023-11-28 | 2026-07-23 |
| `linode` | Installation | 2023-11-28 | 2026-08-05 |
| `meraki` | Installation | 2024-05-24 | 2026-07-23 |
| `nomad` | Installation | 2023-11-28 | 2026-04-20 |
| `okta` | Installation | 2023-11-28 | 2026-08-05 |
| `postgresql` | Installation | 2023-11-28 | 2026-07-16 |
| `spotinst` | Installation | 2023-11-28 | 2026-08-15 |
| `vault` | Installation | 2023-11-28 | 2026-08-15 |
| `vsphere` | Installation | 2023-11-28 | 2026-06-25 |

## Scope

**This tier needs a read-through per package, not a blanket deletion.** For each: identify the unique content, decide whether it belongs on the Overview, move it if so, then delete the page.

The duplicated sections listed above are safe to drop in every case — they are regenerated into the Overview on each release, so the copy here is by definition the stale one.

Worth considering before investing much hand-editing: #12183 (templated installation instructions) and #12185 (templated provider config) will replace this content wholesale. If those land first, most of what is unique here becomes generated output and the hand-reconciliation is wasted. Sequencing this tier after them is defensible; sequencing tiers A and C before them is not, since nothing there is worth preserving.

## Upstream is clear for this tier

Checked every package here for `docs/installation-configuration.md` in its provider repo at the currently released tag: **all return 404.** No upstream file exists to re-create these pages, so deleting in `pulumi/registry` is sufficient and durable.

## Re-check upstream when the copy bug is fixed

`installation-configuration.md` is not reliably copied from upstream today — that is the underlying bug, and it is why these pages froze while their Overviews kept moving. **The upstream check has to be re-run as part of fixing it.** The clean 404 sweep above describes today's behaviour, not a guarantee: once the copy path works, any provider repo that does ship `docs/installation-configuration.md` will have that content flow into the registry, and if the upstream file is itself a duplicate of the provider's own `docs/_index.md` then deleting the registry copy achieves nothing — it returns on the next release, and the deletion has to happen **upstream in the provider repo** instead.

That shape already exists for 16 packages (`heroku` `ise` `scm` `sdwan` `harness` `junipermist` `slack` `ovh` `cpln` `vercel` `bytepluscc` `checkly` `descope` `formal` `nvidia-aicr` `volcenginecc`), which duplicate the Overview from an upstream-owned config page. They are out of scope here and need upstream PRs — see #12209.

## Blocked on redirects

Deleting the page 404s a live URL. Hugo `aliases:` on the Overview is the right mechanism, but it cannot be hand-added: `tools/resourcedocsgen/cmd/metadata.go` overwrites `_index.md` wholesale on every release, and every package in this tier has a generated Overview, so a hand-written alias is silently erased at the next release. The alias has to be emitted by the generator. Tracking separately.

jkodroff:
Fourteen packages carry an Install & config page whose sections already appear on the Overview beside it. This is tier B of #11866, following #12231 (tier A).

Deleted for: `azuredevops`, `cloudamqp`, `cloudflare`, `dnsimple`, `docker`, `kafka`, `linode`, `meraki`, `nomad`, `okta`, `postgresql`, `spotinst`, `vault`, `vsphere`.

Unlike tier A, #12226 calls for a read-through per package rather than a blanket deletion, since each of these pages has content of its own. Every page was read; the unique content turned out to be obsolete rather than merely redundant, with one exception noted below.

## Why deletion is safe here

All fourteen are hand-written in this repo — none carries a `# WARNING: this file was fetched from` header, and no upstream `docs/installation-configuration.md` exists for any of these providers. The URLs are gone for good rather than pending regeneration. Nothing in the repo links to them (`layouts`, `data`, `tools`, `scripts` all checked).

**The four pages with their own Configuration Reference**, checked against the live provider schemas:

| package | finding |
|---|---|
| `azuredevops` | Overview's 19 options are an exact match for the v3.16.0 schema. The deleted page listed 8 more (`oidcGithubActions`, `tenantIdApply`, `clientIdPlan`, …) — all gone from the schema. |
| `cloudflare` | Same shape: Overview matches v6.20.0 exactly (6 options); the deleted page listed 7 v5-era options removed in the v6 rewrite, and omitted the real `baseUrl`. |
| `cloudamqp` | Identical option sets. |
| `dnsimple` | Identical sets (`token`, `account`, `sandbox`, `prefetch`, `userAgent`); the Overview's wording is fresher. |

Worth flagging, because #12226 originally said to keep whichever list was more complete: for `azuredevops` and `cloudflare` the more complete list is the wrong one. Doing that would have imported 15 nonexistent config options. Corrected in [a comment on #12226](https://github.com/pulumi/registry/issues/12226#issuecomment-5430775420).

**The ten smaller pages** are an install list plus credential prose. Every config key and environment variable on them (`PGHOST`, `VAULT_ADDR`, `LINODE_TOKEN`, `OKTA_API_TOKEN`, `VSPHERE_*`, `NOMAD_*`, `SPOTINST_*`, `merakiDashboardApiKey`, …) is documented on the corresponding Overview.

**These pages were also actively wrong.** Eight advertised a stale Go SDK major on the very page a user opens in order to install — `vault`'s pointed at `sdk/v4`, a module last released in November 2021, while the Overview is on `sdk/v7`. Old majors stay on the module proxy forever, so `go get` succeeds and silently yields a years-old SDK.

## The one exception: docker

`docker/installation-configuration.md` uniquely documented the `DOCKER_HOST` environment variable and the Windows named-pipe host form:

```powershell
pulumi config set docker:host "npipe:////.//pipe//docker_engine"
```

The Overview names `DOCKER_CONTEXT` for `context` but no variable for `host`, and its Remote Hosts section covers `ssh://` only. `resourcedocsgen` rewrites each `_index.md` wholesale on every release, so this cannot be moved to the Overview from here — it is filed upstream as pulumi/pulumi-docker#1776 and tracked in #12252. The page is deleted rather than held back, since its install list still pointed Go users at `sdk/v4`.

## Redirects

Deleting a page 404s a live URL. These packages join the existing rule in `infrastructure/redirects.js`, the CloudFront viewer-request function established as the durable mechanism in #12231 — Hugo `aliases:` can't be used because `resourcedocsgen` erases front matter on every release.

Adds a test that the `docker` and `kafka` alternatives don't swallow `docker-build` and `kafka-connect`, which are real packages that still have an Install & config page.

## Verification

- `node --test infrastructure/redirects.test.js` — 36/36 pass
- `make lint` — clean (golangci-lint, shortcode delimiters, markdown, dark logos, eslint/prettier)
- `redirects.js` is 5,973 bytes against the 10,240-byte CloudFront Function limit (relevant to #12237)
- Spot-checked every URL form: trailing slash, bare, `.md`, and `/index.html` all 301 to the package Overview; `docker-build`, `kafka-connect` and `aws` pass through untouched

Fixes #12226

https://claude.ai/code/session_01TAtonT1zv4mmaxj8Sdn6nf


CamSoper (approved):
This all looks fine to me. That Windows pipe situation is something that a LLM should be able to figure out in the absence of the updated doc.

PR #12279, opened 2026-08-28
Add matthiashamacher/pulumi-ohdear community package

matthiashamacher:
A Pulumi provider for Oh Dear (https://ohdear.app) website monitoring, covering monitors, status pages, tags/tag groups, and notification destinations.

PR #12177, opened 2026-08-21
Add Multipass provider

jagedn:
Adds IncSteps/pulumi-provider-multipass provider to the registry and community package list.

PR #10996, opened 2026-05-13
Add dagsterplus provider package

tehsis:
## Summary
Adds the `dagsterplus` community provider package (`dagster-io/pulumi-dagsterplus` v0.0.1).

Re-push of #10932 to `pulumi/registry` (originally from a fork at `dehume/registry`) so the `Test Live Registry Publish` check can authenticate to ESC — `pull_request` runs from forks don't receive `id-token: write` and can't mint an OIDC token. Authorship preserved.

Original author: @dehume.

## Test plan
- [ ] All required checks pass, including `Test Live Registry Publish`
- [ ] `Sentinel Tower` reports success

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12307, opened 2026-08-31
[registry] Let the community package check post its own fact-sheet

fnune:
[#12177](https://github.com/pulumi/registry/pull/12177) stalled in August. Its fact-sheet comment stopped changing on 2026-08-25, and three of the four failures it still shows are wrong. A contributor is meant to get this feedback early, without asking us for anything. Approving the rest of CI stays our click.

[`community-package-check.yml`](https://github.com/pulumi/registry/blob/master/.github/workflows/community-package-check.yml) reads a contributor's package and writes a fact-sheet. [`community-package-sweep.yml`](https://github.com/pulumi/registry/blob/master/.github/workflows/community-package-sweep.yml) starts it, because GitHub will not start a workflow on a fork pull request from a first-time contributor.

| Defect | Cause | Change |
|---|---|---|
| Nothing posted the fact-sheet | The sweep dispatches with the workflow token, so the actor is `github-actions[bot]`, and GitHub raises no `workflow_run` from a bot actor. The reporter waited for that event. Runs [1](https://github.com/pulumi/registry/actions/runs/33173786096), [2](https://github.com/pulumi/registry/actions/runs/33176390178), [3](https://github.com/pulumi/registry/actions/runs/33183129647) and [4](https://github.com/pulumi/registry/actions/runs/33378230807) posted nothing | The check posts its own fact-sheet, from a second job that holds a token and runs no package. No event, no wait. `community-package-report.yml` is deleted |
| The check used the wrong package name | `--providerName` falls back to the repository name, so `pulumi-provider-multipass` became `provider-multipass`, not `multipass`. It is also wrong for `defang-aws`, `defang-gcp`, `defang-azure` and `kafka-connect` | Pass the schema name, as [`generate-package-metadata.yml` already does](https://github.com/pulumi/registry/blob/master/.github/workflows/generate-package-metadata.yml#L152) |
| The check always reported a missing publisher | It read `publisher-names.json` from `master` | `fetch-pr` copies both allowlisted files from the pull request head |
| `v0.1.0` never matched `v0.1.0` | The code stripped the `v` from the tag only | Strip both sides |
| The contributor saw the tail of a stack trace, not the error | The check kept the last 600 characters, and the message sits on line 1 | Keep the first 600 |
| A crashed check blanked the fact-sheet comment | It posted an empty body and exited 0 | Post a link to the run. The check exits 2 with a fact-sheet that names itself as the fault |

The check is now dispatch-only, so there is one way to start it and one way to report it. That deletes `community-package-report.yml`, the heuristic that guessed a pull request from a branch name, the sweep's question about whether GitHub started a run, and every `pull_request.number || inputs.pr` fallback in the check.

Only one step in that workflow holds a token now, and it is the gate: it refuses a pull request that changes anything outside `package-list.json` and `publisher-names.json`, so the step that runs the contributor's package never starts. The checkout no longer persists credentials, which used to leave a usable token in `.git/config` for every later step. Three tests hold that shape, and each one fails if I undo the thing it guards.

If the sweep itself fails, it opens one `p1` issue, which the daily digest in `priority-digest.yml` already surfaces. It reuses the open issue instead of opening another, so a sweep that stays broken does not file one every five minutes.

The sweep dispatches once per head, so a pull request that already has a run needs `/check` once after this lands.

## Test plan

- 88 unit tests and `mypy --strict` in `scripts/ci/community-package/`
- `resourcedocsgen metadata from-github` on `incsteps/pulumi-provider-multipass@v0.1.0` fails on the name. With `--providerName multipass` it passes the name and fails on the absent `docs/_index.md`, which is the contributor's one real bug
- `cli.py fetch-pr` against the live API: #12177 and #12279 pass the gate and copy exactly the two files. #12307, this pull request, is refused and copies nothing
- `cli.py report` against the live API finds the sticky comment on #12177 and edits it in place
- `cli.py sweep` against the live API dispatches for #10996 and skips the heads that already ran
- The whole workflow, dispatched from this branch against a throwaway pull request ([#12308](https://github.com/pulumi/registry/pull/12308), now closed). The fact-sheet arrived with no `workflow_run` event and read `publisher listed` ✅, `schema version` ✅, and `docs generate` ❌ naming the absent `docs/_index.md` on its first line. A second commit adding a file outside the allowlist made the gate refuse: the step that runs the package reported `skipped`, and the reporter edited the same comment instead of adding one


.github/workflows/community-package-sweep.yml
```diff
@@ -30,3 +34,25 @@ jobs:
...
+        uses: pulumi/esc-action@32dda43ffe7a2ab21e6bce11c781b44a3e5623cf # v3
+      - name: Slack Notification
+        uses: docker://sholung/action-slack-notify:v2.3.0
+        env:
+          SLACK_CHANNEL: registry-ops
+          SLACK_COLOR: "#F54242"
+          SLACK_MESSAGE: "community package sweep failure in pulumi/registry :meow_sad:"
+          SLACK_USERNAME: registrybot
+          SLACK_WEBHOOK: ${{ steps.esc-secrets.outputs.SLACK_WEBHOOK_URL }}
+          SLACK_ICON: https://www.pulumi.com/logos/brand/avatar-on-white.png
```
borisschlosser: Can we notify Slack channels by id instead of pre-defined url or by channel name?

borisschlosser: As an alternative we could create a p0/p1 issue which notifies daily already

fnune: Ah yes, doing the issues, otherwise we'll get spammed.

borisschlosser (approved):

PR #12308, opened 2026-08-31
[TEST - DO NOT MERGE] CI test fixture for #12307

fnune:
**This is a test fixture, not a contribution. Do not merge and do not review.**

It reproduces the exact two-file diff of #12177 so that the reworked community package check in #12307 can be dispatched against a pull request we own, instead of a contributor's.

Expected result once #12307's workflow runs here:

- the fact-sheet arrives with no `workflow_run` event, posted by the check's own second job
- `publisher listed` passes, because `fetch-pr` now copies `publisher-names.json` from the head
- `schema version` passes, because the `v` is stripped from both sides
- `docs generate` fails on the absent `docs/_index.md` upstream, and says so on the first line instead of showing the tail of a Go stack trace

I will close this once #12307 is verified.

### DIFF
--- a/TEST-DELETE-ME-outside-allowlist.md
+++ b/TEST-DELETE-ME-outside-allowlist.md
@@ -1,0 +1,1 @@
+This file exists to prove the community package check refuses a PR that touches anything outside the allowlist. Delete me.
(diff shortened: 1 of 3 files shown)


fnune:
Test fixture for #12307. Verification is done: the reworked check posted a correct fact-sheet with no `workflow_run` event, and the allowlist gate refused the second commit before the package step ran. Closing.

PR #12341, opened 2026-09-01
Require only the Overview page, not Install & Configuration

### LINKED ISSUE #12247: Only overview/index is required, not install & config
We say in the README that we require 2 pages: Overview (_index.md) and Install and Config. However, only 1 page is actually required in terms of what's needed to communicate the necessary information to the reader.

In order to clean this up:

- Make sure only 1 page is required in the code
- Make sure only 1 page is indicated as required in any docs in this repo
- Create an issue to indicate that only 1 page is required in pulumi/docs

jkodroff:
# Josh Says

The existing registry "required" pages (Overview, Install & Config) are inconsistent. This PR is an attempt to clarify our documentation about what it actually required (both in the code and our documentation):

- We only need the Overview (index.md) page. Most providers who have/had Install & Config do/did not need it: It was either an unnecessary split between 2 pages, or it duplicated information already on the Overview page.
- The instructions for what's required should only live in one place, and they should (obviously) agree with code that validates what's required.

# Claude Says

Fixes #12247

## What

Our contributor-facing docs told package authors they must supply **two** pages — `docs/_index.md` and `docs/installation-configuration.md`. Only one is actually needed. An Overview can carry the SDK links and configuration instructions perfectly well, and **130 of the 312** packages in the registry today ship exactly that.

The code already agreed. Nothing in this repo hard-fails on a missing install page:

- `resourcedocsgen` marks only `--indexFileURL` required and emits the install page only when a URL/file is present.
- `scripts/ci/validate-packages.sh` errors on a missing `_index.md`, but only `WARN`ed on a missing install page.
- `verify_entry.py` gates the community-package fact-sheet's green status on `_index.md` alone.
- `package-nav.html` and `package.llmsitemap.json` guard the nav entry behind an existence check.
- `infrastructure/redirects.js` goes further and 301s 25 packages' *deleted* install pages to their Overview (#12225, #12226).

So this is mostly a prose change, bringing the docs in line with behavior that already ships.

The framing kept throughout: **one page is required; the information is not optional.** SDK links for each published language and credentials/environment setup are still required — they just live in `_index.md` unless an author chooses to split them out.

## Docs

| File | Change |
|---|---|
| `docs/overview-page.md` | **New.** The single place that says what an Overview page must contain: front-matter keys, and the `## Installation` / `## Example Usage` / `## Configuration` sections. Everything else now links here instead of restating it, which is what stops the two copies drifting apart again. |
| `README.md` | The author instructions were a duplicate of `docs/adding-a-new-package.md` that had drifted. Collapsed to a short summary plus a pointer. |
| `docs/adding-a-new-package.md` | Absorbs the author-facing steps, so contributor and maintainer instructions live in one file. The maintainer checklist's two install-page items become one item pointing at the Overview guidelines. |
| `.github/PULL_REQUEST_TEMPLATE/community-package.md` | The pre-submission checkbox names one required file and links the Overview guidelines. |
| `AGENTS.md`, `.claude/commands/_common/review-criteria.md` | Reworded so a reviewing agent doesn't flag a single-page package as incomplete. Commit policy is unchanged. |
| `BUILD-AND-DEPLOY.md` | Line 945 corrected alongside the `push-registry.py` fix; see the separate note below on the publication section. |

`BUILD-AND-DEPLOY.md:738` already said the Overview is the only required page, and was left alone as the precedent for the wording elsewhere.

## Code

Three changes the audit surfaced:

**`tools/resourcedocsgen/cmd/metadata.go`** — the one place in Go that appeared to declare a page requirement, and it was inert: `metadata from-github` declared a `required` field on its `remoteFiles` table that was never read. The real requirement lived somewhere less obvious and cut the wrong way.

`readRemoteFile` only tolerates a 404 when the repo owner is literally `pulumi`. For any other owner a missing `docs/installation-configuration.md` came back as an **error**, and the loop propagates it immediately — the `len(content) == 0 { continue }` skip below it was unreachable. So a community provider that followed the instructions this PR writes and shipped a single Overview would have hard-failed `metadata from-github`, taking the fact sheet red (`verify_entry.py` gates `green` on it) and the nightly matrix job with it.

Nothing in the registry hits this today. I probed all 100 community repos in `package-list.json` at the version each is pinned to: every one publishes both pages, and the only 404s are `pulumi`-owned repos already covered by the escape hatch. So it is a latent defect — but the docs half of this PR invites exactly the shape that trips it.

So the `required` field comes back, this time load-bearing: `readRemoteFile` takes it as a parameter and treats a 404 on a non-required file as a skip for every owner. The pulumi-only tolerance of a missing `_index.md` is untouched. `metadata_test.go` gains a test over the three cases; it fails if you revert the condition.

**`scripts/ci/push-registry.py`** — the existence check tested for `_installation-configuration.md` (leading underscore), a filename that exists nowhere in the tree, so `--installation-configuration` has **never** been passed to `pulumi package publish`. Dropping the underscore makes the conditional do what it was written to do.

Scope, because an earlier revision of this description overstated it: `push-registry.py` is no longer the primary publisher. `push.yml` runs `publish_to_registry.py` first and reaches `push-registry.py` only `if: steps.mirror-publish.outcome == 'failure'`. So nothing changes on `master` unless the mirror publisher fails first. And what gets sent is the **Pulumi Cloud registry entry** — the store behind `pulumi package add` — not the rendered pages; no site content changes either way.

Verified the command line rather than trusting it, since this branch has never fired: `pulumi package publish --help` on v3.258.0 lists `--installation-configuration string`. The adjacent `--source` flag the script passes is hidden — absent from `--help` — but it parses; an unknown flag errors with `unknown flag: …`, and `--source pulumi` gets through to the arg-count check.

**`scripts/ci/validate-packages.sh`** — dropped the `WARN` for a missing install page. It fired for 130 of 312 packages on every run: ~130 lines of noise about a condition that is now explicitly fine. The `_index.md` existence and 250-character checks above it are untouched.

## Stale publication docs, fixed while I was in there

Checking the `--installation-configuration` flag turned up a `BUILD-AND-DEPLOY.md` section describing a pipeline that no longer exists. Not caused by this PR — it went stale when `publish_to_registry.py` landed — but it is the section this PR already edits, so it is fixed here.

It claimed `push-registry.py` runs "On every push to `master`" (it is the fallback, gated on the primary failing) and "In dry-run mode on every PR" (that job runs `publish_to_registry.py --validate-all`; `push-registry.py` does not run on PRs at all). The primary publisher was undocumented. The same claim also appeared in the table of contents, two workflow diagrams, section 12, and `AGENTS.md`.

The rewritten section also states the distinction that made the flag question confusing in the first place: publication writes the **Pulumi Cloud registry entry**, a different store from the Hugo pages the deploy syncs, and the two can drift.

## Verification

- `cd tools/resourcedocsgen && go build ./... && go test ./... && golangci-lint run` — all pass, 0 issues
- `make lint-markdown` — 159 files, 0 errors
- `bash scripts/ci/validate-packages.sh` — exits 0 with **zero** `WARN` lines
- Temporarily renamed `logfire/_index.md` and confirmed the validator still emits `ERROR:` and exits 1, then restored it

On the `readRemoteFile` fix:

- The new `TestReadRemoteFileNotFound` fails on the "optional file, community owner" case if you revert the condition to `repoOwner == "pulumi" && …`, so it is testing the thing it claims to.
- End-to-end: `resourcedocsgen metadata from-github --repoSlug DefangLabs/pulumi-defang --schemaFile provider/cmd/pulumi-resource-defang-aws/schema.json --version v2.6.0 --providerName defang-aws` exits 0 and emits both pages, unchanged.
- Probed all 100 community repos in `package-list.json` at each package's pinned version: every non-`pulumi` repo returns 200 for both `docs/_index.md` and `docs/installation-configuration.md`. That is what makes the defect latent rather than live — and also why there is no real package to reproduce it against, hence the unit test.

On the publish path:

- `pulumi package publish --help` on v3.258.0 lists `--installation-configuration string`; `--source` is hidden but parses (verified against an `unknown flag` control).
- The old underscored check returns `False` for `aws`, `logfire`, and `random` alike; the corrected check returns `True` for `aws` and `logfire` and `False` for `random` (one of the packages whose page was deleted in #12225).

No layout, partial, or styling changes, so no dark-mode pass applies.

## Follow-up

The third bullet of #12247 asks for a companion issue in `pulumi/docs`, where `publishing-packages.md` states the same two-page requirement. Filed separately and linked from the issue.

https://claude.ai/code/session_01VM9ZNx8n7giJNQVuGWtmzz


CamSoper (approved):
This is an overdue simplification, let's ship it.

jkodroff:
Addressed in 0f5ab67.

**🔴 The one-page promise** — you're right about the mechanism, and it's fixed. `readRemoteFile` now takes `required`, and a 404 on a non-required file is a skip for every owner rather than only `pulumi`; the pulumi-only tolerance of a missing `_index.md` is untouched. The `required` field is back on `remoteFiles` and load-bearing this time. `TestReadRemoteFileNotFound` covers the three cases and fails if you revert the condition.

Two corrections to the supporting evidence, in both directions:

- **`defang` is not an example of this.** It publishes both pages at its release (`v2.6.0` → 200 for each). I made the same mistake checking it, against a tag that doesn't exist, which 404s for a different reason.
- **No package in the registry hits this today.** I probed all 100 community repos in `package-list.json` at each package's pinned version: every non-`pulumi` one returns 200 for both docs. Every 404 is a `pulumi`-owned repo already covered by the escape hatch. So this is latent, not live — nothing is currently failing.

That doesn't change the fix, it changes why: the docs half of this PR invites the shape that trips it, so the code should be right before a contributor is the one who finds out.

**🟡 `--installation-configuration`** — verified. `pulumi package publish --help` on v3.258.0 lists `--installation-configuration string`. CI installs the CLI unpinned via `pulumi/actions@v7`, so it gets that or newer. Worth noting the neighbouring `--source` the script also passes is *hidden* — absent from `--help` — but it parses; an unknown flag errors with `unknown flag: …` while `--source pulumi` reaches the arg-count check.

Your blast-radius correction was right and the description now says so. It also now says what an earlier revision didn't: the content goes to the **Pulumi Cloud registry entry**, not the rendered pages.

That question turned up more staleness in the same section. `BUILD-AND-DEPLOY.md` still described `push-registry.py` as running on every push to `master` and in dry-run on every PR; neither has been true since `publish_to_registry.py` landed, and the primary publisher was undocumented. Same claim in the TOC, two diagrams, section 12, and `AGENTS.md`. Fixed in the same commit — not this PR's doing, but it's the section this PR already edits.

**Nits** — both are against wording that no longer exists; the branch was force-pushed after your review. `README.md`'s author instructions were a drifted duplicate and are now a pointer, and the maintainer checklist's install-page items point at the new `docs/overview-page.md`.

jkodroff:
@claude Review this again

PR #12345, opened 2026-09-02
Match registry search against package name and keywords

### LINKED ISSUE #4902: Strata Cloud Manager Provider - Make the Provider Searchable by More Terms

iwahbe:
The filter box on the registry index page matched only the package title. Strata Cloud Manager was not found by `scm` or `palo alto`, although the schema declares those keywords.

Two commits:

1. **Match registry search against package name and keywords.**
   - `resourcedocsgen` copies the schema keywords into the package YAML as `keywords`, without `pulumi` and without `category/` and `kind/` tags.
   - The package card exposes `data-name` and `data-keywords`.
   - The filter splits the query on whitespace and requires every token to appear in the title, the name, or the keywords. `palo alto` matches `paloaltonetworks`.
   - `docs/adding-a-new-package.md` explains how to add search terms.
2. **Backfill package keywords from each schema.** Every package YAML with a `schema_file_url` gets the keywords the tool would emit, so the preview shows the new search across the whole registry. The output was checked byte for byte against `resourcedocsgen` for six packages.

The `prisma` term arrives with pulumi/pulumi-scm#606. The hard-coded AWS and GCP synonyms stay for now; #12346 tracks their removal.

Fixes #4902.

### DIFF
--- a/cypress/e2e/site.cy.js
+++ b/cypress/e2e/site.cy.js
@@ -42,4 +42,51 @@ describe("www.pulumi.com/registry", () => {
             cy.get(".docs-main-nav").should("have.attr", "style").and("match", /height:\s*\d+px/);
         });
     });
+
+    describe("package list filter", () => {
+        beforeEach(() => {
+            cy.visit("/registry/");
+        });
+
+        const filterBy = (query) => {
+            cy.get(".registry-filter-input").clear().type(query);
+            // The search component debounces its event. The random package matches
+            // none of the queries below, so wait until it is hidden before asserting
+            // on the result.
+            cy.get(".all-packages .package.hidden [data-name='random']").should("exist");
+        };
+
+        const expectShown = (name) => {
+            cy.get(`.all-packages .package:not(.hidden) [data-name='${name}']`).should("exist");
+        };
+
+        it("matches the package name", () => {
+            filterBy("scm");
+            expectShown("scm");
+        });
+
+        it("matches every token against the package keywords", () => {
+            filterBy("palo alto");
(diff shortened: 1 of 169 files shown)


fnune (approved):
I don't love how this search works but that exists before this PR 😄 this seems like a positive extension. I wonder whether the frontend could be tested.

iwahbe:
> I don't love how this search works but that exists before this PR 😄

I agree. My first Claude prompt was "can we switch this to use Algolia", but it turns out much more complex.

> I wonder whether the frontend could be tested.

I think so... adding that now.

cypress/e2e/site.cy.js
```diff
@@ -42,4 +42,51 @@ describe("www.pulumi.com/registry", () => {
             cy.get(".docs-main-nav").should("have.attr", "style").and("match", /height:\s*\d+px/);
         });
     });
+
+    describe("package list filter", () => {
+        beforeEach(() => {
+            cy.visit("/registry/");
+        });
```
fnune, line 49: Sweet!

PR #12350, opened 2026-09-02
Render {{% ref %}} shortcodes per-language in provider docs

Frassle:
Provider schemas now emit `{{% ref DEST %}}` shortcodes to reference other schema entities (added in https://github.com/pulumi/pulumi/pull/21369), but resourcedocsgen was passing them through verbatim, leaving `{{% ref ... %}}` visible in the rendered registry pages.

Resolve each ref by delegating to every language's `DocLanguageHelper.ResolveDocRef` via `schema.PackageReference.InterpretPulumiRefs`. When all languages produce the same name (e.g. an unqualified property), emit it as plain text; otherwise wrap the per-language names in inline `<pulumi-choosable>` elements so the registry frontend shows each language its own rendering (e.g. Go's `module.ResourceArgs.IntegerProp` vs Python's `_module.ResourceArgs.integer_prop`).

Applied from both `decomposeDocstring` and `processDescription` (covers descriptions and import-details sections). Malformed refs are left as-is so the failure is visible in the rendered docs.


Frassle:
@claude do another review

Frassle:
@claude fix up the points you raised

tools/resourcedocsgen/pkg/docs/refs.go
```diff
@@ -0,0 +1,201 @@
...
+		return description
+	}
+	pkgRef := dctx.pkg.Reference()
+	rendered, err := pkgRef.InterpretPulumiRefs(description, func(ref schema.DocRef) (string, bool) {
+		return substitute(pkgRef, ref), true
+	})
+	if err != nil {
+		slog.Warn("interpreting pulumi refs; leaving description unchanged",
+			"pkg", dctx.pkg.Name, "err", err)
+		return description
```
fnune, line 114: Can we avoid bailing for the whole description and maybe do a prepass? Otherwise, a parse error will result in the whole description rendering `{{% ref...`, which apparently Hugo will just hide by default.

tools/resourcedocsgen/pkg/docs/refs.go
```diff
@@ -0,0 +1,201 @@
...
+func (dctx *Context) renderRef(pkgRef schema.PackageReference, ref schema.DocRef) string {
+	type refGroup struct {
+		name string
+		tags []string
+	}
+	var groups []*refGroup
+	byName := map[string]*refGroup{}
+	for _, entry := range refLanguages {
+		helper := dctx.getLanguageDocHelper(entry.lang)
+		name, ok, err := helper.ResolveDocRef(pkgRef, schema.DocRef{}, ref)
```
fnune: Did we forget to pass a full second argument ehre? I think without that we'll always get the fully-qualified ref.

tools/resourcedocsgen/pkg/docs/refs_test.go
fnune, line 1: Let's add coverage for refs to external packages? I can't tell if the feature works for those.

tools/resourcedocsgen/pkg/docs/refs.go
```diff
@@ -0,0 +1,201 @@
...
+		return groups[0].name
+	}
+
+	var b strings.Builder
+	for _, group := range groups {
+		// class="inline" is load-bearing: the registry theme sets
+		// `pulumi-choosable { display: block }` in @layer base
+		// (theme/src/scss/main.scss), so without the class every ref would
+		// break its sentence onto its own line.
+		fmt.Fprintf(&b, `<pulumi-choosable type="language" values=%q class="inline">%s</pulumi-choosable>`,
```
fnune: Needs HTML escaping.

fnune (approved):
Approving preemptively, however I find it would be good to fix some bugs, particularly the HTML escaping. It would be nice to add some refs to `scripts/tests/schema.json` so that the preview build can be used to test things.

Nit about code style: let's just remove all those comments, they make the code very hard to read.

tools/resourcedocsgen/pkg/docs/refs.go
Frassle: Wonder if that's best fixed by changing InterpretPulumiRefs itself. Make it return a diagnostic list and the translated as much as possible description. I think this is fine for now but I'll change IPR to do that and then pull it in here on the next update.

tools/resourcedocsgen/pkg/docs/refs.go
Frassle: Yeh I think this needs passing in from the top

PR #12375, opened 2026-09-03
Generate Overview page Installation and Configuration sections from the schema

### LINKED ISSUE #12352: Automate Overview page installation and configuration sections from the schema
Follow-up to #12247 / #12341, which established [the Overview page standard](https://github.com/pulumi/registry/blob/master/docs/overview-page.md). Two of the sections that standard requires are mechanically derivable from a package's `schema.json`, and today every package author writes them by hand — which is why they are so often incomplete or stale.

## 1. Template the installation instructions from the schema

`## Installation` needs an install command for each supported language: npm, pip, `go get`, `dotnet add package`, Maven/Gradle coordinates, and `pulumi package add` for YAML and HCL. Everything needed to render those is already in the schema — the package name, version, and the per-language `packages`/`respectRequiredParameters` naming overrides that decide what the published artifact is actually called.

**Output a snippet the author pastes into `docs/_index.md`,** rather than generating the page. Authoring stays in the provider repo, and the tool stays advisory — that matches how `_index.md` is sourced today (`resourcedocsgen metadata from-github` fetches it from the release tag; we do not own it).

Worth handling: a package that publishes no SDKs should emit just the `pulumi package add` command. 77% of single-page packages are in that category, and many currently title the section "Generate Provider".

## 2. Configuration reference tooling

`## Configuration` asks for every parameter with its name, whether it is required, whether it is secret, and a description. The schema's `config` block carries all four.

**a. A drift linter.** Compare the provider `config` in the schema against the hand-written configuration reference in `_index.md`, and report divergence **in both directions**:

- parameters in the schema that the docs never mention (the common case after a provider adds config)
- parameters documented that no longer exist in the schema (the common case after a major-version rewrite — exactly what made the deleted install pages in #12225/#12226 wrong)
- `required` / `secret` disagreeing between the two

Not necessarily a CI gate to start with. A linter we can run across the registry and act on is the valuable part; wiring it into `validate-packages.sh` or the community-package fact-sheet can come later once we know the false-positive rate. Parsing a hand-written markdown table or list is the hard part — worth deciding early whether we lint loosely (name presence) or ask for a recognizable format.

**b. A snippet generator.** Produce an initial configuration reference from the schema's `config` block, in the format the standard asks for, for an author to paste and then edit.

### The part the schema can't give you

Both config tools need to leave room for two things the standard explicitly requires and the schema cannot express — the generator should emit a placeholder for them, and the linter must not delete or flag them:

- **Environment-variable fallbacks.** Frequently not in the Pulumi (or, for a bridged provider, the Terraform) schema at all, because they are read by the vendor SDK a layer beneath it.
- **Mutually exclusive options.** Which parameters conflict, and which must be supplied together. This lives only in prose, on every parameter involved.

## Notes

- Both tools are natural subcommands of `resourcedocsgen`, which already reads schemas and knows the per-language naming rules — reusing that beats a standalone script.
- The generated install snippet and the config reference are the two things the [community-package fact-sheet](https://github.com/pulumi/registry/blob/master/scripts/ci/community-package/fact_sheet.py) could eventually check on a submission, so keeping the output machine-comparable is worth doing from the start.

jkodroff:
Fixes #12352 (partially — the configuration drift linter, item 2a, lands separately).

# Josh Says

This is intended as the beginnings of a standardized structure for the registry pages that should be the same everywhere: code that will give new registry packages a starting point for better-looking docs.

It does 2 things:

- Standardize the installation instructions based off the schema for all languages, with an emphasis on CLI tools rather than a link to the package in e.g. npm
- Template out a starting point for the provider configuration parameters based on the schema (with an option for both bulleted lists, the default, and a MD table). This is just a starting point because the info in the schema may not be exhaustive for _each_ option (sometimes it is, sometimes it's not), but _the schema contains all of the config options that should be documented_.

Evolving the standard provider index pages is a challenging problem given that the source docs live upstream in the provider repos as full Markdown docs, but this at least puts "better" and "standardized" into code. I'm not sure yet how we might roll out the improved structure to all providers in the Registry.

## Sample Screenshots - Install

With a published SDK:

image

Without a published SDK:

image

YAML (never uses an SDK):

image

HCL (a special little snowflake):

image

## Sample Screenshots - Provider Config

Bullet list (more compact):

image

MD tables:

(Note that we can't template these for a nice display because they have to be hand-edited after the fact and there's no way to get a macro to render where in the provider repo, where these need to be sourced.)

image

# Claude Says

## What

Two of the sections [the Overview page standard](https://github.com/pulumi/registry/blob/master/docs/overview-page.md) requires are mechanically derivable from a package's `schema.json`, and today every author writes them by hand — which is why they are so often incomplete or stale.

This adds two **experimental** `resourcedocsgen` commands that print a markdown snippet to stdout:

```bash
resourcedocsgen gen-install --schemaFile provider/cmd/pulumi-resource-example/schema.json --version v1.2.3
resourcedocsgen gen-config  --schemaFile provider/cmd/pulumi-resource-example/schema.json
```

Both are **advisory**. The Overview page is authored in the provider's own repository and fetched from its release tag by `metadata from-github`, so these generate something to paste and edit, not a page. Nothing in CI calls them, and no behavior changes for any existing build.

## What the schema can and cannot answer

**It cannot tell you which SDKs a package publishes.** The `language` blob is written by the code generator, not by the release pipeline. `chainguard-dev/cosign` declares all five languages and publishes none; `pulumi-vault` declares no `java` block yet publishes `com.pulumi/vault`. So `gen-install` guesses and takes `--languages` as the correction, with `--languages none` for the ~100 bridged packages that publish nothing.

**Every one of the seven chooser languages gets a tab regardless.** A reader who picks C# and finds an empty panel has been told nothing; one told to run `pulumi package add` has a working path. Verified empirically that the command reads the same for every language, because it takes the SDK language from the `runtime` in the reader's `Pulumi.yaml`:

| Probe | Result |
|---|---|
| `pulumi package add random` in a `runtime: nodejs` project | generates SDK, installs deps, records `packages:` in Pulumi.yaml, prints `import * as random from "@pulumi/random"` |
| same command in a `runtime: go` project | same, prints the Go import path |
| outside any project | requires `--language nodejs\|python\|go\|dotnet\|java` |

**HCL is the exception and never uses `pulumi package add`.** Its tab emits a `required_providers` block plus `pulumi install`. A native provider is sourced as `pulumi/<name>` and pinned to an exact semver version; a package parameterized over a Terraform provider reuses that provider's upstream source and version, recovered from the base64 `parameterization.parameter` — `cosign` → `chainguard-dev/cosign`, matching its committed page exactly.

**`gen-config`** reads the provider `config` block, resolving one level of `$ref` and folding in `defaultInfo.environment`. Requiredness comes from the config block's `defaults` list (the JSON key does not match the Go field name), and is only as accurate as the schema is — `pulumi-vault` marks nothing required and does not mark its `token` secret. The output says so honestly rather than inventing it, and ends with placeholders for the two things the standard requires that no schema can express: environment-variable fallbacks and mutually exclusive options.

It emits a bullet list by default, matching ~110 of the 149 existing configuration references; `--style table` emits a GFM table instead.

## Docs

`docs/overview-page.md` gains a canonical configuration example in both shapes, plus three corrections the implementation surfaced:

- **Installation** now says to give every language a tab, reversing "list only the languages your package actually supports".
- Its **HCL tab showed `pulumi package add`**, which HCL does not use at all.
- Its **`pulumi plugin install` note** now leads with "you almost certainly do not need this". `pluginDownloadURL` is compiled into every SDK, so its presence is evidence the manual step is *not* needed — the opposite of what the note implied.

`## Example Usage` also gains chooser guidance and a worked three-language example. **Nothing here generates that section**; it is included because it had the same gap — it demanded a program in every language without saying how to present them.

## Risk

The only change to pre-existing code is exporting `docs.SanitizeDescription`, reused to strip the `<span pulumi-lang-*>` markup bridged providers carry in config descriptions. The `cmd/testdata` goldens are byte-identical, which is the check that the rename altered nothing.

## Testing

`go test ./...`, `make lint-resourcedocsgen` and `make lint-markdown` pass. Golden fixtures are trimmed real schemas, each chosen because it breaks a naive implementation: `ise` (five languages on pure defaults), `vault` (`/sdk/v7` Go path, no `java` block despite a published Java SDK, no `defaults` key, a `$ref` config type), `logfire` (scoped npm name, `<span pulumi-lang-*>` in a description), `cosign` (all five language blocks, zero SDKs, parameterized), and an empty-config package. No test touches the network.

Rendered locally against the Logfire page to confirm the chooser, the HCL block and the configuration list all display correctly.

https://claude.ai/code/session_01GXzNvw9yTH4Kcdx6ZhsKKt

### DIFF
--- a/docs/overview-page.md
+++ b/docs/overview-page.md
@@ -52,7 +52,7 @@ Keep it to a short paragraph. The detail goes in the sections below.
 
 Give the installation command for **each language the package supports**. A command is more useful than a link, because a reader can copy it and run it.
 
-Wrap them in the site's language chooser so a reader sees only the language they use. The chooser is a pair of Hugo shortcodes — `chooser` with angle-bracket delimiters, `choosable` with percent delimiters — and the language keys are `typescript`, `python`, `go`, `csharp`, `java`, `yaml` and `hcl`. List in the `chooser` tag only the languages your package actually supports:
+Wrap them in the site's language chooser so a reader sees only the language they use. The chooser is a pair of Hugo shortcodes — `chooser` with angle-bracket delimiters, `choosable` with percent delimiters — and the language keys are `typescript`, `python`, `go`, `csharp`, `java`, `yaml` and `hcl`. List all seven — a language you publish no SDK for still gets a tab, carrying `pulumi package add` (see the notes below):
 
 ````markdown
 ## Installation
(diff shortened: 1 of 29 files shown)


cnunciato (approved):
This looks good! I'm not sure I _love_ the bullet lists for config items (I just find them hard to parse visually), but we can explore some alternative (maybe more tabular somehow?) approaches in the new design.

PR #12402, opened 2026-09-07
add incsteps/pulumi-provider-multipass

jagedn:
I would like to add our Multipass provider to the community


fnune:
Thanks for the provider.

I have some LLM-assisted review items. Please consider them before we merge. Some look like real bugs. Either way we need a rebase due to conflicts at the moment.

- `restore` is an invoke. Pulumi runs invokes during `pulumi preview`, so a preview restores the VM and discards its state. Make it a resource.
- `Instance.Delete` calls `multipass purge`. That destroys every trashed instance on the host, not only yours.
- `Diff` omits `name`, so a rename does nothing.
- `launchTimeout` and `operationTimeout` are never read. `Launch` hardcodes `--timeout 900`.
- The 13 resource inputs have no descriptions. Add `Annotate` to the args structs, as `Config` does.
- The `--server` line in `docs/_index.md` has a Markdown link inside the code block, and ends with `\\` instead of `\`.

jagedn:
dear @fnune just released v0.3.0 fixing issues your commented , hope this time the provider will pass the QA

fnune:
Thanks for v0.3.0, most of the items look addressed now.

One question: [`Instance.Delete`](https://github.com/incsteps/pulumi-provider-multipass/blob/v0.3.0/resources/instance.go#L151) still calls `multipass purge` after `multipass delete <name>`. I'm not very familiar with Multipass, but from what I understand `purge` doesn't take an instance name, so it might also remove other deleted instances on the host that Pulumi doesn't manage. Is that safe? If not, maybe `multipass delete --purge <name>` would work instead.

Two smaller things I noticed:

- `docs/_index.md` still mentions `v0.2.0` in the install command, and it mentions snapshots, which seem to be gone in v0.3.0.
- `operationTimeout` doesn't seem to be read anywhere.

If you publish a new release, comment `/check` so the check runs against that version.

jagedn:
> Thanks for v0.3.0, most of the items look addressed now.
> 
> One question: [`Instance.Delete`](https://github.com/incsteps/pulumi-provider-multipass/blob/v0.3.0/resources/instance.go#L151) still calls `multipass purge` after `multipass delete <name>`. I'm not very familiar with Multipass, but from what I understand `purge` doesn't take an instance name, so it might also remove other deleted instances on the host that Pulumi doesn't manage. Is that safe? If not, maybe `multipass delete --purge <name>` would work instead.
> 
> Two smaller things I noticed:
> 
>     * `docs/_index.md` still mentions `v0.2.0` in the install command, and it mentions snapshots, which seem to be gone in v0.3.0.
> 
>     * `operationTimeout` doesn't seem to be read anywhere.
> 
> 
> If you publish a new release, comment `/check` so the check runs against that version.

yes, you're right, maybe better to avoid purge not managed by Pulumi. I'll create a 0.3.1 version an use the comment to re-run the check 

thanks

fnune (approved):

fnune:
Thanks for your contribution & your patience @jagedn!

PR #12430, opened 2026-09-08
Suggest OpenTofu providers when registry search finds nothing

### LINKED ISSUE #7299: Add a helpful message RE Any TF Provider if no registry search results are returned
If I search for a provider, and nothing is found, we should make sure the user knows they can use Any TF provider to generate an SDK for any provider that's in the OpenTofu registry.

* Good: a link to the docs showing how to use Any Terraform Provider and a link to the OpenTofu registry (both open in new tabs)
* Better: Dynamically generate a search URI for the OpenTofu registry based on the user's input
ew browser tab.
* Best: Automatically search the OpenTofu registry (or give them a button to do the search) and display the results inline alongside a `pulumi package add` command that also includes the version of the provider (for maximum build repeatability), e.g. `pulumi package add terraform-provider hashicorp/random 3.7.1`

Good is good enough for me.

jkodroff:
Closes #7299.

Filtering the registry for a provider Pulumi doesn't publish was a dead end — even though Pulumi can run any Terraform or OpenTofu provider. The empty state now looks the query up in the OpenTofu registry and offers each match with a version-pinned `pulumi package add` command.

```
No packages in the Pulumi Registry match "sops".
────────────────────────────────────────────────
5 providers in the OpenTofu registry match "sops":

  carlpett/sops  ★ 584
  $ pulumi package add terraform-provider carlpett/sops 1.4.1
  …

Pulumi can use any Terraform or OpenTofu provider… Learn how →
```

The issue scoped this as good/better/best. **Better turned out not to be buildable** — `search.opentofu.org` keeps its query in React local state and has no `/search` route, so no URL can pre-fill it. **Best turned out cheap**, because the API is CORS-open and its `popularity` field is GitHub stars — the signal needed to tell a real provider from its forks.

## Ranking

The API's own order is not trustworthy: for `random` it returns `brenthc/random` (★0) above `hashicorp/random` (★233). Results are re-sorted by stars, then by a namespace matching the provider name so the vendor's own publication wins a tie (`namecheap/namecheap` over `adamdecaf/namecheap`, both ★168), then alphabetically for stable ordering.

## Safety

`addr` and `version` are publisher-controlled, so both are allowlist-validated before reaching a URL or a command — failures are dropped, not escaped — and every node is built with `textContent`. Verified against payloads carrying HTML injection, path traversal, `javascript:`, and shell metacharacters.

Every failure path (offline, non-200, malformed, timeout, no match) degrades to the static Any Terraform Provider docs link, which is server-rendered and needs no JavaScript.

## Also fixes: "Clear all filters" has never worked

`pulumi-registry-list-search.reset()` lacked a `@Method()` decorator, so Stencil never exposed it on the host element and the handler threw a `TypeError` on its first line — the button did nothing at all. Exposing it revealed a second bug: `pulumi-filter-select.reset()` is async, so its `filterSelect` emit landed *after* the handler's own re-filter and re-hid everything using stale input text. Both resets are now awaited.

Pre-existing and outside the issue's scope, but the button sits inside the panel this PR rebuilds. Happy to split it out.

## Testing

12 new Cypress tests (23 passing total), stubbing the API so they never hit the network: star ranking, vendor tie-break, non-provider filtering, per-`addr` de-duplication, version pinning, link format, deferred loading heading, empty/failure fallbacks, XSS rejection, the two no-fetch guards, and a regression test for the reset button. `make lint` passes.

The no-fetch guards are verified by mutation rather than by inspection: setting `MIN_QUERY_LENGTH` to 1 fails the single-character test. An earlier draft of that test used `"q"`, which matches ten visible packages, so it passed whether the guard worked or not.

Note `cypress/e2e/site.cy.js` isn't wired into any CI workflow, so these are local-verification only — tracked in #12450.

## Follow-ups

- #12450 — nothing in this PR runs in CI, and the feature depends on an external API contract (`type`, `addr`, `version`, `popularity`) nobody here controls.
- The zero-result query is sent to `api.opentofu.org` with no consent gate. No identifier is attached and it only fires once the empty state is reached, but it's the first client-side request to a non-Pulumi origin driven by user input in this bundle, so it's worth an explicit call from whoever owns pulumi.com privacy. (There's no CSP to worry about — `infrastructure/index.ts:218` is the AWS managed `SecurityHeadersPolicy`, which sets no `Content-Security-Policy`.)

## Known limitation

OpenTofu's search substring-matches the namespace, so `plex` returns `simplexiengage/postgresql`. Left as-is rather than filtering unasked — worth a follow-up on whether to require the query to match the provider name.

https://claude.ai/code/session_01Dk8u8qqdtjHxAfgcgQehad

### DIFF
--- a/cypress/e2e/site.cy.js
+++ b/cypress/e2e/site.cy.js
@@ -110,10 +110,23 @@ describe("www.pulumi.com/registry", () => {
         };
 
         // Types into the filter and waits for the empty state to actually be reached,
-        // so assertions don't race the search component's 300ms debounce.
+        // so assertions don't race the search component's 300ms debounce. The
+        // zero-visible-packages assertion is retried, and it can't pass until
+        // filterByTextAndTags -> updateAllCount -> syncEmptyState has run, which is
+        // what makes the "did we fetch?" assertions below deterministic.
         const filterToEmpty = (query) => {
             cy.get(".registry-filter-input").clear().type(query);
             cy.get(".all-packages .package:not(.hidden)").should("have.length", 0);
+        };
+
+        // Ticks a tag filter. The option checkboxes live in shadow DOM behind an
+        // opacity-transitioned menu, so this drives pulumi-filter-select's own
+        // select() @Method() instead. The should() retries until Stencil has
+        // upgraded the element and put the method on it.
+        const selectFilter = (value) => {
+            cy.get("pulumi-filter-select")
+                .should(($el) => expect($el[0].select).to.be.a("function"))
+                .then(($el) => $el[0].select({ value }));
         };
 
         beforeEach(() => {
@@ -153,6 +166,46 @@ describe("www.pulumi.com/registry", () => {
                     "have.text",
                     "$ pulumi package add terraform-provider carlpett/sops 1.4.1",
                 );
+        });
+
+        it("collapses repeated rows for the same provider", () => {
+            // The endpoint looks like one row per documentation page, so a provider
+            // could plausibly return several type:"provider" rows and eat several of
+            // the five slots. Only the best-ranked row for each addr survives.
+            stubTofu({
+                body: [
+                    { type: "provider", addr: "carlpett/sops", version: "v1.4.1", popularity: 584 },
+                    { type: "provider", addr: "carlpett/sops", version: "v1.3.0", popularity: 584 },
+                    { type: "provider", addr: "nobbs/sops", version: "v0.3.3", popularity: 18 },
+                ],
+            });
+            filterToEmpty("sops");
+            cy.wait("@tofu");
+
+            cy.get(".tf-suggest .tf-suggest-item").should("have.length", 2);
+            cy.get(".tf-suggest .tf-suggest-item")
+                .first()
+                .find("code")
+                .should("contain.text", "carlpett/sops 1.4.1");
+        });
+
+        it("announces the search only once the lookup is visibly slow", () => {
+            // The loading heading is deferred so a lookup that resolves at typing
+            // speed never paints an intermediate state into the aria-live region.
+            // A slow one still says what it is doing.
+            cy.intercept("GET", TOFU_API, { body: twoProviders, delay: 1500 }).as("tofu");
+            filterToEmpty("sops");
+
+            cy.get(".tf-suggest .tf-suggest-message").should(
+                "contain.text",
+                "Searching the OpenTofu registry",
+            );
+
+            cy.wait("@tofu");
+            cy.get(".tf-suggest .tf-suggest-message").should(
+                "contain.text",
+                "2 providers in the OpenTofu registry",
+            );
         });
 
         it("prefers the vendor's own namespace when stars tie", () => {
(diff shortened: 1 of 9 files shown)


jkodroff:
Thanks — all four code-level items are fixed in 6299254. Two of your findings needed correcting, and one answers itself.

## 1. Bogus "no fetch" tests — confirmed, and worse than you found

You were right about the mechanism and understated the scale. `"q"` matches **ten** visible packages, not two: `azure-quickstart-acr-geo-replication`, `cloudamqp`, `equinix`, `mssql`, `mysql`, `netskope-publisher`, `postgresql`, `qdrant-cloud` and more. Your second example, `qovery`, is `publisher: DEPRECATED` and hidden by default, so it never entered into it — but `qdrant-cloud` alone was enough to keep the panel CSS-hidden and the guard unreached.

`~` checks out: zero matches across every package name, title, and keyword. Both tests now use it or real tag filters, and I verified they're load-bearing rather than trusting that they look right — setting `MIN_QUERY_LENGTH` to 1, rebuilding, and re-running fails the single-character test. It passed that mutation before.

The tag-only test now selects Deprecated + Version Control. Both version-control packages are current, so the pair genuinely empties the list with an empty search box. It drives `pulumi-filter-select`'s own `select()` `@Method()` rather than clicking, since the option checkboxes are in shadow DOM behind an opacity-transitioned menu.

Duplicate test deleted, and both `cy.wait(500)` calls are gone — the retried zero-visible-packages assertion can't resolve until `filterByTextAndTags` → `updateAllCount` → `syncEmptyState` has run, which is the gate those waits were guessing at. No bare waits left in the block.

## 2. De-dupe — taken, with a test

Applied exactly as suggested, between the sort and the slice so each `addr` keeps its best-ranked row. Added a test with two `type: "provider"` rows sharing an `addr`.

## 3. Live region — taken

Went further than deferring alone: the region is now **cleared** immediately (removing nodes announces nothing, and `.tf-suggest:empty` collapses the panel), and only the heading is deferred, behind a 300ms timer matching the search component's own lodash debounce. Clearing first matters because the Pulumi message above has already been rewritten for the new query — deferring without clearing would leave a heading naming one query above results for another.

Net effect: a lookup that settles at typing speed paints and announces nothing but its result. Slow ones still get the heading, which is where it earns its keep. Test stubs an 800ms+ delay to prove that path still fires.

## 4. Nothing runs in CI — filed as #12450

Agreed, and worth stating as plainly as you did. I filed #12450 rather than widening this PR, and included your point about the external contract: if OpenTofu renames `popularity`, `Number(undefined)` is `NaN`, everything sorts as zero stars, ranking degrades to alphabetical, and nothing fails or alerts. The issue notes that a PR-triggered suite with stubbed responses can't catch that on its own, so a scheduled check against the live endpoint is a separate question.

## 5. CSP — no action needed; privacy — flagged, not decided

The CSP half answers itself. `infrastructure/index.ts:218` uses the AWS **managed** `SecurityHeadersPolicy` (`67f7725c-6f97-4210-82d7-5512b31e9d03`), which sets HSTS, frame-options, referrer-policy and nosniff but **no `Content-Security-Policy`**. There's no `connect-src` to add and nothing to coordinate with edge config. Good instinct to check — the failure mode you describe (dead in production, green locally) would have been nasty.

The privacy half is a real question and not one to settle in review, so I'm flagging it for whoever owns pulumi.com privacy rather than gating the fetch here. Noting for the record what actually leaves the browser: a search string that matched no Pulumi package, sent to `api.opentofu.org`, with no identifier attached and only when the empty state is reached.

## Nits

All four taken: error-path comment reworded to credit `renderedKey` instead of the cache; `formatStars` boundary moved to 999500, the exact point where `Math.round(n / 1000)` hits 1000; one `QUERY_ECHO_MAX` constant replacing the 60/80 split; and `.tf-suggest-fallback` now carries `text-gray-700` in light, matching `.tf-suggest-meta` beside it. Verified both colors in the compiled bundle rather than by eye — light resolves to `var(--color-gray-700)`, dark to `var(--docs-fg-muted)`, identical to `.tf-suggest-meta`.

## Verification

`make lint` passes. Cypress is at 23 tests, green across seven consecutive runs. `make build-assets` compiles clean.

Not done: a visual pass in a real browser — the Chrome extension isn't connected on this machine, so the light/dark check and the no-flicker check are from the compiled CSS and the test suite rather than from looking at it.

https://claude.ai/code/session_016mVvwFtf3W2pX6B6AE5PEW

jkodroff:
Confirmed with @tyler-dunkel that there's no serious privacy concerns, but we do want to proxy the OpenTofu search so that there's no CORS concerns, etc. That's covered in https://github.com/pulumi/registry/issues/12451

cnunciato (approved):
LGTM! We'll be sure to add this in the reboot as well. (Great idea!)

PR #12523, opened 2026-09-14
Port the smart link checker from pulumi/docs

CamSoper:
Context: https://pulumi.slack.com/archives/C07QLREMFU6/p1789403976847699?thread_ts=1789401399.038679&cid=C07QLREMFU6

The weekly link check has been dumping every raw hit into #registry-ops, false positives included (Maven Central 403s, ansible.com 429s, GitHub's own logged-out nav on github.com/pulumi/pulumi). Nobody was triaging it. This ports the pulumi/docs approach so the bot weeds out the noise, categorizes what's left, and ships a PR instead of a wall of links.

## What changes

**`scripts/link-checker/check-links.js`**
- Writes filtered results to `.broken-links.json` (`internal` / `external`) instead of posting to Slack. Clean runs write empty lists so the workflow can branch on it.
- Ports the docs false-positive filters: all 429s, 502/415, and 401/403/202 on external hosts. Internal links with those codes are still reported.
- Excludes known-noisy hosts from registry runs: Maven Central, npm, Docker Hub, status pages, GitHub's logged-out chrome.
- Fixes a long-standing bug: BLC's glob matcher requires at least one character after `*`, so the existing `https://github.com/marketplace*` exclusion never matched the bare URL. That's why `github.com/marketplace` showed up every single week.
- Drops `@slack/web-api` (nothing else used it).

**`.github/workflows/check-links.yml`**
- Mirrors the docs flow: ESC before checkout so the PR goes out as pulumi-bot, a detect step that skips everything on a clean run, `claude-code-action` driving the fix, then a Slack post to #registry-ops with just the PR link.
- Still weekly on Mondays. Timeout bumped to 300 min to leave room for the fix step after the crawl.

**`.claude/commands/fix-broken-links/SKILL.md`** (new, workflow-only)
- Registry-specific triage. The key difference from docs: package overview and installation-configuration pages are fetched from the upstream provider repo, so the skill never hand-edits generated content. Instead:
  - retired registry path with a successor -> S3 redirect in `scripts/redirects/`
  - content this repo owns -> edit at source
  - generated page, upstream under `github.com/pulumi/` -> file an issue in the provider repo (deduped first)
  - generated page, third-party upstream -> recorded in a "Third-party packages" section of the PR, not chased (per Josh's point that we can't control those)
  - recurring dead/bot-protected external -> exclusion list
- Re-verifies every link before acting and dedupes against open PRs/issues, same as docs.
- Runs `yarn run lint` + `make lint-markdown`, never `make build`.

**Housekeeping:** `.gitignore` entries for the two output files, `BUILD-AND-DEPLOY.md` updated.

## Verification

- `yarn run lint` and `make lint-markdown` pass.
- Smoke-tested the checker against a local throwaway site with a real 404, a self-link, and an excluded URL: the 404 lands in `.broken-links.json`, the excluded URL doesn't, and `github.com/marketplace` is finally filtered.
- Not yet run end-to-end in CI. Plan is to merge, then `workflow_dispatch` it and watch the first PR the bot opens.


.claude/commands/fix-broken-links/SKILL.md
```diff
@@ -0,0 +1,143 @@
+---
+name: fix-broken-links
+description: Fix broken links reported by the weekly registry link checker. Reads .broken-links.json, triages each entry to a fix strategy (S3 redirect, source edit, exclusion, upstream issue, or third-party/not actioned), and opens an auditable PR. Invoked by the check-links workflow; not user-invocable.
+user-invocable: false
+---
+
+# Fix Broken Links
+
+You are fixing the broken links found by the weekly link checker (`scripts/link-checker/check-links.js`, run against `https://www.pulumi.com/registry`). Work through every reported entry, pick the right strategy for each, open one auditable PR, and file upstream issues only where Pulumi owns the source.
```
jkodroff, line 9: > and file upstream issues only where Pulumi owns the source.

There could be a problem here in potentially flooding the upstream repos with issues that can't really be resolved absent a major change in how we manage docs.

.claude/commands/fix-broken-links/SKILL.md
```diff
@@ -0,0 +1,143 @@
...
+
+The registry is different from `pulumi/docs` in one way that drives most of the triage below: **almost every page a reader sees is generated from a provider repository, not authored here.** A package's overview (`_index.md`) and `installation-configuration.md` are fetched verbatim from `docs/` in the upstream provider repo at publish time, and its API docs are generated from the provider schema. A broken link on one of those pages can't be fixed in this repo. Whether anyone can fix it depends on who maintains the upstream repo.
```
jkodroff, line 11: >  A package's overview (`_index.md`) and `installation-configuration.md` are fetched verbatim from `docs/` in the upstream provider repo at publish time

Not always. Some of them are hosted here.

.claude/commands/fix-broken-links/SKILL.md
CamSoper: Agreed, fixed in 451c4ce. The skill files no issues anywhere now, and `gh issue create` is out of the tool grant so it can't. Broken links on upstream-owned pages go in an "Upstream-owned pages" section of the PR, split first-party / third-party with each page's `edit_url`, so a human can decide if any are worth chasing. If nothing's fixable here, there's no PR at all and the Slack post just summarizes the counts.

---

.claude/commands/fix-broken-links/SKILL.md
CamSoper: Yep, good catch. 87 of the overview/install pages have no fetched-from header, plus every guide-style sub-page (the aws 7.0 migration guide and azure-native v2-to-v3 pages from the Slack dump included). The skill now decides ownership per file: the `# WARNING: this file was fetched from` header plus `edit_url` means upstream, and no header means we own it and fix it at source. Fixed in 451c4ce.

---

CamSoper:
Addressed the Claude review in 451c4ce:

1. **Stale Slack bullet in BUILD-AND-DEPLOY.md** -- removed.
2. **Silent failure** -- took the suggestion. Slack step runs on `always()`, falls back to a count + run link when `.broken-links-pr.txt` is missing, and pipes the response through `jq -e '.ok'` so a Slack error fails the step.
3. **calmcard** -- checked against BLC's `lib/internal/matchUrl.js`. It tries a literal `indexOf` first and only falls back to calmcard, so the host-root entries (`https://www.npmjs.com/`, `https://search.maven.org/`, etc.) do exclude everything under them. Smoke-tested with `npmjs.com/package/does-not-exist-...` and it's excluded. The calmcard fallback *is* anchored with `*` consuming 1+ chars, though, which means the existing `/registry/packages/*/api-docs` keyword has never matched anything. That's pre-existing and changing it would alter what gets checked, so I left it alone here. Reworded the marketplace comment and the skill's matching explanation to match the real behavior.
4. **Tool grant** -- dropped `gh api`, `curl`, `cd`, and (per Josh's feedback) `gh issue create`. The skill now re-verifies with `WebFetch` only.

Nits: `env:` moved back above `jobs:`, and I dropped `name:` from the skill front matter. Kept `--effort low` on purpose. It matches the docs workflow, and I'd rather see the first real run before paying for more.

---

PR #4595, opened 2024-05-23
Add files from pulumi-hugo

sean1588:
## Description

Remove reliance on pulumi-hugo

PR #12609, opened 2026-09-21
Remove unused language shortcodes

iwahbe:
The `pulumi-new`, `langfile`, `pulumi-host`, `langhost`, `pulumi-getproject`, `example-program`, and `example-program-snippet` shortcodes came over from `pulumi-hugo` in #4595. No content, layout, or `resourcedocsgen` template in this repository uses them, and the two `example-program` partials have no caller either. No provider repository in the `pulumi` organization uses them in its registry docs.

Each of these files carries its own list of languages, so each one is a place that goes stale when the chooser component gains a language. They all stop at YAML and omit HCL. Delete them instead of updating them.

`shortcodes/langname` goes too. It is a sibling of `langhost.html` with the same language list, and it has no file extension, so Hugo never registered it as a shortcode.

The `example-program` templates were the only emitters of the `example-program` and `example-program-footer` classes, so the style rules for those classes can no longer match. This removes the rule blocks, the `.example-program` arms of the shared selector lists, and the `.example-program *` entries in the `:not()` lists. The compiled `bundle-registry` style sheet differs only by those selectors: each rule that remains equals its old form without the `.example-program` arm, in light mode and in dark mode. The `marketing-registry` bundle does not change.


iwahbe:
@claude I removed `shortcodes/langname` in 2565af384. The orphaned `.example-program` style rules stay for a follow-up PR, so that this PR only deletes templates. Please review again.

iwahbe:
@claude My last comment is out of date. a0a587da4 removes the orphaned `.example-program` style rules in this PR, including the `:not()` entries and the comments that named the class. Please review again.

PR #12648, opened 2026-09-23
Add incident-io/pulumi-incident to the community package list

joladev:
## Summary
- Add incident-io/pulumi-incident to the community package list
- Provider v0.1.1, bridged from incident-io/terraform-provider-incident v7.0.0+

## Test plan
- [ ] CI lint passes
- [ ] Pulumi team triggers registry preview build


djgrove (approved):

djgrove:
@joladev hey thanks for the contrib!

The registry already lists an incident package that Pulumi generates automatically from your Terraform provider: https://www.pulumi.com/registry/packages/incident/. Your package uses the same name, which is what we want, but our pipeline would keep overwriting your listing.

I'm removing the automatic listing on our side first, then merging this.

One thing you'll see after the switch: the page will show 0.1.2 instead of 7.2.0, and it will say your provider tracks upstream v7.0.0 until your sync job bumps it.

